PR #1, opened 2018-06-21
Bash script reworking existing embedded image to a Mender image by re…

apodogrocki:
…structuring partition table

1st stage script which operates directly on the embedded image and block device, such as SD Card.
Final aim is to compose standalone *.sdimg file which can be used to flash the SD Card.

Issues: MEN-1867

### DIFF
--- a/convert-stage-4.sh
+++ b/convert-stage-4.sh
@@ -1,0 +1,351 @@
+#!/bin/bash
+
+show_help() {
+  cat << EOF
+
+Mender executables, service and configuration files installer.
+
+Usage: $0 [options]
+
+    Options: [-i|--image | -m|--mender | -a|--artifact | -d|--device-type |
+              -p|--demo-ip | -u| --production-url | -o| --hosted-token]
+
+        --image          - .sdimg generated with mender-conversion-tool
+        --mender         - mender client binary file
+        --artifact       - artifact info
+        --device-type    - target device type identification
+        --demo-ip        - Mender demo server IP address
+        --production-url - Mender production server url
+        --certificate    - Mender server certificate
+
+    Examples:
+
+        ./mender-conversion-tool.sh install_mender --image <sdimg_file_path>
+                --device-type beaglebone --artifact release-1_1.5.0
+                --server 192.168.10.2 --mender <mender_binary_path>
+
+EOF
+  exit 1
+}
+
+tool_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+output_dir=${tool_dir}/output
+
+mender_client_repo="https://raw.githubusercontent.com/mendersoftware/mender"
+mender_client_revision="1.6.x"
+meta_mender_repo="https://raw.githubusercontent.com/mendersoftware/meta-mender"
+meta_mender_revision="sumo"
+
+mender_dir=$output_dir/mender
+device_type=
+artifact=
+# Mender demo server IP address.
+demo_ip=
+# Mender production server url.
+production_url=
+# Used server url.
+server_url="https://docker.mender.io"
+# Mender production certificate.
+certificate=
+# Mender hosted token.
+hosted_token=
+# Mender tenant token.
+tenant_token="dummy"
+
+declare -a sdimgmappings
+
+create_client_files() {
+  cat <<- EOF > $mender_dir/mender.service
+	[Unit]
+	Description=Mender OTA update service
+	After=systemd-resolved.service
+
+	[Service]
+	Type=idle
+	User=root
+	Group=root
+	ExecStartPre=/bin/mkdir -p -m 0700 /data/mender
+	ExecStartPre=/bin/ln -sf /etc/mender/tenant.conf /var/lib/mender/authtentoken
+	ExecStart=/usr/bin/mender -daemon
+	Restart=on-abort
+
+	[Install]
+	WantedBy=multi-user.target
+	EOF
+
+  cat <<- EOF > $mender_dir/mender.conf
+	{
+	    "InventoryPollIntervalSeconds": 5,
+	    "RetryPollIntervalSeconds": 1,
+	    "RootfsPartA": "/dev/mmcblk0p2",
+	    "RootfsPartB": "/dev/mmcblk0p3",
+	    "ServerCertificate": "/etc/mender/server.crt",
+	    "ServerURL": "$server_url",
+	    "TenantToken": "$tenant_token",
+	    "UpdatePollIntervalSeconds": 5
+	}
+	EOF
+
+  cat <<- EOF > $mender_dir/artifact_info
+	artifact_name=${artifact}
+	EOF
+
+  # Version file
+  echo -n "2" > $mender_dir/version
+
+  cat <<- EOF > $mender_dir/device_type
+	device_type=${device_type}
+	EOF
+
+  case "$device_type" in
+    "beaglebone")
+      cat <<- EOF > $mender_dir/fw_env.config
+	/dev/mmcblk0 0x800000 0x20000
+	/dev/mmcblk0 0x1000000 0x20000
+	EOF
+      ;;
+    "raspberrypi3")
+      cat <<- EOF > $mender_dir/fw_env.config
+	/dev/mmcblk0 0x400000 0x4000
+	/dev/mmcblk0 0x800000 0x4000
+	EOF
+      ;;
+  esac
+}
+
+get_mender_files_from_upstream() {
+  mkdir -p $mender_dir
+
+  echo -e "Downloading inventory & identity scripts..."
+
+  wget -q -O $mender_dir/mender-device-identity \
+    $mender_client_repo/$mender_client_revision/support/mender-device-identity
+  wget -q -O $mender_dir/mender-inventory-bootloader-integration \
+    $mender_client_repo/$mender_client_revision/support/mender-inventory-bootloader-integration
+  wget -q -O $mender_dir/mender-inventory-hostinfo \
+    $mender_client_repo/$mender_client_revision/support/mender-inventory-hostinfo
+  wget -q -O $mender_dir/mender-inventory-network \
+    $mender_client_repo/$mender_client_revision/support/mender-inventory-network
+  wget -q -O $mender_dir/mender-inventory-os \
+    $mender_client_repo/$mender_client_revision/support/mender-inventory-os
+  wget -q -O $mender_dir/mender-inventory-rootfs-type \
+    $mender_client_repo/$mender_client_revision/support/mender-inventory-rootfs-type
+  wget -q -O $mender_dir/server.crt \
+    $meta_mender_repo/$meta_mender_revision/meta-mender-demo/recipes-mender/mender/files/server.crt
+}
+
+install_files() {
+  local primary_dir=$1
+  local data_dir=$2
+
+  identitydir="usr/share/mender/identity"
+  inventorydir="usr/share/mender/inventory"
+  sysconfdir="etc/mender"
+  bindir="usr/bin"
+  systemd_unitdir="lib/systemd/system"
+  localstatedir="var/lib/mender"
+  dataconfdir="mender"
+  databootdir="u-boot"
+
+  # Prepare 'data' partition
+  sudo install -d -m 755 ${data_dir}/${dataconfdir}
+  sudo install -d -m 755 ${data_dir}/${databootdir}
+
+  sudo install -m 0644 ${mender_dir}/device_type ${data_dir}/${dataconfdir}
+  sudo install -m 0644 ${mender_dir}/fw_env.config ${data_dir}/${databootdir}
+
+  sudo ln -sf /data/${databootdir}/fw_env.config ${primary_dir}/etc/fw_env.config
+
+  # Prepare 'primary' partition
+  [ ! -d "$primary_dir/data" ] && \
+      { echo "'data' mountpoint missing. Adding"; \
+        sudo install -d -m 755 ${primary_dir}/data; }
+
+  case "$device_type" in
+    "beaglebone")
+      [ ! -d "$primary_dir/boot/efi" ] && \
+          { echo "'/boot/efi' mountpoint missing. Adding"; \
+            sudo install -d -m 755 ${primary_dir}/boot/efi; }
+      ;;
+    "raspberrypi3")
+      [ ! -d "$primary_dir/uboot" ] && \
+          { echo "'/boot/efi' mountpoint missing. Adding"; \
+            sudo install -d -m 755 ${primary_dir}/uboot; }
+      ;;
+  esac
+
+  sudo install -d ${primary_dir}/${identitydir}
+  sudo install -d ${primary_dir}/${inventorydir}
+  sudo install -d ${primary_dir}/${sysconfdir}
+  sudo install -d ${primary_dir}/${sysconfdir}/scripts
+
+  sudo ln -s /data/${dataconfdir} ${primary_dir}/${localstatedir}
+
+  sudo install -m 0755 ${mender} ${primary_dir}/${bindir}/mender
+
+  sudo install -t ${primary_dir}/${identitydir} -m 0755 \
+      ${mender_dir}/mender-device-identity
+
+  sudo install -t ${primary_dir}/${inventorydir} -m 0755 \
+      ${mender_dir}/mender-inventory-*
+
+  sudo install -m 0644 ${mender_dir}/mender.service ${primary_dir}/${systemd_unitdir}
+
+  # Enable menderd service starting on boot.
+  sudo ln -s /lib/systemd/system/mender.service \
+      ${primary_dir}/etc/systemd/system/multi-user.target.wants/mender.service
+
+  sudo install -m 0644 ${mender_dir}/mender.conf ${primary_dir}/${sysconfdir}
+
+  sudo install -m 0444 ${mender_dir}/server.crt ${primary_dir}/${sysconfdir}
+
+  sudo install -m 0644 ${mender_dir}/artifact_info ${primary_dir}/${sysconfdir}
+
+  sudo install -m 0644 ${mender_dir}/version ${primary_dir}/${sysconfdir}/scripts
+
+  if [ -n "${demo_ip}" ]; then
+    echo "$demo_ip docker.mender.io s3.docker.mender.io" | sudo tee -a $primary_dir/etc/hosts
+  fi
+
+  if [ -n "${certificate}" ]; then
+    sudo install -m 0444 ${certificate} ${primary_dir}/${sysconfdir}
+  fi
+}
+
+do_install_mender() {
+  if [ -z "${image}" ]; then
+    echo ".sdimg image file not set. Aborting."
+    show_help
+  fi
+
+  if [ -z "${mender}" ]; then
+    echo "Mender client binary not set. Aborting."
+    show_help
+  fi
+
+  if [ -z "${device_type}" ]; then
+    echo "Target device type name not set. Aborting."
+    show_help
+  fi
+
+  if [ -z "${artifact}" ]; then
+    echo "Artifact info not set. Aborting."
+    show_help
+  fi
+
+  if [ -z "${production_url}" ] && [ -z "${demo_ip}" ] && \
+     [ -z "${hosted_token}" ]; then
+    echo "No server type specified. Aborting."
+    show_help
+  fi
+
+  if [ -n "${production_url}" ] && [ -n "${demo_ip}" ]; then
+    echo "Incompatible server type choice. Aborting."
+    show_help
+  fi
+
+  # TODO: more error checking of server types
+  if [ -n "${hosted_token}" ]; then
+    tenant_token=$(cat ${hosted_token} | tr -d '\n')
+    server_url="https://hosted.mender.io"
+  fi
+
+  if [ -n "${production_url}" ]; then
+    server_url=${production_url}
+  fi
+
+  [ ! -f $image ] && { echo "$image - file not found. Aborting."; exit 1; }
+
+  # Mount rootfs partition A.
+  create_device_maps $image sdimgmappings
+
+  mkdir -p $output_dir && cd $output_dir
+
+  primary=${sdimgmappings[1]}
+  data=${sdimgmappings[3]}
+
+  map_primary=/dev/mapper/"$primary"
+  map_data=/dev/mapper/"$data"
+  path_primary=$output_dir/sdimg/primary
+  path_data=$output_dir/sdimg/data
+  mkdir -p ${path_primary} ${path_data}
+
+  sudo mount ${map_primary} ${path_primary}
+  sudo mount ${map_data} ${path_data}
+
+  # Get Mender client related files.
+  get_mender_files_from_upstream
+  # Create all necessary client's files.
+  create_client_files
+
+  # Create all required paths and install files.
+  install_files ${path_primary} ${path_data}
+
+  # Back to working directory.
+  cd $tool_dir && sync
+
+  # Clean stuff.
+  detach_device_maps ${sdimgmappings[@]}
+  rm -rf $output_dir/sdimg
+}
+
+PARAMS=""
+
+while (( "$#" )); do
+  case "$1" in
+    -i | --image)
+      image=$2
+      shift 2
+      ;;
+    -m | --mender)
+      mender=$2
+      shift 2
+      ;;
+    -d | --device-type)
+      device_type=$2
+      shift 2
+      ;;
+    -a | --artifact)
+      artifact=$2
+      shift 2
+      ;;
+    -p | --demo-ip)
+      demo_ip=$2
+      shift 2
+      ;;
+    -c | --certificate)
+      certificate=$2
+      shift 2
+      ;;
+    -u | --production-url)
+      production_url=$2
+      shift 2
+      ;;
+    -o | --hosted-token)
+      hosted_token=$2
+      shift 2
+      ;;
+    -h | --help)
+      show_help
+      ;;
+    --)
+      shift
+      break
+      ;;
+    -*)
+      echo "Error: unsupported option $1" >&2
+      exit 1
+      ;;
+    *)
+      PARAMS="$PARAMS $1"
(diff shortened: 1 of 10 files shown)


mender-image-prepare.sh
```diff
@@ -0,0 +1,305 @@
...
+
+set_bootloader() {
+  echo "Setting bootloader..."
+  [ ! -d "$BASE_DIR/mnt/rootfs/opt/backup/uboot" ] && { echo "Error: check mounted location."; exit 1; }
+
+  cd $BASE_DIR/mnt/rootfs/opt/backup/uboot
+  [[ ! -f MLO || ! -f u-boot.img ]] && { echo "Error: cannot find bootloader."; exit 1; } 
+
+  cp MLO $BASE_DIR/mnt/boot/
+  cp u-boot.img $BASE_DIR/mnt/boot/
```
drewmoseley: This seems to be very board dependent.  We will probably need a way to conditionalize this on board type.

mender-image-prepare.sh
```diff
@@ -0,0 +1,535 @@
...
+  is_mounted=`grep ${target_dev} /proc/self/mounts | wc -l`
+  if [ ${is_mounted} -ne 0 ]; then
+    umount ${target_dev}?*
+  fi
+}
+
+erase_part_table() {
+  echo "2. Erase partition table..."
+  wipefs --all --force ${target_dev}?*
+  dd if=/dev/zero of=${target_dev} bs=1M count=100
```
kacf: This erases quite a lot more than just the partition table. Perhaps it should be called `erase_filesystem` instead?

mender-image-prepare.sh
```diff
@@ -0,0 +1,535 @@
...
+	loadximage=echo debug: [/boot/vmlinuz-${uname_r}] ... ; load mmc 0:2 ${loadaddr} /boot/vmlinuz-${uname_r}
+	loadxfdt=echo debug: [/boot/dtbs/${uname_r}/${fdtfile}] ... ;load mmc 0:2 ${fdtaddr} /boot/dtbs/${uname_r}/${fdtfile}
+	loadxrd=echo debug: [/boot/initrd.img-${uname_r}] ... ; load mmc 0:2 ${rdaddr} /boot/initrd.img-${uname_r}; setenv rdsize ${filesize}
+	loaduEnvtxt=load mmc 0:2 ${loadaddr} /boot/uEnv.txt ; env import -t ${loadaddr} ${filesize};
+	check_dtb=if test -n ${dtb}; then setenv fdtfile ${dtb};fi;
+	loadall=run loaduEnvtxt; run check_dtb; run loadximage; run loadxrd; run loadxfdt;
+
+	mmcargs=setenv bootargs console=tty0 console=${console} ${optargs} ${cape_disable} ${cape_enable} root=/dev/mmcblk0p2 rootfstype=${mmcrootfstype} ${cmdline}
+
+	uenvcmd=run loadall; run mmcargs; echo debug: [${bootargs}] ... ; echo debug: [bootz ${loadaddr} ${rdaddr}:${rdsize} ${fdtaddr}] ... ; bootz ${loadaddr} ${rdaddr}:${rdsize} ${fdtaddr};
```
kacf: What is the plan with `uenv.txt`? Ideally we should let U-Boot run `distro_bootcmd` which should run GRUB through `bootefi`, which in turn means we should not need any custom U-Boot commands at all. But one step at a time, I guess, just curious.

mender-image-prepare.sh
kacf: Indeed. See also my comment about GRUB above.

mender-image-prepare.sh
```diff
@@ -0,0 +1,535 @@
...
+                        obligatory parameters: --image
+
+        make_sdcard   - formats, partitions and sets the sd card
+                        obligatory parameters: --image & --rootsize
+                        
+                        note: rootsize can found as an output from
+                        'prepare_image' command
+
+        make_sdimg    - composes image file compliant with Mender partition
+                        table scheme
```
kacf: I didn't really understand the difference between `make_sdcard` and `make_sdimg`. Is one for the image file, and one for the physical card?

mender-image-prepare.sh
```diff
@@ -0,0 +1,535 @@
...
+EOF
+}
+
+if [ $# -eq 0 ]; then
+  show_help
+  exit 1
+fi
+
+if ! id | grep -q root; then
+        echo "must be run as root"
```
mirzak: I am not really comfortable to require to run the whole script as root.  Things like this become very dangerous:

    rm -rf $BASE_DIR/mnt

Maybe just `sudo` the commands that actually require elevated privileges

mender-image-prepare.sh
apodogrocki: Exactly, just for convenience.

mender-image-prepare.sh
apodogrocki: Ok, I will rename the method name.

mender-image-prepare.sh
```diff
@@ -0,0 +1,535 @@
...
+        exit
+fi
+
+BASE_DIR=${PWD}
+
+# Erase block size 8MiB
+erase_block=8388608
```
mirzak: Should probably be possible to override above with an argument

mender-image-prepare.sh
apodogrocki: In final version I think there will be no such customized uEnv.txt file. Since there is no integration with GRUB right now I just wanted to have something working, so that the kernel and filesystem starts.

mender-image-prepare.sh
```diff
@@ -0,0 +1,535 @@
...
+# Boot partition start in sectors (512 bytes per sector).
+pboot_start=16384
```
mirzak: I would expect this value to be parsed from the original image content. Generally I believe the goal should be to have a minimal set of deviations from the original image.

mender-image-prepare.sh
apodogrocki: True, I will modify it when the whole path for Raspbian is added.

mender-image-prepare.sh
```diff
@@ -0,0 +1,535 @@
...
+# Boot partition size in sectors.
+pboot_size=$(( 16777216 / 512 ))
```
mirzak: Same here, parsed from original image.

mender-image-prepare.sh
```diff
@@ -0,0 +1,535 @@
...
+pdata_size=$(( 134217728 / 512 ))
+
+# beaglebone: 4521984
+prootfs_size=
+target_dev=
+image=
+number_of_parts=
+loopsdimg=
+
+declare -a part_names=("boot" "primary" "secondary" "data")
```
mirzak: I believe that we have refereed to the rootfs parts as "rootfsA" and "rootfsB" in other locations, especially in Yocto. Might be nice to be consistent but this is a minor.

mender-image-prepare.sh
mirzak: I kinda expected for this script only to output a `.sdimg` and not actually write it to a physical device. A bit more modular approach and the script is called "mender-image-prepare", we could have a separate script that is "mender-image-write" if we want.

Thoughts?

mender-image-prepare.sh
mirzak: I would suggest to have an argument that takes X number of files that will be copied to `/boot`.

Also we need to make sure that we extract the original content of the boot part of the original image. There could be other files there, e.g. raspberrypi has a lot of configuration and firmware files there along side the u-boot.bin

mender-image-prepare.sh
```diff
@@ -0,0 +1,535 @@
+#!/bin/bash
+
+show_help() {
+cat << EOF
+Usage: $0 [prepare_image | make_sdimg | make_sdcard] [-i|--image | -d|--device -r|--rootsize]
+
+        device - can be /dev/sd* or /dev/mmcblk*
+
+        prepare_image - shrinks existing embedded image
+                        obligatory parameters: --image
```
estenberg: Would say "mandatory" not "obligatory"

mender-image-prepare.sh
```diff
@@ -0,0 +1,535 @@
...
+
+        make_sdcard   - formats, partitions and sets the sd card
+                        obligatory parameters: --image & --rootsize
```
estenberg: mandatory

mender-image-prepare.sh
kacf: Ok, makes sense.

mender-image-prepare.sh
estenberg: would say "partition layout", we use that in docs (not scheme)

kacf: But why is an explicit operation for the physical card needed? Won't the user just convert the whole image, and then flash the entire thing in one go to the physical card?

mender-image-prepare.sh
kacf: +1

mender-image-prepare.sh
apodogrocki: These names are taken from demo image for BeagleBone Black.
With command `lsblk -o name,mountpoint,label,size,uuid` you can easily verify it.

mender-image-prepare.sh
apodogrocki: Original images have 24MB for BBB and 12MB for RP PI3 of unallocated space before boot partition. The main requirement is to provide 8MB (equal to erase sector size) alignment. This is why I decided to have just 8MB unallocated space. If there is some crucial reason why we should stick with mentioned spaces, then of course I will change it. Kristian? Mirza?

mender-image-prepare.sh
apodogrocki: @estenberg: I will use "partition layout" then.
@kacf: "explicit operation for the physical card" - we can think of removing that path later...

mender-image-prepare.sh
apodogrocki: Ok, thing to think through.

mender-image-prepare.sh
kacf: I think @mirzak is right. Originally I was thinking that we should make this match the Yocto build, but since we are going to use GRUB instead of patching U-Boot, it makes more sense to keep what's in the original image (in case the bootloader has some hardcoded offsets in it). Then our GRUB environment data will live inside the boot partition instead.

mirzak: My only motivation is to stay as close as possible to the original images, to avoid breaking anything that the images might contain. Example would be some dumb script that assumes a certain offset for some parsing and if we change the offset we break the script. I do not know if this "dumb script" exists but all deviations from the original image have the potential to break something.

It could also be that the boot ROM expects the boot-part to be at a certain hard-coded offset for the board to start.

apodogrocki: One question: does "original image" refer to Yocto build images or existing embedded images (Debian, Raspbian, etc.)?

So far this 8MiB unallocated space before boot partition comes from alignment used in Yocto builds. The same refers to all partitions sizes - they are all aligned to 8MiB.

In case of existing image for BBB there is only a single partition with 4MiB offset purposed for DOS/MBR boot sector.

For Raspbian there are two partitions: boot & rootfs. Also here an unallocated space before boot partition is 4MiB.

So please let me know whether I should track the unallocated space before boot partition individually for each existing embedded image or I can stay with 8MiB?

mender-image-prepare.sh
apodogrocki: Here I also used the boot partition sizes taken from Yocto images for BBB & RP PI3.

Should I take it from the existing embedded images?

mender-image-prepare.sh
mirzak: > One question: does "original image" refer to Yocto build images or existing embedded images (Debian, Raspbian, etc.)?

Sorry for not being clear. 

"original image" == Debian, Raspbian etc.

But also:

"original image" == "input image"

It should not really matter where the input image is from, our changes should be minimal.

> So please let me know whether I should track the unallocated space before boot partition individually for each existing embedded image or I can stay with 8MiB?

I would suggest "individually for each existing embedded", preferable extracted from input image and note open-coded.

mender-image-prepare.sh
mirzak: Yes I would suggest to take this from the "input image"

apodogrocki:
To sum up:
1. For now I will keep explicit operation for the physical card just for convenience. Can be removed later, on demand.
2. I will track unallocated space before boot partition (for double partitions images, e.g. RP PI3) or rootfs partition (for single partition images, e.g. BBB) and use the same value for newly created .sdimg file.
3. I will track the size of the boot partition for double partitions images or hard-code 16MiB if single partition image is being processed.
4. I will use 'sudo' only for commands that actually require elevated privileges.
5. All spelling comments will be taken into account.

If I overwrite the commit would the whole conversation disappear? How to apply changes and not to lose the history?

kacf:
Sounds good!

Don't worry, the conversation stays, Github fixed that some time ago. The patches that have been discussed will stay as well, even if they are changed, but you will only be able to see the snippet that is in the comment.

apodogrocki:
Latest changes:
1. Use 'sudo' only for commands that actually require elevated privileges.
2. All spelling comments fixed.
3. The size of the boot partition for double partitions images is tracked or hard-coded 16MiB if single partition image is being processed.
4. Unallocated space before boot partition (for double partitions images, e.g. RP PI3) or rootfs partition (for single partition images, e.g. BBB) is tracked and used for the newly created .sdimg file.
5. mender-install-files.sh script added - script adds all necessary Mender client file to the .sdimg file. 

To do:
1. Verify both scripts starting with embedded RP PI3 image.
2. Remove hard-coded values from 'prepare_image' function.
3. Add an option to keep the original source embedded image after shrinking.
4. Integrate GRUB specific changes with both scripts.

apodogrocki:
@kacf @drewmoseley @estenberg @DominikAdamski @mirzak 

Latest changes:
1. Add an option to keep the original source embedded image after shrinking.
2.  Add separate script which integrates GRUB specific files with .sdimg file.

ToDo:
1. Conduct the whole process of .sdimg creation with embedded RP PI3 image.
2. Remove hard-coded values from 'prepare_image' function.

mender-install-files.sh
```diff
@@ -0,0 +1,311 @@
...
+  bindir="usr/bin"
+  systemd_unitdir="lib/systemd/system"
+  localstatedir="var/lib/mender"
+  base_sbindir="sbin"
+  dataconfdir="mender"
+  databootdir="u-boot"
+
+  # Prepare 'data' partition
+  echo $sudoPW | sudo -S install -d -m 755 ${data}/${dataconfdir}
+  echo $sudoPW | sudo -S install -d -m 755 ${data}/${databootdir}
```
mirzak: The `install` commands should not to require `sudo`. This applies to the all the locations of `install`. Why the use of sudo?

If `sudo` is required after all, I would recommend do all the install steps on a "staging area" and do an rsync/copy of staging area to actual rootfs with `sudo`, to reduce number of `sudo` calls.

mender-conversion-tool.sh
```diff
@@ -0,0 +1,688 @@
...
+  echo $sudoPW | sudo -S kpartx -d /dev/"$1" &
+  echo $sudoPW | sudo -S losetup -d /dev/"$1" &
+  wait
+  rm -rf $BASE_DIR/mnt
+  echo $sudoPW | sudo -S losetup -a
+}
+
+do_prepare_image() {
+  loopdevice=($(echo $sudoPW | sudo -S losetup -f))
+  echo $sudoPW | sudo -S losetup $loopdevice $image -o $((8192*512))
```
mirzak: What is `8192*512`?

mender-conversion-tool.sh
```diff
@@ -0,0 +1,688 @@
...
+  echo "2. Erase filesystem..."
+  echo $sudoPW | sudo -S wipefs --all --force ${target_dev}?*
+  echo $sudoPW | sudo -S dd if=/dev/zero of=${target_dev} bs=1M count=100
+}
+
+process_source_image() {
+  local image_name=$1
+  subname=${image_name:0:8}
+
+  image_data="$(echo $sudoPW | sudo -S fdisk -u -l ${image_name})"
```
mirzak: This is an example where you probably do not need `sudo`

mender-conversion-tool.sh
```diff
@@ -0,0 +1,688 @@
...
+
+    if [[ part_no -eq 1 ]]; then
+      echo $sudoPW | sudo -S mkfs.vfat -n ${part_names[${part_no} - 1]} $partition
+    else
+      echo $sudoPW | sudo -S mkfs.ext4 -L ${part_names[${part_no} - 1]} -F $partition
+    fi
+  done
+}
+
+make_filesystem_sdimg() {
```
mirzak: What is the difference between this and `make_filesystem`? Seems like a lot of duplication

mender-install-files.sh
```diff
@@ -0,0 +1,311 @@
...
+  cd $BASE_DIR && sync
+
+  # Clean stuff.
+  detach_mappings
+  rm -rf $BASE_DIR/go
+  rm -rf $BASE_DIR/mnt
+  rm version artifact_info mender.conf mender.service device_type
+}
+
+read -s -p "Enter password for sudo: " sudoPW
```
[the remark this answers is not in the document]
apodogrocki: Sure, will be reworked to use 'sudo true' simply.

mender-conversion-tool.sh
```diff
@@ -0,0 +1,688 @@
...
+BASE_DIR=${PWD}
+
+# Default sector size
+sector_size=
+# Erase block size 8MiB
+erase_block=8388608
+# Boot partition start in sectors (512 bytes per sector).
+pboot_start=
+# Default 'boot' partition size in sectors: 16MiB (16777216 bytes)
+pboot_size=
```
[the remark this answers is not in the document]
apodogrocki: Sure.

grub.cfg
```diff
@@ -0,0 +1,146 @@
...
+# the 'exec tail' line above.
+### END /etc/grub.d/40_custom ###
+
+### BEGIN /etc/grub.d/41_custom ###
+if [ -f  ${config_directory}/custom.cfg ]; then
+  source ${config_directory}/custom.cfg
+elif [ -z "${config_directory}" -a -f  $prefix/custom.cfg ]; then
+  source $prefix/custom.cfg;
+fi
+### END /etc/grub.d/41_custom ###
```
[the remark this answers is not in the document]
apodogrocki: Fine for me.

mender-install-files.sh
```diff
@@ -0,0 +1,311 @@
...
+
+  echo $sudoPW | sudo -S install -m 0644 ${BASE_DIR}/mender.service ${primary}/${systemd_unitdir}
+
+  # Enable menderd service starting on boot.
+  echo $sudoPW | sudo -S ln -s /lib/systemd/system/mender.service \
+      ${primary}/etc/systemd/system/multi-user.target.wants/mender.service
+
+  echo $sudoPW | sudo -S install -m 0644 ${BASE_DIR}/mender.conf ${primary}/${sysconfdir}
+
+  echo $sudoPW | sudo -S install -m 0444 ${BASE_DIR}/server.crt ${primary}/${sysconfdir}
```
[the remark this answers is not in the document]
apodogrocki: Ok, I will add an option to provide user's certificate. If it is not provided then the demo certificate will be used. Is that fine for you?

mender-conversion-tool.sh
```diff
@@ -0,0 +1,688 @@
...
+	proc                 /proc                proc       defaults              0  0
+	devpts               /dev/pts             devpts     mode=0620,gid=5       0  0
+	tmpfs                /run                 tmpfs      mode=0755,nodev,nosuid,strictatime 0  0
+	tmpfs                /var/volatile        tmpfs      defaults              0  0
+
+	# uncomment this if your device has a SD/MMC/Transflash slot
+	#/dev/mmcblk0p1       /media/card          auto       defaults,sync,noauto  0  0
+
+	# Where the U-Boot environment resides; for devices with SD card support ONLY!
+	/dev/mmcblk0p1   /uboot               auto       defaults,sync    0  0
```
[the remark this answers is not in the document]
apodogrocki: Ok, I will change the mount-point to /boot/efi.

mender-conversion-tool.sh
```diff
@@ -0,0 +1,688 @@
...
+convert it to a Mender image by restructuring partition table and adding 
+necessary files.
+
+Usage: $0 [prepare_image | make_sdimg | make_sdcard] [options]
+
+Actions:
+
+        prepare_image - shrinks existing embedded image
+
+        make_sdcard   - formats, creates partitions and sets the sd card
```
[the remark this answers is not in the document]
apodogrocki: make_sdcard - will be removed with the next amend commit.
prepare_image - rootfs partitions in existing embedded images can be quite big; 'prepare_image' just shrinks the file system to minimize its size as much as possible; can be modified later to let the user decide;

mender-install-files.sh
apodogrocki: OK.

mender-conversion-tool.sh
apodogrocki: New method will be added with the amend commit to analyse a source embedded image. No more hard-coded values.

mender-conversion-tool.sh
apodogrocki: Right.

mender-conversion-tool.sh
apodogrocki: Will be fixed.

mender-install-files.sh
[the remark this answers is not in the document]
kacf: Yep. With a suitable warning that the certificate should not be used for production.

grub.cfg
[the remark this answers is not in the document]
kacf: Created tasks:

* https://tracker.mender.io/browse/MEN-1989
* https://tracker.mender.io/browse/MEN-1990

apodogrocki:
Recent changes include:
- full support for BBB (Grub integration)
- full support for RPI3 (patched U-Boot - currently taken from Mirza's repo)
- creating .sdimg and .mender files as well
- all recent inventory scripts added
- Grub related files built from **grub-mender-grubenv** repository
- **install_bootloader** sub-command requires _--toolchain_ option as it is not yet decided whether the script should allow users to build it on their own or not

@kacf @mirzak @drewmoseley @DominikAdamski please do a review

convert-stage-4.sh
```diff
@@ -0,0 +1,308 @@
...
+
+  sudo install -m 0644 ${output_dir}/mender.service ${primary_dir}/${systemd_unitdir}
+
+  # Enable menderd service starting on boot.
+  sudo ln -s /lib/systemd/system/mender.service \
+      ${primary_dir}/etc/systemd/system/multi-user.target.wants/mender.service
+
+  sudo install -m 0644 ${output_dir}/mender.conf ${primary_dir}/${sysconfdir}
+
+  sudo install -m 0444 ${tool_dir}/server.crt ${primary_dir}/${sysconfdir}
```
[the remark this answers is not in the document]
apodogrocki: Sure, will do that...

convert-stage-4.sh
```diff
@@ -0,0 +1,308 @@
...
+	[Install]
+	WantedBy=multi-user.target
+	EOF
+
+  cat <<- EOF > $output_dir/mender.conf
+	{
+	    "InventoryPollIntervalSeconds": 5,
+	    "RetryPollIntervalSeconds": 1,
+	    "RootfsPartA": "/dev/mmcblk0p2",
+	    "RootfsPartB": "/dev/mmcblk0p3",
```
mirzak: Above should also be configurable.

convert-stage-4.sh
```diff
@@ -0,0 +1,308 @@
...
+	    "ServerCertificate": "/etc/mender/server.crt",
+	    "ServerURL": "https://docker.mender.io",
+	    "TenantToken": "dummy",
```
mirzak: Need an option to set this as well. This is used together with Hosted Mender.

convert-stage-4.sh
```diff
@@ -0,0 +1,308 @@
...
+	[Unit]
+	Description=Mender OTA update service
+	After=systemd-resolved.service
+
+	[Service]
+	Type=idle
+	User=root
+	Group=root
+	ExecStartPre=/bin/mkdir -p -m 0700 /data/mender
+	ExecStartPre=/bin/ln -sf /etc/mender/tenant.conf /var/lib/mender/authtentoken
```
mirzak: Above is not needed and was replaced with the "TenantToken" option in mender.conf

mender-conversion-tool.sh
```diff
@@ -0,0 +1,571 @@
...
+Actions:
+
+        prepare_image - shrinks existing embedded image
+
+        make_sdimg    - composes image file compliant with Mender partition
+                        layout
+
+        install_files - installs Mender client related files
+
+        install_grub  - installs Grub related files
```
DominikAdamski: install_grub option was renamed into install_bootloader

convert-stage-4.sh
apodogrocki: Will be done in a separate commit.

convert-stage-4.sh
apodogrocki: Will be done in a separate commit.

apodogrocki:
@kacf @estenberg @drewmoseley @mirzak @DominikAdamski 
As stated during Mender Client Team meeting, I am going to merge the PR as it is.

Latest changes:
1. add option to build .sdimg in a single step
2. add parameters for: demo server ip / server url / server certificate / hosted token
3. some amendments in "help" messages

ToDo:
1. parametrize polling intervals
2. use mendersoftware/uboot-mender repository instead of Mirza's private repo
3. add code to let user build patched U-Boot (code almost ready)
4. add "init_resize.sh" to repository - at the moment it is taken from Mirza'a repo
5. review of help messages if the text is adequate and all params well described
6. integrate path for x86_64 Ubuntu

estenberg:
Woho, nice!

Please make sure there are tasks for each of these ToDos so we can prioritize and track. At least for 1, 3, 4 I don't think there are explicit tasks.

In general for these options I would also strongly support just providing a way to import mender.conf file. Otherwise this will be a neverending game. :)

PR #3, opened 2018-09-11
Switch mender-conversion-tools to use mendersoftware/uboot-mender rep…

apodogrocki:
…ository

U-Boot related file like: fw_printenv, u-boot.bin are
cross-compiled now.

boot.scr on the other hand is built with mkimage tool
created in the previous step.

Issues: MEN-2089

Changelog: None

### DIFF
--- a/rpi3-convert-stage-5.sh
+++ b/rpi3-convert-stage-5.sh
@@ -16,26 +16,64 @@
 
 application_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 output_dir=${application_dir}/output
+uboot_dir=${output_dir}/uboot-mender
 bin_dir_pi=${output_dir}/bin/raspberrypi
 sdimg_base_dir=$output_dir/sdimg
+GCC_VERSION="6.3.1"
 
 echo "Running: $(basename $0)"
 declare -a sdimgmappings
 declare -a sdimg_partitions=("boot" "primary" "secondary" "data")
 
+version() {
+  echo "$@" | awk -F. '{ printf("%d%03d%03d\n", $1,$2,$3); }'
+}
+
+# Takes following arguments:
+#
+#  $1 - ARM toolchain
 build_uboot_files() {
+  local CROSS_COMPILE=${1}-
+  local ARCH=arm
+  local branch="mender-rpi-2017.09"
+  local commit="988e0ec54"
+  local uboot_repo_vc_dir=$uboot_dir/.git
+
+  export CROSS_COMPILE=$CROSS_COMPILE
+  export ARCH=$ARCH
+
   mkdir -p $bin_dir_pi
 
-  echo -e "Downloading U-Boot related files..."
+  echo -e "Building U-Boot related files..."
 
-  wget -q -O $bin_dir_pi/boot.scr \
-    https://github.com/mirzak/mender-conversion-tools/raw/mirza/wip/bin/raspberrypi/boot.scr
-  wget -q -O $bin_dir_pi/fw_printenv \
-    https://github.com/mirzak/mender-conversion-tools/raw/mirza/wip/bin/raspberrypi/fw_printenv
+  if [ ! -d $uboot_repo_vc_dir ]; then
+    git clone https://github.com/mendersoftware/uboot-mender.git -b $branch
+    git checkout $commit
+  fi
+
+  cd $uboot_dir
+
+  make --quiet distclean
+  make rpi_3_32b_defconfig && make
+  make envtools
+
+  cat<<-'EOF' >boot.cmd
+	fdt addr ${fdt_addr} && fdt get value bootargs /chosen bootargs
+	run mender_setup
+	mmc dev ${mender_uboot_dev}
+	load ${mender_uboot_root} ${kernel_addr_r} /boot/zImage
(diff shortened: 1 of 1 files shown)


rpi3-convert-stage-5.sh
```diff
@@ -16,26 +16,60 @@
...
 declare -a sdimgmappings
 declare -a sdimg_partitions=("boot" "primary" "secondary" "data")
 
+# Takes following arguments:
+#
+#  $1 - ARM toolchain
 build_uboot_files() {
+  local CROSS_COMPILE=${1}-
+  local ARCH=arm
+  local revision="mender-rpi-2017.09"
```
mirzak: I would use a fixed revision here instead of branch.

rpi3-convert-stage-5.sh
```diff
@@ -16,26 +16,60 @@
...
+  echo -e "Building U-Boot related files..."
+
+  if [ ! -d $uboot_repo_vc_dir ]; then
+    git clone https://github.com/mendersoftware/uboot-mender.git -b $revision
+    repo_clean=1
+  fi
+
+  cd $uboot_dir
+
+  [[ repo_clean -eq 0 ]] && { make --quiet clean; make --quiet distclean; }
```
mirzak: Seems a bit unnecessary to have a conditional here. 

You can always run the the following sequence. 

```
make --quiet distclean
make rpi_3_32b_defconfig
make
make envtools
```

`distclean`will succeed even if there is nothing to clean. 

You can drop `make clean`, `distclean` target has a dependency on `clean` target.

rpi3-convert-stage-5.sh
```diff
@@ -16,26 +16,60 @@
...
+
+  make rpi_3_32b_defconfig && make
+  make envtools
+
+  cat<<-'EOF' >boot.cmd
+	fdt addr ${fdt_addr} && fdt get value bootargs /chosen bootargs
+	run mender_setup
+	mmc dev ${mender_uboot_dev}
+	load ${mender_uboot_root} ${kernel_addr_r} /boot/zImage
+	bootz ${kernel_addr_r} - ${fdt_addr}
```
mirzak: Dont you need to escape `$` here? 

It seems that you do not after I tested running the command. I am curios do you know why this is?

rpi3-convert-stage-5.sh
apodogrocki: OK

rpi3-convert-stage-5.sh
apodogrocki: Expansion is disabled by quoting the EOF-doc delimiter.

rpi3-convert-stage-5.sh
apodogrocki: Do you mean that I should add a tag to mendersoftware/uboot-mender repository to the latest commit (and checkout tagged version), or simply checkout specific commit?

mirzak: Specific commit is enough I would say.

rpi3-convert-stage-5.sh
mirzak: Cool, learnt something new :)

files/init_resize.sh
```diff
@@ -0,0 +1,108 @@
...
+
+if main; then
+  whiptail --infobox "Resized data partition. Rebooting in 5 seconds..." 20 60
+  sleep 5
+else
+  sleep 5
+  whiptail --msgbox "Could not expand filesystem, please try raspi-config.\n${FAIL_REASON}" 20 60
+fi
+
+reboot_pi
```
mirzak: Where did this come from? This was not in the initial PR and unrelated to U-boot?

apodogrocki: It is taken from you repository. Not even a single change here.

mirzak: Yeye, but it is unrelated to this commit. You message says:

"Switch mender-conversion-tools to use mendersoftware/uboot-mender repository"

Either create another commit or create a separate PR (preferred) for this. And it was not in the initial PR, which is confusing as I have reviewed certain code, which you have fixed them but now you have added (sneaked in :)) more code to review which is not related to the initial PR and the initial review. 

So please keep the changes consistent and on topic to make it more effective.

apodogrocki: Ok, commit modified so that only change in rpi3-convert-stage-5.sh is visible.
Adding init_resize.sh file will be done in a separate PR.

rpi3-convert-stage-5.sh
```diff
@@ -80,7 +118,6 @@ install_files() {
 
   sudo cp ${bin_dir_pi}/u-boot.bin ${boot_dir}/kernel7.img
   sudo cp ${bin_dir_pi}/boot.scr ${boot_dir}
-  sudo cp ${bin_dir_pi}/init_resize.sh  ${rootfs_dir}/usr/lib/raspi-config/init-resize.sh
```
mirzak: This change seems out of place, but otherwise it looks good to me.

apodogrocki: init_resize.sh was just installed/copied in two places - I removed surplus line.

apodogrocki: init_resize.sh was just installed/copied in two places - I removed surplus line. So if the rest is ok, then please resolve conversation.

PR #5, opened 2018-09-18
Drop $embeddedimage check duplicate in make_sdimg

denismosolov:
$embeddedimage is checked twice inside the do_make_sdimg()


apodogrocki:
No change required.

PR #8, opened 2018-09-19
Clean-up README

mirzak:


mirzak:
Yeah makes sense.

@apodogrocki, could you help out with an a few instructions on the quickest way to get started. Meaning least number of commands to convert an image.

mirzak:
Thanks for the reviews. 

I am looking in to adding a "quick start" section but it becomes a bit lengthy, mostly because you need to cross-compile the Mender client before you run the tool and found out that we do not have any docs on how to cross-compile the Mender client so intended to fix that first

estenberg:
OK, if it is too long then I would simply ignore my comment and refer to the docs :)

mirzak:
Ok, merging. 

But intend to come back to it when it is a bit more "quick startable"

PR #11, opened 2018-09-26
MEN- 2147 improved keep parameter handling

DominikAdamski:
If user set --keep param for conversion script, then intermediate files are saved, otherwise they are removed.

MEN-2147

### DIFF
--- a/mender-convert.sh
+++ b/mender-convert.sh
@@ -397,6 +397,10 @@ do_install_mender_to_mender_disk_image() {
     stage_4_args="${stage_4_args} -t ${tenant_token}"
   fi
 
+  if [ -n "${keep}" ]; then
+    stage_4_args="${stage_4_args} -k"
+  fi
+
   eval set -- " ${stage_4_args}"
 
   export -f create_device_maps
--- a/mender-convert.sh
+++ b/mender-convert.sh
@@ -369,6 +369,11 @@ do_make_sdimg_raspberrypi3() {
   sudo install -d -m 755 ${sdimg_primary_dir}/data
 
   set_fstab $device_type
+
+  [[ $keep = "-k" ]] && { rm -f $output_dir/boot.vfat\
+     $output_dir/cmdline.txt  $output_dir/config.txt\
+     $output_dir/rootfs.img; }
+
 }
 
 do_install_mender_to_mender_disk_image() {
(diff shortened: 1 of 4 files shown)


mender-convert.sh
```diff
@@ -332,6 +332,11 @@ do_make_sdimg_raspberrypi3() {
   sudo install -d -m 755 ${sdimg_primary_dir}/data
 
   set_fstab $device_type
+
+  [[ $keep -eq "-k" ]] && { rm -rf $output_dir/boot.vfat\
+     $output_dir/cmdline.txt  $output_dir/config.txt\
```
apodogrocki: In this place there are no cmdline.txt & config.txt files. You deal with them in stage 5 for RP PI3.

rpi3-convert-stage-5.sh
```diff
@@ -180,6 +181,9 @@ do_install_bootloader() {
 
   detach_device_maps ${sdimgmappings[@]}
 
+  [[ $keep -eq 0 ]] && { rm -rf ${output_dir}/config.txt ${output_dir}/cmdline.txt
```
apodogrocki: Separate **rm** commands with **;** sign. No need to use **recursive** for removing files.

mender-convert.sh
DominikAdamski: config.txt and cmdline.txt are also copied to output directory in stage 2 for RPi3.

rpi3-convert-stage-5.sh
DominikAdamski: Done

mender-convert.sh
apodogrocki: Right!

mender-convert.sh
```diff
@@ -332,6 +332,11 @@ do_make_sdimg_raspberrypi3() {
   sudo install -d -m 755 ${sdimg_primary_dir}/data
 
   set_fstab $device_type
+
+  [[ $keep -eq "-k" ]] && { rm -rf $output_dir/boot.vfat\
```
kacf: `-eq` is an arithmetic operator, I don't think it will work with a "-k" argument. Use `=` instead.

DominikAdamski: Done

PR #12, opened 2018-09-27
Do not override mender.conf

DominikAdamski:
Prevent artifact_info file from being overwritten when Mender artifact is created.


kacf:
I'm a little bit unsure about the context here. If you create a new artifact though, it should be possible to change things like the tenant token. I would say that the `mender.conf` *should* be changed when building an image. At least that's what we do in Yocto.

PR #14, opened 2018-10-01
mender-conversion-tool.sh: Remove apostrophes from string

nandra:
Otherwise we fail during make_artifact like:

Error: .mender and .sdimg device type not matching. Aborting.

### DIFF
--- a/mender-convert.sh
+++ b/mender-convert.sh
@@ -448,7 +448,7 @@ do_make_artifact() {
   fi
 
   # Find .sdimg file's dedicated device type.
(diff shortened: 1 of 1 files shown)


apodogrocki:
@nandra Could you please resolve conflicts? I have not observed problems with apostrophes. Please share the command you are using.

nandra:
@apodogrocki hmm strange. I was adding support for raspberrypi0 (just hacky one) and hit this issue. Let me test with latest master. Thanks.

apodogrocki:
@nandra Is this PR still valid?

nandra:
Nope let it close

PR #16, opened 2018-10-03
Go through CLI options and ensure consistency

apodogrocki:


mender-convert.sh
```diff
@@ -5,28 +5,36 @@ cat << EOF
...
 Mender conversion tool
 
-A tool for taking an existing embedded image (Debian, Ubuntu, Raspbian, etc) and
-convert it to a Mender image by restructuring partition table and adding
+A tool for taking an existing embedded image (Debian, Ubuntu, Raspbian, etc)
+and convert it to a Mender image by restructuring partition table and adding
 necessary files.
 
-Usage: $0 [prepare_image | make_sdimg | install_mender | install_bootloader | make_all] [options]
+Usage: $0 [raw-disk-image-shrink-rootfs |
```
estenberg: Since there are so many actions, would rather say

Usage: $0 ACTION [options]

mender-convert.sh
```diff
@@ -5,28 +5,36 @@ cat << EOF
...
+           raw-disk-image-create-partitions |
+           install-mender-to-mender-disk-image |
+           install-bootloader-to-mender-disk-image |
+           from-raw-disk-image] [options]
 
 Actions:
```
estenberg: If incorporating above, would uppercase "ACTIONS:"

mender-convert.sh
```diff
@@ -5,28 +5,36 @@ cat << EOF
...
-                              with Mender partition layout, having all necessary
-                              files installed
+        from-raw-disk-image                     - composes fully functional Mender
+                                                  image compliant with Mender
+                                                  partition layout, having all
+                                                  necessary files installed
+        mender-disk-image-to-artifact           - creates Mender artifact file
+                                                  from Mender image
+
+Expert actions:
```
estenberg: Same here "Expert ACTIONS:"

mender-convert.sh
apodogrocki: I prefer using COMMAND instead of ACTION. As you noticed it is more common for CLI. Even **mender-artifact** uses such notation.

mender-convert.sh
apodogrocki: Here instead of "COMMANDS" I would use "General commands:"

mender-convert.sh
apodogrocki: ...and here "Expert commands:" I see that uppercase is commonly used in the "Usage" line, afterwards lowercase can be seen. Is it fine?

apodogrocki: Oh, and there is of course follow-up commit for options names.

drewmoseley:
Looks ok to me but I'll defer to @estenberg for this one.

mender-convert.sh
estenberg: OK, let's use COMMAND! :+1:

mender-convert.sh
estenberg: :+1:

mender-convert.sh
estenberg: :+1:

estenberg:
The options here look ok. Are you done with this PR now? If not would be nice if you could let us know once it is done from your point of view, I am not sure now, because:

* Only some of the options are there, e.g. raw-disk-image, rootfs-partition-id, data-part-size-mb (and more) are missing
* Commands seems to be gone?
* It says Usage: $0 [options], should be "Usage: $0 COMMAND [options]"

However, it seems like there have been some changes in a commit not part of this PR?

apodogrocki:
Some files like _convert-stage-4.sh, bbb-convert-stage-5.sh, rpi3-convert-stage-5.sh_ can be run independently - in a way how they are called in the mender-convert.sh. This is why in _Usage_ lines there are no COMMANDS placed and number of options is limited.
Question: how the show_help() methods in those scripts should be formatted, especially the _Examples:_ section? Should it contain info how to call those scripts as stand-alone scripts?

All changes in the commits are related to the PR. Besides changing commands & options names I also changed some variables names to match the re-named options names.

Those two commits contain all necessary changes required by the [MEN-2162](https://tracker.mender.io/browse/MEN-2162).

estenberg:
@apodogrocki you are right. sorry, my fault as the github interface did not load the diff of the main file by default.

Looks great and agree it resolves MEN-2162!

estenberg:
Oh, and would not worry about the standalone scripts, at least for now. This is good as-is!

PR #18, opened 2018-10-08
Convert from raw disk image and generate Artifact

apodogrocki:


mender-convert.sh
```diff
@@ -552,31 +543,57 @@ do_mender_disk_image_to_artifact() {
...
+      prootfs_start=$rootfs_a_start
+      prootfs_size=$rootfs_a_size
+      rootfs_path=$sdimg_primary_dir
+    elif [[ $rootfs_partition_id == "secondary" ]]; then
+      prootfs_start=$rootfs_b_start
+      prootfs_size=$rootfs_b_size
+      rootfs_path=$sdimg_secondary_dir
+    fi
+
+    mender_disk_basename=$(basename -- "$mender_disk_image")
```
DominikAdamski: Please add a comment why do you extract root fs from mender image

mender-convert.sh
```diff
@@ -551,31 +543,57 @@ do_mender_disk_image_to_artifact() {
...
 
   if [ $ret -eq 0 ]; then
-    local rootfs_file=${output_dir}/rootfs.ext4
-
-    echo "Creating a ext4 file-system image from modified root file-system"
-    dd if=/dev/zero of=$rootfs_file seek=${prootfs_size} count=0 bs=512 status=none
-
-    sudo mkfs.ext4 -FF $rootfs_file -d $rootfs_path
-
-    fsck.ext4 -fp $rootfs_file
```
DominikAdamski: Why have you removed checking of extracted file system?

apodogrocki: I extract the ext4 root filesystem from Mender image. I would expect it is correct. But ok, I can add it back.

PR #20, opened 2018-10-09
Fixed checksum verification.

DominikAdamski:
We need to unmount converted image before we start creation of
mender artifact. Otherwise we get checksum mismatch.


mender-convert.sh
```diff
@@ -582,7 +582,11 @@ do_mender_disk_image_to_artifact() {
       mender_artifact=${output_dir}/${mender_disk_filename}_${artifact_name}.mender
       echo "Writing Mender artifact to: ${mender_artifact}"
 
-      #Create Mender artifact
+      sudo umount -l ${rootfs_mountpoint}
```
apodogrocki: @DominikAdamski move it even higher, to line 582 and add a comment in line 574 saying: "Mount extracted ext4 partition to verify 'artifact_info' file content."

DominikAdamski: done

PR #21, opened 2018-10-12
Bring back two-stage building process for GRUB

apodogrocki:
Two-stage process is needed to build GRUB tools (e.g. grub-mkimage)
for the host and GRUB modules and grub-editenv binary purposed
for ARM architecture.

Issues: MEN-2204

Changelog: None

### DIFF
--- a/bbb-convert-stage-5.sh
+++ b/bbb-convert-stage-5.sh
@@ -36,8 +36,8 @@ EOF
 
 tool_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 output_dir=${tool_dir}/output
+grub_dir=$output_dir/grub
 grubenv_dir=$output_dir/grubenv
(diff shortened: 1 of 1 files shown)


bbb-convert-stage-5.sh
```diff
@@ -254,8 +275,9 @@ do_install_bootloader() {
   # Clean files.
   rm -rf $output_dir/sdimg
 
-  [[ $keep -eq 0 ]] && { rm -rf $grubenv_dir $grubenv_build_dir $grub_dir; }
-  [[ $rc -ne 0 ]] && { echo -e "\nStage failure."; exit 1; } || { echo -e "\nStage done."; }
+  [[ $keep -eq 0 ]] && { rm -rf $grubenv_dir $grub_dir; }
+  [[ $rc -ne 0 ]] && { echo -e "\nStage failure."; exit 1; } \
```
DominikAdamski: why should we use exit instead of return?

PR #23, opened 2018-10-16
Set naming scheme as mender-<device-type>-<artifact-name>.[sdimg|mend…

apodogrocki:
…er|ext4]

All output images (image/artifact/rootfs) should follow the scheme
presented above. Unless the user provides own Mender image name.

Moreover the only one acceptable extension for Mender image
is 'sdimg'.

Issues: MEN-2204

Changelog: None

### DIFF
--- a/mender-convert
+++ b/mender-convert
@@ -67,7 +67,7 @@ Examples:
 
         ./mender-convert from-raw-disk-image
                 --raw-disk-image <raw_disk_image_path>
-                --mender-disk-image <mender_image_name>
+                [--mender-disk-image <mender_image_name>]
                 --device-type <beaglebone | raspberrypi3>
                 --mender-client <mender_binary_path>
                 --artifact-name release-1_1.5.0
@@ -103,9 +103,10 @@ Examples for expert actions:
 
         ./mender-convert raw-disk-image-create-partitions
                 --raw-disk-image <raw_disk_image_path>
-                --mender-disk-image <mender_image_name>
+                [--mender-disk-image <mender_image_name>]
+                --artifact-name release-1_1.5.0
                 --device-type <beaglebone | raspberrypi3>
-                --data-part-size-mb 128
+                [--data-part-size-mb 128]
 
 	Output: repartitioned (respectively to Mender layout) raw disk image
 
@@ -244,13 +245,9 @@ do_raw_disk_image_shrink_rootfs() {
 }
 
 do_raw_disk_image_create_partitions() {
-  if [ -z "${raw_disk_image}" ]; then
-    echo "Raw disk image not set. Aborting."
-    exit 1
-  fi
-
-  if [ -z "${device_type}" ]; then
-    echo "Target device type name not set. Aborting."
+  if [ -z "$raw_disk_image" ] || [ -z "$device_type" ] || \
+     [ -z "$artifact_name" ]; then
+    show_help
     exit 1
   fi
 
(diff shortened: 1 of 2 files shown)


mirzak:
MEN-2204, which you mention in the commit message seems unrelated to this.

mender-convert
```diff
@@ -266,9 +263,21 @@ do_raw_disk_image_create_partitions() {
...
+  if [ -n "${mender_disk_image}" ]; then
+    mender_disk_basename=$(basename -- "$mender_disk_image")
+    if [[ $mender_disk_basename =~ \.sdimg$ ]]; then
+      mender_disk_image=$output_dir/$mender_disk_basename
+    else
+      mender_disk_filename="${mender_disk_basename%.*}"
+      mender_disk_image=$output_dir/${mender_disk_filename}.sdimg
+    fi
+  else
+    [ -n "$artifact_name" ] \
```
mirzak: You already have check (-z) on `artifact_name` at the beginning of the function, meaning that this check (-n) is not really needed.

mender-convert
```diff
@@ -266,9 +263,21 @@ do_raw_disk_image_create_partitions() {
 
   mkdir -p $output_dir && cd $output_dir
 
-  # In case of missing .sdimg name use the default format.
-  [ -z $mender_disk_image ] && mender_disk_image=$output_dir/mender_${device_type}.sdimg \
-                            || mender_disk_image=$output_dir/$mender_disk_image
+  # Make sure the user's given Mender image name has a correct extension.
+  # If Mender image name is not provided, then use following syntax:
+  # mender-<device_name>-<artifact_name>.sdimg
+  if [ -n "${mender_disk_image}" ]; then
```
mirzak: I am not really sure what this options gives us, beside adding unnecessary complexity. I would simply drop this option. 

The user can and must set the `artifact_name` and that should be enough as customization and separation from previous builds as we require the `artifact_name` to be unique.

mender-convert
```diff
@@ -103,9 +103,10 @@ Examples for expert actions:
 
         ./mender-convert raw-disk-image-create-partitions
                 --raw-disk-image <raw_disk_image_path>
-                --mender-disk-image <mender_image_name>
+                [--mender-disk-image <mender_image_name>]
+                --artifact-name release-1_1.5.0
                 --device-type <beaglebone | raspberrypi3>
-                --data-part-size-mb 128
+                [--data-part-size-mb 128]
```
mirzak: This change seems to be unrelated. Intentional?

mender-convert
```diff
@@ -266,9 +263,21 @@ do_raw_disk_image_create_partitions() {
...
-                            || mender_disk_image=$output_dir/$mender_disk_image
+  # Make sure the user's given Mender image name has a correct extension.
+  # If Mender image name is not provided, then use following syntax:
+  # mender-<device_name>-<artifact_name>.sdimg
+  if [ -n "${mender_disk_image}" ]; then
+    mender_disk_basename=$(basename -- "$mender_disk_image")
+    if [[ $mender_disk_basename =~ \.sdimg$ ]]; then
+      mender_disk_image=$output_dir/$mender_disk_basename
+    else
+      mender_disk_filename="${mender_disk_basename%.*}"
```
mirzak: Have you tried running the script without "--mender-disk-image" argument? It seems that you only set certain variables if  that argument is passed but these variables are used unconditionally in other locations.

mender-convert
apodogrocki: Yes, intentional. I marked --data-part-size-mb as optional just like --mender-disk-image.

mender-convert
apodogrocki: The only result of that part of the script is that the variable _mender_disk_image_ is set. In other locations only that variable is used. 
_mender_disk_image_ is used in two ways:
- when _from-raw-disk-image_ is called, then its real content is build in _raw-disk-image-create-partitions_ and is propagated to other commands
- when one of the _Expert commands_ is invoked, then the path to Mender image is passed in CLI - _Expert command_ checks whether such a file exists

mender-convert
mirzak: I do not follow, is this variable used for different things based on which command you run?

mender-convert
mirzak: I will trust you on that :)

mender-convert
apodogrocki: When _from-raw-disk-image_ is called, then only the desired name for Mender image must be provided. But when you run _mender-disk-image-to-artifact_ or expert_ commands_ like _install-mender-to-mender-disk-image_ & _install-bootloader-to-mender-disk-image_ then _mender-disk-name_ passed in the CLI must be the path. _mender_disk_image_ variable always holds the path to the Mender image.

mirzak:
Looks good

PR #26, opened 2018-10-22
Make output more user-friendly

apodogrocki:


estenberg:
Looks good!

Two comments:
* Can we refer to the log file if there is any failure? E.g. "ERROR: Bootloader compilation failed. Log file is at $build_log.
* (less important) Does the log file get overwritten when starting from scratch? If not then it could contain quite a lot of old log entries. Perhaps either overwrite it every time you start the tool again?

apodogrocki:
@estenberg 

1. in case any COMMAND fails, there is a log saying what went wrong, e.g.:
`Repartitioning of $raw_disk_image image failed.` or
`Installing Mender to Mender disk image failed.`
- each COMMAND then returns value != 0 and the log
`Check $build_log for details.`
is displayed
2. every time the _mender-convert_ is started the _build.log_ file is being overwritten

PR #39, opened 2018-11-06
Unification of file names for BeagleBone.

DominikAdamski:
Different names for each release of Debian unables upgrading from one Debian version to another. Unification of names eliminates this problem.

Schema of unification:
/boot/vmlinuz_kernel_version -> /boot/vmlinuz
/boot/dtbs/kernel_version/am335x-boneblack.dtb -> /boot/dtbs/am335x-boneblack.dtb

### DIFF
--- a/bbb-convert-stage-5.sh
+++ b/bbb-convert-stage-5.sh
@@ -65,9 +65,6 @@ get_kernel_version() {
   return 0
 }
 
-# Takes following arguments:
-#
-#  $1 - linux kernel version
 build_env_lock_boot_files() {
   log "\tBuilding boot scripts and tools."
   local grubenv_repo_vc_dir=$grubenv_dir/.git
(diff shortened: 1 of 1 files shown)


bbb-convert-stage-5.sh
```diff
@@ -215,6 +213,13 @@ install_files() {
   sudo ln -fs /sbin/fw_printenv $rootfs_dir/usr/bin/fw_printenv
   sudo ln -fs /sbin/fw_setenv $rootfs_dir/usr/bin/fw_setenv
 
+  #Create links for grub
+  local current_directory=$(pwd)
```
apodogrocki: I would avoid changing directories here. You have available full paths for mounted boot & rootfs partitions. Just use them...

DominikAdamski: I need to create relative symbolic links. Absolute symbolic links does not work under different root file system.

DominikAdamski: done

bbb-convert-stage-5.sh
```diff
@@ -215,6 +213,9 @@ install_files() {
   sudo ln -fs /sbin/fw_printenv $rootfs_dir/usr/bin/fw_printenv
   sudo ln -fs /sbin/fw_setenv $rootfs_dir/usr/bin/fw_setenv
 
+  #Create links for grub
+  sudo ln -sf /boot/dtbs/$linux_version/am335x-boneblack.dtb $rootfs_dir/boot/dtbs/am335x-boneblack.dtb
+  sudo ln -sf /boot/vmlinuz-$linux_version  $rootfs_dir/boot/vmlinuz
```
mirzak: Please use generic names here. e.g:

/boot/dtb
/boot/kernel

I would expect that we will have this problem on more boards running Debian, and no point in hard-coding it to "bbb".

mirzak: And the kernel image format will differ between boards.

DominikAdamski: done

PR #41, opened 2018-11-13
initial docker support

estenberg:
Not sure about the right place to put this, I don't see Dockerfile in other repos (e.g. integration).
@kacf maybe you know if this is appropriate?

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,52 @@
+FROM ubuntu:18.04
+
+ARG MENDER_ARTIFACT_VERSION=2.3.0
+ARG MENDER_CONVERT_VERSION=1.0.0
+
+RUN apt-get update && apt-get install -y \
+    kpartx \
+    bison \
+    flex \
+    mtools \
+    parted \
+    mtd-utils \
+    e2fsprogs \
+    u-boot-tools \
+    pigz \
+    device-tree-compiler \
+    autoconf \
+    autotools-dev \
+    libtool \
+    pkg-config \
+    python \
+# for mender-convert to run (mkfs.vfat is required for boot partition)
+    sudo \
+    dosfstools \
+# to compile U-Boot
+    bc \
+# to download gcc toolchain and mender-artifact
+    wget \
+# to extract gcc toolchain
+    xz-utils \
+# to download mender-convert and U-Boot sources
+    git
+
+# Disable sanity checks made by mtools. These checks reject copy/paste operations on converted disk images.
+RUN echo "mtools_skip_check=1" >> $HOME/.mtoolsrc
+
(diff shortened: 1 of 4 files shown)


kacf:
`integration` doesn't have any Dockerfile because it is only a user of already existing Docker containers. It's more appropriate to looks inside the services themselves, like [this one](https://github.com/mendersoftware/useradm/blob/master/Dockerfile). So yeah, at least to me this is appropriate. I'll let the mender-convert guys review the details though.

docker/README.md
```diff
@@ -0,0 +1,78 @@
...
+
+Create a directory `input` under the directory where you copied these files (`build.sh`, `run.sh`, etc.):
+
+```bash
+mkdir input
+```
+
+Put your raw disk image into `input`, e.g. `input/2018-10-09-raspbian-stretch.img`.
+Then put a Mender client binary compiled for your device into `input/mender-1.6.0-arm-linux-gnueabihf-gcc`.
+See [the Mender documentation on how to cross-compile the Mender client for your device](https://docs.mender.io/development/client-configuration/cross-compiling).
```
mirzak: Would be nice to add this to the container as well, either pre-built "mender" binary or the steps necessary to compile it.

Might be an optimization for later.

DominikAdamski: @estenberg  In my opinion we should firstly perform manual conversion inside docker, and then try to add scripts to do it in automatic way. Adding initial support for docker without checking everything carefully can cost us a lot of effort without any significant results.

estenberg: @mirzak agreed, we need to make the mender client packaging/compilation easier. This was just the first step in many. :) I was also planning to
* add support for taking everything as arguments to the container (either ARG or ENV), so we don't have to work inside the container itself
* publish it on docker hub somewhere, so users don't have to build it
* get the right Mender client for the board, like you mention. This is the most "hairy" one, as it needs to match the architecture (at least ARM vs. x86). Any help here appreciated! :)

estenberg:
@kacf thanks! looking at the useradm reference, maybe this should even be moved to the root directory (one level up)?

docker/Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+    bison \
+    flex \
+    mtools \
+    parted \
+    mtd-utils \
+    e2fsprogs \
+    u-boot-tools \
+    pigz \
+    device-tree-compiler \
+    autoconf \
```
apodogrocki: add:
autotools-dev \
libtool \
pkg-config \

estenberg:
Thanks for the comments everyone. Dependencies updated and verified it builds for both BBB and Rpi3. However, testing it on real hw is still pending, so should be considered experimental.

PR #42, opened 2018-11-13
Use fix GRUB version for kernels newer than 4.12

DominikAdamski:
Previously we have always builded the newest version of the GRUB. It was unstable because we cannot track all changes in GRUB repository. Now we build only checked version of GRUB/

### DIFF
--- a/bbb-convert-stage-5.sh
+++ b/bbb-convert-stage-5.sh
@@ -123,6 +123,8 @@ build_grub_efi() {
     # To avoid error message: "plain image kernel not supported - rebuild
     # with CONFIG_(U)EFI_STUB enabled" - use a specific commit.
     git checkout 9b37229f0 >> "$build_log" 2>&1
+  else
+    git checkout 72e80c025 >> "$build_log" 2>&1
   fi
 
   mkdir -p $grub_arm_dir


bbb-convert-stage-5.sh
```diff
@@ -123,6 +123,8 @@ build_grub_efi() {
     # To avoid error message: "plain image kernel not supported - rebuild
     # with CONFIG_(U)EFI_STUB enabled" - use a specific commit.
     git checkout 9b37229f0 >> "$build_log" 2>&1
+  else
+    git checkout fa926cb4b >> "$build_log" 2>&1
```
mirzak: Is this hash equivalent to a release tag?

DominikAdamski: No. GRUB development is characterized by long periods between releases. As far as I know only the newest changes in GRUB allow us to boot Linux on ARM. That's why we use commit ID. You can look at GRUB repo here: http://git.savannah.gnu.org/cgit/grub.git/log/

mirzak: As far as I know the 2.02 release is used in Yocto successfully on BBB. Anyway, this looks good.

kacf: Yes, but Yocto is probably building the kernel in a different way than Debian is, and it's the entry point to the kernel which is the problem, if I have understood correctly.

PR #43, opened 2018-11-14
Fixed vfat storage offset for RPi.

DominikAdamski:
The overlap was between boot partition and storage location of U-boot environment.

### DIFF
--- a/mender-convert-functions.sh
+++ b/mender-convert-functions.sh
@@ -2,8 +2,11 @@
 
 # Partition alignment value in bytes (8MB).
 partition_alignment=8388608
-# Boot partition storage offset in bytes (8MB).
-vfat_storage_offset=8388608
+# Boot partition storage offset in bytes (equal to alignment).
+vfat_storage_offset_regular=$partition_alignment
+#erase_block=12582912
+# Boot partition storage offset in bytes (alignment * 2).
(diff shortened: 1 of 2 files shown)


mender-convert-functions.sh
```diff
@@ -3,7 +3,10 @@
 # Partition alignment value in bytes (8MB).
 partition_alignment=8388608
 # Boot partition storage offset in bytes (8MB).
-vfat_storage_offset=8388608
+vfat_storage_offset_regular=8388608
+#erase_block=12582912
+# Boot partition storage offset in bytes (8MB).
+vfat_storage_offset_extended=12582912
```
mirzak: Why at 12 MB? If the partition alignment is defined as 8MB, it should be a multiple of that. Also update the comment, it still says 8MB

mirzak:
> Previously boot partition overlaps with root partition.

The overlap was between boot partition and storage location of U-boot environment.

mender-convert-functions.sh
DominikAdamski: Fixed

estenberg:
I confirm this works now with RPi3 and 2018-10-09-raspbian-stretch.img - thanks @mirzak @DominikAdamski !

PR #45, opened 2018-11-16
Make vfat partition offset dependent on alignment and U-Boot env space

DominikAdamski:

### DIFF
--- a/mender-convert-functions.sh
+++ b/mender-convert-functions.sh
@@ -1,16 +1,17 @@
 #!/bin/bash
 
-# Partition alignment value in bytes (8MB).
-partition_alignment=8388608
-# Boot partition storage offset in bytes (equal to alignment).
-vfat_storage_offset_regular=$partition_alignment
-#erase_block=12582912
-# Boot partition storage offset in bytes (alignment * 2).
-vfat_storage_offset_extended=$(($partition_alignment*2 ))
+# Partition alignment value in bytes.
+declare -i partition_alignment
+# Boot partition storage offset in bytes.
+declare -i vfat_storage_offset
+
+PART_ALIGN_4MB=4194304
+PART_ALIGN_8MB=8388608
+
 # Number of required heads in a final image.
-heads=255
+declare -i -r heads=255
 # Number of required sectors in a final image.
-sectors=63
+declare -i -r sectors=63
 
 declare -a mender_disk_partitions=("boot" "primary" "secondary" "data")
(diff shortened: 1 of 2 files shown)


DominikAdamski:
Here is my motivation:
1) Uniform boot partition layout with yocto images for BB
2) Fix broken unit tests

apodogrocki:
@mirzak Having issue with _the overlap between boot partition and storage location of U-boot environment_ fixed (we start boot partition with 16MB offest now), we have broken tests, i.e. _test_part_image.py line 107_.
Partition alignment for both BBB & RPI3 is set to 8MB. That causes the assert line:
`assert(parts_start[0] == alignment + uboot_env_size)`
fails, as the _uboot_env_size_ is set only during Yocto build.

After discussion with Kristian we came to conclusion that we can:
1. add `mender-uboot` to DISTRO_FEATURES
2. thanks to that we can set `uboot_env_size` manually
3. as there are two spaces for U-Boot environment the size is set to double partition alignment
4. that implies the boot partition has to be started with 24MB offset
5. as a result also _fw_env.config_ file must be modified

mirzak:
Note that BBB images and RPi images are different in Yocto. E.g Raspberry Pi uses a 4MB alignment, instead of the default 8 MB. That is why the current environment is 4 MB aligned, as the U-boot patches have been extracted from Yocto. Diverging from that adds another maintenance step, because changing fw_env.config is not enough, you also must change [defines in U-boot](https://github.com/mendersoftware/uboot-mender/commit/5226dbc3870fd621d8a5fcd3463e0b16a88c7698)

mender-convert-functions.sh
```diff
@@ -1,12 +1,14 @@
 #!/bin/bash
 
-# Partition alignment value in bytes (8MB).
-partition_alignment=8388608
+# Partition alignment value in bytes (4MB).
+partition_alignment=4194304
```
apodogrocki: In that way you have affected both BBB & RPI3. Alignment equal to 4MB should apply only to RPI3.

mender-convert-functions.sh
```diff
@@ -1,12 +1,14 @@
...
 # Boot partition storage offset in bytes (equal to alignment).
 vfat_storage_offset_regular=$partition_alignment
```
apodogrocki: Also here we have differentiate between BBB & RPI3. Regular is for BBB and should be 8MB, extended should be for RPI3 and equal to 12MB.

PR #49, opened 2018-11-19
remove erase_filesystem function

mirzak:
It has no users.

Changelog: None


mirzak:
@apodogrocki , please verify that there are no users. I could not find any at least.

Also this is a somewhat dangerous function anyway and glad it does not have any users :)

PR #53, opened 2018-11-28
Convert qemux86-64 image built using Yocto with mender-convert tool

apodogrocki:


kacf:
> But we are making it harder to "generalize" if we keep adding more platform specific options and execution paths....

+1

apodogrocki:
@mirzak I agree. My aim was to add support for qemu x86-64, so that we can finally test the update procedure. Some common parts has been already "defined" here. In i2 Epic there is place to make it far more 'general' I think.
@estenberg Please let me know how to proceed with that PR then.

PR #54, opened 2018-11-28
Include cross compiler

estenberg:
* Cross compile default generic ARM Mender client
* run docker-mender-convert directly (non-interactive)


estenberg:
FYI @mirzak :)

mirzak:
Good stuff! :D

PR #55, opened 2018-11-28
Move container env to root

estenberg:
* uses host's (checkout) of mender-convert directory
* simplifies versioning of mender-convert and avoids cloning it inside the container
* makes local development of mender-convert easier


mirzak:
I also had the opportunity to test this. It worked :+1:

mirzak:
@estenberg , got inspired by your work and I have two additional patches [here](https://github.com/mirzak/mender-conversion-tools/commits/compile-mender-during-image-creation). If you think they make sense we could integrate them in this PR.

estenberg:
Thanks @mirzak !

I was considering building the Mender client during the container build as well. This was my pros/cons:

Pros:
* Makes conversion faster (save maybe 1-2 minutes)
* Easier to provide client as argument (container run already has a lot of options/arguments)

Cons:
* New Mender clients released after container build time would not be available

I weighted the con (maybe too) heavily so I went for run-time building of the Mender client. But I suppose it is not that hard to update the Mender client this way; just change the Dockerfile / ENV and run ./docker-build.

So now I am in favor of your approach. :+1: 
Will add your commits!

estenberg:
Yep, tested & works!

One more think to keep in mind is that if we provide a standard mender-convert container in Docker Hub, we need to update it with new Mender client versions as well :)

estenberg:
Cherry-picked to 1.0.x as well.

PR #56, opened 2018-12-04
Add version option for mender-convert

apodogrocki:
Issues: MEN-2257

Changelog: Title

### DIFF
--- a/mender-convert
+++ b/mender-convert
@@ -133,6 +133,11 @@ Examples for expert actions:
 EOF
 }
 
+show_version() {
+  local version=$(cd "$tool_dir" && (git describe --tags --dirty --exact-match 2>/dev/null || git rev-parse --short HEAD))
+  echo "mender-convert version: $version"
+}
+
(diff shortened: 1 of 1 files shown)


mender-convert
```diff
@@ -133,6 +133,11 @@ Examples for expert actions:
 EOF
 }
 
+show_version() {
+  local version=$(git describe --tags --dirty --exact-match 2>/dev/null || git rev-parse --short HEAD)
```
kacf: I would write it like this:
```
local version=$(cd "$(dirname "$0")" && ( git describe --tags --dirty --exact-match 2>/dev/null || git rev-parse --short HEAD ) )
```
This way it will work also if called from outside of the folder.

apodogrocki: Right. Done.

estenberg:
@apodogrocki nice but two issues:
* there is no help text for this new option
* it needs to be backported to 1.0.x

Could you please follow up on these?

PR #65, opened 2018-12-21
Provide systemd service for Raspberry Pi boards to resize data partition

apodogrocki:
As the partition table has been modified 'resize2fs_once' service
no longer fulfils its purpose.
That service has been replaced with 'resizefs' systemd service
which extends data partition's size and resizes filesystem too
to fill the whole free unallocated space.

Issues: MEN-2254

Changelog: Title

### DIFF
--- a/files/init_resize.sh
+++ b/files/init_resize.sh
@@ -75,7 +75,7 @@ main () {
     reboot_pi
   fi
 
-if ! parted -m $ROOT_DEV u s resizepart $ROOT_PART_NUM $TARGET_END; then
+  if ! parted -m $ROOT_DEV u s resizepart $ROOT_PART_NUM $TARGET_END; then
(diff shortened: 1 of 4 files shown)


apodogrocki:
Hey @nandra could you please take a look?

nandra:
actually when tested rasbian-lite on rpi0 after create image with mender-convert after first boot I get prompt about resizing partition. But I didn't check if it's resized or not though.

Few comments to service file. There is dependency on mender.service (doesn't mender need data to be available already)? I've once did project where create mount service and also expand service which was run before mount (same like this approach). Just want to share that. Thanks.

apodogrocki:
@nandra Mender service indeed requires /dev/mmcblk0p4 is mounted. But resize2fs can handle it without any complains.

PR #69, opened 2019-01-15
Add flag for rootfs partition size

apodogrocki:


mirzak:
Overall looks good and a clean implementation, I do have one comment though,

You current changes only affect the partition sizes in MBR, but it does not touch the file-system sizes at all. Meaning the following result when I mount the root filesystem part after converting `2018-11-13-raspbian-stretch-lite.img`

     /dev/loop3                                  1,7G  962M  609M  62% /mnt

The file system still has the size of the input image. This is not necessarily a bad thing, I actually like this implementation because we are not filling it up with zeroes to fill up the whole allocated range (3,8G with the defaults) and this will also produce smaller Mender Artifact's. You can always run `resize2fs` on it once booted to expand the file system to all available blocks.

But I do anticipate confusion by users here, and we need to be more explicit in our "printouts". Maybe even generate a summary at the end of conversion that will clarify things up. Detailing disk, part, and file system sizes and maybe even free space on root parts.

E.g this printout is actually wrong:

```
	Creating blank Mender disk image:       
		image size: 8388608000 bytes       
		boot partition size: 46137344 bytes       
		root filesystem size: 4097835008 bytes       
		data partition size: 134217728 bytes
```

The "root filesystem size" is **not** 4097835008 bytes, the root filesystem **partition** is 4097835008 bytes.

One more thing :).

If the input image has been booted once on a Raspberry Pi, it will typically resize the root part to fill the disk which can result in huge root file system parts (mostly empty). This becomes problematic with the current approach, and I guess you need to run a "shrink" operation on it first right?

@estenberg, care to comment on the user-facing implications?

apodogrocki:
@mirzak Not expanding the filesystem, frankly speaking, was not my intention. @estenberg please let me know what desirable approach would be here: 
1. leave filesystem untouched and give user a note in documentation to resize it if necessary, or
2. add `resize2fs` command in `convert-stage-3.sh` script to do it automatically

Speaking about printout, I think it wouldn't be a problem to have there both information: rootfs partition size and filesystem size, if we choose point 1.

Running `shrink` operation on input image was my intention from the very beginning. Maybe it is not so crucial when working with original RPi3/BBB images, but as @mirzak noticed, when user resizes the rootfs partition during first boot or runtime, then we can have big images. Then `shrinking` comes in handy. If there are plans to get rid of that `expert option`, I would insist on keeping it.

mirzak:
> How is this handled in Yocto (https://docs.mender.io/1.7/artifacts/yocto-project/variables#mender_storage_total_size_mb)?
> I believe the full size is exposed when booting and we should keep it consistent, I think.

Yes, this is how it is handled by Yocto by default.

> The Mender Artifact files are compressed so extra zeros should not make them significantly larger. So I suppose the major downside is generation & installation time?

Yes also correct, and the annoying part is the installation time :). In the examples I earlier presented the actual file-system size is 1,7G. Expanding this with the defaults will result in 3.8G, which results in writing 2.1G of zero data.

But I do not want to scope creep this task, the intention was expanding rootfs so lets do that.

estenberg:
:+1:

apodogrocki:
@mirzak @estenberg ok, so I will do an amendment to invoke `resize2fs` so that resizing partition is followed by resizing filesystem too.

mirzak:
I was just writing this as you merged.

The addition of `resize2fs` does not seem to have helped, I tested a conversion, the size of the filesystem is still not expanded.

```
2018-11-13-raspbian-stretch-lite.sdimg2        114688  8118271 8003584  3,8G 83 Linux
```

```
$ sudo mount -o loop,offset=$((114688*512)) 2018-11-13-raspbian-stretch-lite.sdimg /mnt
```

```
/dev/loop2      1,7G  962M  609M  62% /mnt
```

apodogrocki:
@mirzak Strange, I've checked the result image by mounting and `df -h` output confirmed filesystem being expanded. I will verify it once again.

apodogrocki:
Hmm, I checked last Jenkins log - no such entry coming from `convert-stage-3.sh` script. Unfortunately _Workspace_ is not available. Start checking on my side.

mirzak:
Note that I run the conversions in the docker container.

apodogrocki:
@mirzak I understand that you use docker environment added by Eystein, don't you? Locally I do not observe any issues. I will check docker and try to add `e2fsck` command before `resize2fs`? Btw, which platform you are converting, BBB or RPi3?

mirzak:
> @mirzak I understand that you use docker environment added by Eystein, don't you?

Correct. My run command,

```
./docker-mender-convert from-raw-disk-image                 \
            --raw-disk-image $RAW_DISK_IMAGE                \
            --mender-disk-image $MENDER_DISK_IMAGE          \
            --device-type $DEVICE_TYPE                      \
            --mender-client /mender                         \
            --artifact-name $ARTIFACT_NAME                  \
            --bootloader-toolchain arm-linux-gnueabihf      \
            --server-url "https://hosted.mender.io"         \
            --tenant-token $TENANT_TOKEN
```

Target is raspberrypi3

mirzak:
Did a test run with this diff:

```
diff --git a/convert-stage-3.sh b/convert-stage-3.sh
index 6587356..ec6463e 100755
--- a/convert-stage-3.sh
+++ b/convert-stage-3.sh
@@ -9,6 +9,8 @@ build_log=$output_dir/build.log
 
 sudo dd if=${output_dir}/rootfs.img of=/dev/mapper/${rootfs_mapping} bs=8M conv=sparse >> "$build_log" 2>&1
 sync
+
+sudo e2fsck -y -f /dev/mapper/${rootfs_mapping} >> "$build_log" 2>&1
 sudo resize2fs /dev/mapper/${rootfs_mapping} >> "$build_log" 2>&1
 # Check Linux ext4 file system just in case.
 sudo fsck.ext4 -fp /dev/mapper/${rootfs_mapping} >> "$build_log" 2>&1
```

Worked better.

PR #70, opened 2019-01-17
Check a Linux ext4 file system before resizing 'primary' partition

apodogrocki:
Issues: MEN-2242

Changelog: Title

### DIFF
--- a/convert-stage-3.sh
+++ b/convert-stage-3.sh
@@ -9,6 +9,8 @@ build_log=$output_dir/build.log
 
 sudo dd if=${output_dir}/rootfs.img of=/dev/mapper/${rootfs_mapping} bs=8M conv=sparse >> "$build_log" 2>&1
 sync
+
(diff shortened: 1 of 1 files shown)


apodogrocki:
@mirzak One more question: what about 'secondary' partition? Its physical size is modified, file system is left untouched. Is it update process responsibility to take care about file system when rootfs B is flashed?

mirzak:
> Is it update process responsibility to take care about file system when rootfs B is flashed?

Yes it is depended on how you generate the artifact. I have not verified that part yet, but we can double check that the .mender content has the correct size. Which I assume it has as you extract the image from "primary" right?

mirzak:
> Which I assume it has as you extract the image from "primary" right?

Checked the .mender file, it contains a resized ext4. All is good.

apodogrocki:
:+1:

PR #71, opened 2019-01-27
Fix docker-mender-convert for paths with spaces

FSMaxB:


FSMaxB:
Also there are more problems in the actual `mender-convert`, but I gave up and renamed my directories. But please look into fixing it. `mender-convert-functions.sh`'s `get_path` seems incorrect!

FSMaxB:
(as well as several uses of `get_path`)

apodogrocki:
@FSMaxB As to `get_path`, we will take a look at that.

FSMaxB:
Sorry, I just did a quick change using the GitHub web ui. I changed the commit message now and did a force push.

PR #72, opened 2019-02-04
MEN-2248: Set production/demo intervals conditionally

lluiscampos:
Select demo default intervals only when user specifies --demo-host-ip,
else use production defaults (updated in parallel in MEN-2248)

Changelog: Title

### DIFF
--- a/convert-stage-4.sh
+++ b/convert-stage-4.sh
@@ -9,7 +9,7 @@ Usage: $0 [options]
 
     Options: [-m|--mender-disk-image | -g|--mender-client | -a|--artifact-name |
               -d|--device-type | -p|--demo-host-ip | -u| --server-url |
-              -t| --tenant-token]
+              -c|--server-cert -t| --tenant-token -k|--keep -h|--help]
 
         --mender-disk-image - Mender raw disk image
         --mender-client     - Mender client binary file
@@ -18,6 +18,9 @@ Usage: $0 [options]
         --demo-host-ip      - Mender demo server IP address
         --server-url        - Mender production server url
         --server-cert       - Mender server certificate
+        --tenant-token      - Mender tenant token
+        --keep              - Keep intermediate files in output directory
+        --help              - Show help and exit
 
     Examples:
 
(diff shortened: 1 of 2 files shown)


lluiscampos:
I submitted the pull request before getting all the information from the P.O. The JIRA issue has been clarified and the changes in this PR need some rework. See updated description in MEN-2248
Basically, we want to use a new flag --demo instead of using the --demo-host-ip
Sorry for the inconvenience.

apodogrocki:
@lluiscampos Ok, so waiting for final clarification from @estenberg .

lluiscampos:
Hi @apodogrocki 
I have reworked the code following the latest description in JIRA. Please have a look again.
As I have implemented it, --demo will require --demo-host-ip but not vice-versa. I couldn't decide on how to handle all the related demo params that we can potentially add.

convert-stage-4.sh
```diff
@@ -8,13 +8,14 @@ Mender executables, service and configuration files installer.
 Usage: $0 [options]
 
     Options: [-m|--mender-disk-image | -g|--mender-client | -a|--artifact-name |
-              -d|--device-type | -p|--demo-host-ip | -u| --server-url |
+              -d|--device-type | -e|--demo | -p|--demo-host-ip | -u| --server-url |
```
apodogrocki: '-e' option is reserved in the main 'mender-convert' script for 'storage-total-size-mb' - so maybe it is worth to not mix that

mender-convert
```diff
@@ -33,7 +33,7 @@ Expert commands:
                                                   related files
 
 Options: [-r|--raw-disk-image | -m|--mender-disk-image | -s|--data-part-size-mb |
-          -d|--device-type | -p|--rootfs-partition-id | -i|--demo-host-ip |
+          -d|--device-type | -p|--rootfs-partition-id | -e|--demo | -i|--demo-host-ip |
```
apodogrocki: Conflict between '--demo' and '--storage-total-size-mb' - both use short name as '-e'

lluiscampos: Good catch, thank you.

PR #74, opened 2019-02-12
convert-stage-4: interval settings must be 'int' and not strings

mirzak:
Fixes: 151eb4d784 ("MEN-2248: Set production/demo intervals conditionally")

Changelog: None


lluiscampos:
Thank you for fixing :)

PR #76, opened 2019-02-22
Improve documentation

nandra:
- add notes about toolchain issue with rpi0w
- add note about not possible to compile on osx + windows dues filesystem case sensitivity issues

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,18 @@ After it finishes, you can find your images in the `output` directory on your ho
 
 ### Known issues
 * BeagleBone images might not convert properly using this docker envirnoment due to permission issues: `mount: /mender-convert/output/embedded/rootfs: WARNING: device write-protected, mounted read-only.`
-* Raspberrypi0w doesn't work with precompiled mender client (pls use --mender-client option to specify build one)
+* Raspberrypi0w cpu isn't armv7+ architecture (it's armv6) and because of this mender client + u-boot fw_set/getenv tools are crashing when compiled with armv7 toolchain added in docker image. Pls use this forked [repo](https://github.com/nandra/mender-conversion-tools/commits/rpi0w-toolchain) to have it properly build with other armv6 toolchain.
+* If building U-boot fails with:
+```
+     D	scripts/Kconfig
+     input in flex scanner failed
+     ....
+     include/linux/kconfig.h:4:32: fatal error: generated/autoconf.h: No such file or directory
+     #include <generated/autoconf.h>
+```
+you might be using a case-sensitive filesystem which is not supported. Case-sensitive filesystems are typically used on OSX (Mac) and Windows but you can also run in to this on Linux if running on a NTFS formatted partition. 
+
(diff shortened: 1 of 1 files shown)


lluiscampos:
Hi @nandra and thanks for the PR.

We are enforcing now commit validation in `mender-convert` repo. Your commit is missing the `Changelog` tag. Please see [CONTRIBUTING](https://github.com/mendersoftware/mender/blob/master/CONTRIBUTING.md) doc. Thank you!

nandra:
@lluiscampos thanks updated :+1:

estenberg:
@nandra thanks for the contributions!

A few comments:
* Is there a way to integrate the raspberry pi 0 support into this repository? I am not sure if we should link to a fork for every board, it would become quite unmaintainable.

* Your second note about Linux-only makes sense, may I just suggest a slight rewording: "Currently building U-Boot is only supported on Linux (as macOS and Windows have case-insensitive file systems). For more details see [this discussion](https://hub.mender.io/t/raspberry-pi-3-model-b-b-raspbian/140/10)."

mirzak:
> * building U-Boot is only supported on Linux (as macOS and Windows have case-insensitive file systems).

Where does the conclusion that this is related to case-insensitive file-systems come from?

Today had a user report the exact problem as observed on Windows & Mac, on a Linux machine. He was running NTFS, and NTFS is case-sensitive from my understanding.

What filesystem do you normally run on OSX?

nandra:
@lluiscampos We can merge somehow my changes to mender-convert but I'm not sure if it's way to go. I think best would be to provide pre-compiled bootloaders for boards.

@mirzak 
I have hands-on experience on my osx. When try to build u-boot on osxfs it fails but when on separate partition which is case insensitive then it works fine. I can imaging this case on windows also. Maybe some description can be found here: https://github.com/docker/for-mac/issues/320

mirzak:
> @mirzak
 I have hands-on experience on my osx. When try to build u-boot on osxfs it fails but when on separate > partition which is case insensitive then it works fine. I can imaging this case on windows also. Maybe some description can be found here: [docker/for-mac#320](https://github.com/docker/for-mac/issues/320)

Got it :).

Got confused, because your comment on the PR said one thing and your changes in the commit another.

> * as macOS and Windows have case-insensitive file systems

> osx + windows not working due to the issues with filesystem case sensitivity

I would reword to:

```
If building U-boot fails with:

     D	scripts/Kconfig
     input in flex scanner failed
     ....
     include/linux/kconfig.h:4:32: fatal error: generated/autoconf.h: No such file or directory
     #include <generated/autoconf.h>

you might be using a case-sensitive filesystem which is not supported. Case-sensitive filesystems are typically used on OSX (Mac) and Windows but you can also run in to this on Linux if running on a NTFS formatted partition. 

For details see this [discussion](https://hub.mender.io/t/raspberry-pi-3-model-b-b-raspbian/140/10)
```

nandra:
Updated with latest comments.

PR #77, opened 2019-03-05
Give container access to host's kernel modules

apodogrocki:
To get rid of such errors/warnings as:
- mount: could not find any device /dev/loop#
- mount: ... WARNING: device write-protected, mounted read-only
access to host's kernel modules must be added.

Changelog: Title

Issues: MEN-2255

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,6 @@ After it finishes, you can find your images in the `output` directory on your ho
 
 
 ### Known issues
-* BeagleBone images might not convert properly using this docker envirnoment due to permission issues: `mount: /mender-convert/output/embedded/rootfs: WARNING: device write-protected, mounted read-only.`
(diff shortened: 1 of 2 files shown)


estenberg:
Nitpick, but do we need both "--privileged=true" and "--cap-add=ALL", or will it suffice with one of these?

apodogrocki:
@estenberg Docker has a default list of capabilities that are kept. None of them refers to kernel modules. To have some better control over the capabilities, the operator can use --cap-add in addition to 
--privileged. Your are right that ALL is to wide, it would be enough to limit it to SYS_MODULE.
Adding appropriate change...

apodogrocki:
@mirzak Right, forgot about README.md file. Adding an amendment.

PR #78, opened 2019-03-07
Update U-boot on Raspberry Pi builds to get rid of gcc6 requirement/dependency 

mirzak:


rpi-convert-stage-5.sh
```diff
@@ -190,13 +189,6 @@ do_install_bootloader() {
     exit 1
   fi
 
-  local gcc_version=$(${bootloader_toolchain}-gcc -dumpversion)
-
-  if [ $(version $gcc_version) -ne $(version $GCC_VERSION) ]; then
```
apodogrocki: Function version() is used only here, so you could also remove the function definition.

mirzak: done

mirzak:
@nandra , if you get a chance to verify this on a rpi0w it would be great

nandra:
@mirzak I have branch (referenced in README) which do the same (see: https://github.com/nandra/mender-conversion-tools/commit/0e2807929c1e8955eb6da31051abdf25280a768d) 
I've tried stock ubuntu arm toolchain but it's also build for armv7 arch not armv6 so problems with crashes will remain.

mirzak:
Oooh yeah, was more thing about the U-boot update. But it is not possible to test in the main repo

nandra:
@mirzak ok I'll get changes and test if u-boot works and let you know.

nandra:
@mirzak for rpi0w it won't work as u-boot 2018.07 have strict requirement to be build with toolchain gcc 6.0 (see https://patchwork.ozlabs.org/patch/843954/) and raspberrypi foundation provide only 4.8.3. Option would be to make it generic compile toolchain with crosstool-ng (shouldn't take so long) and use this toolchain for building u-boot, mender and others. Problem only is that it can be specific to board (like if build for rpi0w then it cannot be used for rpi3).

mirzak:
rpi0w is a monstrosity :), lets keep it your separate fork for now. 

> generic compile toolchain with crosstool-ng

Indeed it is an option but would really like to avoid this. 

Have you tried on of the https://toolchains.bootlin.com/ toolchains? The have armv6 ones.

nandra:
@mirzak thanks. bootlin toolchain for rpi0w looks promising. Tried to install armv6 7.3.1 and I compiled fw_printenv put on board and it's not crashing. So what about installing ubuntu armv7 toolchain and then bootling for rpi0w. Then we can distinguish before building u-boot which toolchain to use for specific device. I can image also adding aarch64 if we have some 64 bit arm board. What do you think? Thx.

nandra:
Hmm there is still issue with compiling mender during docker creation :(. What about moving it to scripts instead of building it during dockerize?

mirzak:
> @mirzak thanks. bootlin toolchain for rpi0w looks promising. Tried to install armv6 7.3.1 and I compiled fw_printenv put on board and it's not crashing. So what about installing ubuntu armv7 toolchain and then bootling for rpi0w. Then we can distinguish before building u-boot which toolchain to use for specific device. I can image also adding aarch64 if we have some 64 bit arm board. What do you think? Thx.

At the end we probably need to something like this. 

But I would like to keep this out of the mender-convert tool, and hopefully we can compile the binaries somewhere else e.g in the Docker container creation process, where we today compile the mender client but then we compile U-boot inside the mender-convert tool.

Can you maybe create a ticket on http://tracker.mender.io/, and lets continue the discussions there. For this PR I will ignore it as rpi0w does not work on master anyway, and this PR wont make it worse.

nandra:
:+1:

nandra:
Done: https://tracker.mender.io/browse/MEN-2399

PR #80, opened 2019-03-22
MEN-2411: Refactor mender-convert to use make install target

lluiscampos:
When installing the client, call "make install" from source directory
and then modify the default setup with the user input (tenant token,
server url, demo...), eliminating this way the duplication on the
install recipe.

Also, setting the default build version to 2.0.0 beta

Other minor changes along the way:
- Include liblzma build dependency
- Change --mender-client parameter to be optional
- Abort on error in convert-stage-4.sh
- Bugfix in getting server_cert from command line option
- Error reporting when using unsupported device

Changelog: Title

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,6 @@ TENANT_TOKEN="<INSERT-TOKEN-FROM Hosted Mender>"
             --raw-disk-image $RAW_DISK_IMAGE                \
             --mender-disk-image $MENDER_DISK_IMAGE          \
             --device-type $DEVICE_TYPE                      \
-            --mender-client /mender                         \
             --artifact-name $ARTIFACT_NAME                  \
             --bootloader-toolchain arm-linux-gnueabihf      \
             --server-url "https://hosted.mender.io"         \
--- a/convert-stage-4.sh
+++ b/convert-stage-4.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+
 show_help() {
   cat << EOF
 
@@ -23,26 +25,23 @@ Usage: $0 [options]
         --keep              - Keep intermediate files in output directory
         --help              - Show help and exit
 
-    Examples:
-
-        ./mender-convert install-mender-to-mender-disk-image
-                --mender-disk-image <mender_image_path>
-                --device-type <beaglebone | raspberrypi3>
-                --artifact-name release-1_1.5.0
-                --demo-host-ip 192.168.10.2
-                --mender-client <mender_binary_path>
+    For examples, see: ./mender-convert --help
 
 EOF
   exit 1
 }
 
+jq_inplace() {
+  jq_args="$1"
+  dest_file="$2"
+  sudo sh -c -e "jq \"${jq_args}\" ${dest_file} > ${dest_file}.tmp && mv ${dest_file}.tmp ${dest_file}"
+}
+
 tool_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 output_dir=${tool_dir}/output
 
-mender_client_repo="https://raw.githubusercontent.com/mendersoftware/mender"
-mender_client_revision="1.6.x"
 meta_mender_repo="https://raw.githubusercontent.com/mendersoftware/meta-mender"
-meta_mender_revision="sumo"
+meta_mender_revision="thud"
 
 mender_dir=$output_dir/mender
 device_type=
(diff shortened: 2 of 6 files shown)


lluiscampos:
This PR is on hold. Before continuing, we need to migrate the build script from inside Jenkins to the repo.

The reason is that we need to update the mender client being built to 2.0.0b1, and that is done inside such script (btw, the version being built at the moment is 1.5.0)

convert-stage-4.sh
```diff
@@ -25,24 +27,27 @@ Usage: $0 [options]
 
     Examples:
 
-        ./mender-convert install-mender-to-mender-disk-image
+        ./mender-convert from_raw_disk_image
```
apodogrocki: `install-mender-to-mender-disk-image` is one of the so called expert options in `mender-convert` tool. These expert options recently have been hidden by @mirzak. The code hasn't been removed, but they are not listed in `--help` message any more. Besides that `from-raw-disk-image` option requires `--raw-disk-image` CLI argument.
I would leave `install-mender-to-mender-disk-image`  in this place.

convert-stage-4.sh
```diff
@@ -202,41 +148,55 @@ install_files() {
...
-  sudo install -d ${primary_dir}/${sysconfdir}
-  sudo install -d ${primary_dir}/${sysconfdir}/scripts
-
   sudo ln -sf /data/${dataconfdir} ${primary_dir}/${localstatedir}
 
-  sudo install -m 0755 ${mender_client} ${primary_dir}/${bindir}/mender
-
-  sudo install -t ${primary_dir}/${identitydir} -m 0755 \
-      ${mender_dir}/mender-device-identity
+  # Call mender make install target
```
apodogrocki: Maybe you should check if golang is available, mender is downloaded and GOPATH set? It is all done in the docker conversion, but what if one runs the `mender-convert` simply from the shell?

README.md
```diff
@@ -72,7 +72,6 @@ TENANT_TOKEN="<INSERT-TOKEN-FROM Hosted Mender>"
             --raw-disk-image $RAW_DISK_IMAGE                \
             --mender-disk-image $MENDER_DISK_IMAGE          \
             --device-type $DEVICE_TYPE                      \
-            --mender-client /mender                         \
```
apodogrocki: What about adding brackets to emphasize it is optional argument?

mender-convert
```diff
@@ -64,7 +64,6 @@ Examples:
                 --raw-disk-image <raw_disk_image_path>
                 [--mender-disk-image <mender_image_name>]
                 --device-type <beaglebone | raspberrypi3>
-                --mender-client <mender_binary_path>
```
apodogrocki: The same here - just add brackets for optional argument.

convert-stage-4.sh
lluiscampos: I don't understand, `install-mender-to-mender-disk-image` option seems to have been completely removed. Looking at the `mender-convert` main script, I can see that passing such option will just show help and exit. Then I think we should just delete this example from the documentation

@mirzak can you please clarify?

convert-stage-4.sh
lluiscampos: Good point. I will add the checks.

README.md
lluiscampos: This part in the README is meant for users to copy/paste it. I don't think it is a good idea with brackets here.

mender-convert
lluiscampos: Here yes! It makes sense to leave it with brackets :+1:

convert-stage-4.sh
mirzak: The user facing options have been removed, https://github.com/mendersoftware/mender-convert/commit/024ee0efd91431249e0eab21e6b7217f6f0a4f1d

The underlying code is still  there because it is used by the `fram_raw_disk_image` execution path, but there is no way to call that method invoking the `mender-convert` tool.

lluiscampos: Correct. Then I will remove the example from the help text

lluiscampos:
Thank you for your comments @apodogrocki I replied and reworked accordingly. I also prepared a PR in mender-qa and asked for your input there.

Also, keep in mind that we need to merge this PR before the next `mender-convert` release, otherwise it won't work for mender client 2.0.0

lluiscampos:
Thanks for your review.

We will wait few hours for the final 2.0.0 tag before merging it.

lluiscampos:
The final tag will be delayed a bit. Merging this work now.

PR #81, opened 2019-03-22
remove --demo flag dependecy on --demo-host-ip

mirzak:
"--demo" should be possible to use with any server type.

Changelog: Allow to use "--demo" flag with any given server type


lluiscampos:
I agree. This way will be consistent with Yocto configuration :+1:

PR #83, opened 2019-03-26
Use a toolchain tuned for ARMv6 architecture to maintain support for …

apodogrocki:
…Raspberry Pi Zero

ARMv7 is backward compatible with ARMv6, so binaries compiled for ARMv6 should also work on ARMv7.

Changelog: Title

Issue: MEN-2399

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,50 +29,57 @@ RUN apt-get update && apt-get install -y \
 # to download mender-artifact
     wget \
 # to download mender-convert and U-Boot sources
-    git \
-# To compile Mender client, U-boot or GRUB
-    gcc-arm-linux-gnueabihf
+    git
 
 # Disable sanity checks made by mtools. These checks reject copy/paste operations on converted disk images.
 RUN echo "mtools_skip_check=1" >> $HOME/.mtoolsrc
 
+# To provide support for Raspberry Pi Zero W a toolchain tuned for ARMv6 architecture must be used.
+# https://tracker.mender.io/browse/MEN-2399
+# Assumes $(pwd) is /
+RUN wget -nc -q https://toolchains.bootlin.com/downloads/releases/toolchains/armv6-eabihf/tarballs/armv6-eabihf--glibc--stable-2018.11-1.tar.bz2 \
+    && tar -xjf armv6-eabihf--glibc--stable-2018.11-1.tar.bz2 \
+    && rm armv6-eabihf--glibc--stable-2018.11-1.tar.bz2 \
+    && echo export PATH=$PATH:/armv6-eabihf--glibc--stable-2018.11-1/bin >> /root/.bashrc
+
 RUN wget -q -O /usr/bin/mender-artifact https://d1b0l86ne08fsf.cloudfront.net/mender-artifact/$MENDER_ARTIFACT_VERSION/mender-artifact \
     && chmod +x /usr/bin/mender-artifact
-
-# Build liblzma from source
-RUN wget -q https://tukaani.org/xz/xz-5.2.4.tar.gz \
-    && tar -C /root -xzf xz-5.2.4.tar.gz \
-    && cd /root/xz-5.2.4 \
-    && ./configure --host=arm-linux-gnueabihf --prefix=/root/xz-5.2.4/install \
-    && make \
-    && make install
-
-ENV LIBLZMA_INSTALL_PATH "/root/xz-5.2.4/install"
 
 # Golang environment, for cross-compiling the Mender client
 RUN wget https://dl.google.com/go/go$GOLANG_VERSION.linux-amd64.tar.gz \
     && tar -C /usr/local -xzf go$GOLANG_VERSION.linux-amd64.tar.gz \
     && echo export PATH=$PATH:/usr/local/go/bin >> /root/.bashrc
 
-ARG mender_client_version
-
-ENV MENDER_CLIENT_VERSION=$mender_client_version
-
-# NOTE: we are assuming generic ARM board here, needs to be extended later
-
-ENV PATH "$PATH:/usr/local/go/bin"
+ENV PATH "$PATH:/usr/local/go/bin:/armv6-eabihf--glibc--stable-2018.11-1/bin"
 ENV GOPATH "/root/go"
+
+# Download Mender client
+ARG mender_client_version
+ENV MENDER_CLIENT_VERSION=$mender_client_version
 
 RUN go get -d github.com/mendersoftware/mender
 WORKDIR $GOPATH/src/github.com/mendersoftware/mender
 RUN git checkout $MENDER_CLIENT_VERSION
 
+ENV CC "arm-buildroot-linux-gnueabihf-gcc"
+
+# Build liblzma from source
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -60,17 +68,20 @@ ENV MENDER_CLIENT_VERSION=$mender_client_version
...
 
-ENV PATH "$PATH:/usr/local/go/bin"
+ENV PATH "$PATH:/usr/local/go/bin:/armv6-eabihf--glibc--stable-2018.11-1/bin"
 ENV GOPATH "/root/go"
 
 RUN go get -d github.com/mendersoftware/mender
 WORKDIR $GOPATH/src/github.com/mendersoftware/mender
 RUN git checkout $MENDER_CLIENT_VERSION
 
+ENV CC "arm-buildroot-linux-gnueabihf-gcc"
```
lluiscampos: Good.
But we need to be consistent now and use only one toolchain. So I think we need to:
- Remove gcc-arm-linux-gnueabihf from the apt-get packages to install
- Modify the --host tag in the ./configure command for liblzma
- Delete the extra CC param in the mender make build call

Dockerfile
```diff
@@ -60,17 +68,20 @@ ENV MENDER_CLIENT_VERSION=$mender_client_version
...
+
 RUN env CGO_ENABLED=1 \
     CGO_CFLAGS="-I${LIBLZMA_INSTALL_PATH}/include" \
     CGO_LDFLAGS="-L${LIBLZMA_INSTALL_PATH}/lib" \
     CC=arm-linux-gnueabihf-gcc \
+    CC=$CC \
     GOOS=linux \
     GOARCH=arm make build
```
lluiscampos: We should add the parameter `GOARM=5`, same way we cross-compiling in `mender-dist-packages`. Or at least v6, else Go could potentially decide to use v7 instructions

README.md
```diff
@@ -89,7 +95,6 @@ After it finishes, you can find your images in the `output` directory on your ho
 
 
 ### Known issues
```
lluiscampos: Shoudn't we comment [this issue](https://tracker.mender.io/browse/MEN-2399?focusedCommentId=96474&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-96474)?

Dockerfile
apodogrocki: @lluiscampos I assume that Go wouldn't manage to use v7 instructions as `arm-buildroot-linux-gnueabihf-gcc` is tuned for v6, so you can set GOARM to whatever value, but it will be ignored.

Dockerfile
apodogrocki: Amendment on its way. Making PR during Sprint review is not recommended.

README.md
apodogrocki: Short note also added.

Dockerfile
kacf: gcc is only used for the C parts of the client, the Go native compiler is still used for the Go parts. So indeed `GOARM` should be set to 6 to match what gcc will produce.

lluiscampos:
Discussed with @kacf in person

> IMO, I think we should get the change in the beta release. If there is any issue we can catch it between the beta and the final release. If we don't get it in the beta, we should then block the merge until the final release is out.

We both agree in getting this change into the beta release

> But this also brings me to another question, should we do the same tool-chain change in `mender-dist-packages` repo and ship the .deb package to be compatible with Raspberry Pi 0 W?
> 

I will follow up on this point out of this PR

Dockerfile
lluiscampos: Yes, I think it is ignored as the linker will not support it anyway.

But I still think better to set it up for v6/v5 in case this triggers errors in future releases of Go or something.

kacf:
> IMO, I think we should get the change in the beta release. If there is any issue we can catch it between the beta and the final release. If we don't get it in the beta, we should then block the merge until the final release is out.

What you should also do is to branch 1.1.x once you tag the beta, so that you can keep merging new stuff into master without risking the stability of the 1.1.0 final release. Fixes for 1.1.0 will then need to be cherry-picked as we do with our other components.

Dockerfile
```diff
@@ -29,21 +29,27 @@ RUN apt-get update && apt-get install -y \
 # to download mender-artifact
     wget \
 # to download mender-convert and U-Boot sources
-    git \
```
lluiscampos: We need git to checkout the specific version to build. See line 74

Dockerfile
```diff
@@ -29,21 +29,27 @@ RUN apt-get update && apt-get install -y \
...
+
 RUN wget -q -O /usr/bin/mender-artifact https://d1b0l86ne08fsf.cloudfront.net/mender-artifact/$MENDER_ARTIFACT_VERSION/mender-artifact \
     && chmod +x /usr/bin/mender-artifact
 
 # Build liblzma from source
 RUN wget -q https://tukaani.org/xz/xz-5.2.4.tar.gz \
     && tar -C /root -xzf xz-5.2.4.tar.gz \
     && cd /root/xz-5.2.4 \
-    && ./configure --host=arm-linux-gnueabihf --prefix=/root/xz-5.2.4/install \
+    && ./configure --host=arm-buildroot-linux-gnueabihf-gcc --prefix=/root/xz-5.2.4/install \
```
lluiscampos: The `-gcc` is not necessary here, only the prefix of the toolchain

Honestly, I think this --host option is ignored...

Dockerfile
lluiscampos: Hehe. No problem. This part is good now

README.md
lluiscampos: I can't see it

Dockerfile
apodogrocki: Ok, GOARM=6 added.

README.md
apodogrocki: Please check now.

Dockerfile
apodogrocki: Please check now.

Dockerfile
lluiscampos: :+1:

README.md
```diff
@@ -89,7 +95,10 @@ After it finishes, you can find your images in the `output` directory on your ho
...
-* Raspberrypi0w cpu isn't armv7+ architecture (it's armv6) and because of this mender client + u-boot fw_set/getenv tools are crashing when compiled with armv7 toolchain added in docker image. Pls use this forked [repo](https://github.com/nandra/mender-conversion-tools/commits/rpi0w-toolchain) to have it properly build with other armv6 toolchain.
```
mirzak: We should re-word this and keep a note about RPi0W not working due to, https://tracker.mender.io/browse/MEN-2436

Dockerfile
lluiscampos: :+1:

README.md
apodogrocki: Done.

lluiscampos:
As a follow-up for the problems that we have seen in this PR, I created [this ticket](https://tracker.mender.io/browse/MEN-2437) to catch the errors on time in future PRs

PR #84, opened 2019-03-26
Simplify docker usage

ryansch:
Changelog: Simplify docker usage by moving the source code into the
container image instead of bind mounting
Signed-off-by: Ryan Schlesinger <ryan@ryanschlesinger.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,10 +45,11 @@ This will create a container image you can use to run mender-convert.
 
 ### Use the mender-convert container image
 
-Create a directory `input` under the directory where you copied these files (`docker-build`, `docker-mender-convert`, etc.):
+Create the directories `input` and `output` under the directory where you copied these files (`docker-build`, `docker-mender-convert`, etc.):
 
 ```bash
 mkdir input
+mkdir output
 ```
 
 Then put your raw disk image into `input/`, e.g.
@@ -90,18 +91,6 @@ After it finishes, you can find your images in the `output` directory on your ho
 
 ### Known issues
 * Raspberrypi0w cpu isn't armv7+ architecture (it's armv6) and because of this mender client + u-boot fw_set/getenv tools are crashing when compiled with armv7 toolchain added in docker image. Pls use this forked [repo](https://github.com/nandra/mender-conversion-tools/commits/rpi0w-toolchain) to have it properly build with other armv6 toolchain.
-* If building U-boot fails with:
-```
-     D	scripts/Kconfig
-     input in flex scanner failed
-     ....
-     include/linux/kconfig.h:4:32: fatal error: generated/autoconf.h: No such file or directory
-     #include <generated/autoconf.h>
-```
-you might be using a case-sensitive filesystem which is not supported. Case-sensitive filesystems are typically used on OSX (Mac) and Windows but you can also run in to this on Linux if running on a NTFS formatted partition. 
-
-For details see this [discussion](https://hub.mender.io/t/raspberry-pi-3-model-b-b-raspbian/140/10)
-
 
 ## Contributing
 
--- a/docker-mender-convert
+++ b/docker-mender-convert
@@ -8,7 +8,8 @@ MENDER_CONVERT_DIR="$(pwd)"
 mkdir -p output
 
 docker run \
-       --mount type=bind,source="$MENDER_CONVERT_DIR,target=/mender-convert" \
+       -v $MENDER_CONVERT_DIR/input:/mender-convert/input \
+       -v $MENDER_CONVERT_DIR/output:/mender-convert/output \
        --privileged=true \
        --cap-add=SYS_MODULE \
        -v /dev:/dev \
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -90,18 +91,6 @@ After it finishes, you can find your images in the `output` directory on your ho
 
 ### Known issues
 * Raspberrypi0w cpu isn't armv7+ architecture (it's armv6) and because of this mender client + u-boot fw_set/getenv tools are crashing when compiled with armv7 toolchain added in docker image. Pls use this forked [repo](https://github.com/nandra/mender-conversion-tools/commits/rpi0w-toolchain) to have it properly build with other armv6 toolchain.
-* If building U-boot fails with:
```
ryansch: This PR is a result of me hitting this issue on a macOS host and fixing it.

estenberg:
Hi @ryansch - thanks for your contribution!

This is actually how it used to be, take a look at this commit: https://github.com/mendersoftware/mender-convert/commit/f570575f41b80738ab24d792bf8eb9b77e202fe6

We'll think a bit more about this and get back to you.

ryansch:
It sounds like you want two different Dockerfiles.  One for local development and one for end users to actually use `mender-convert`.

docker-mender-convert
```diff
@@ -8,7 +8,8 @@ MENDER_CONVERT_DIR="$(pwd)"
 mkdir -p output
 
 docker run \
-       --mount type=bind,source="$MENDER_CONVERT_DIR,target=/mender-convert" \
+       -v $MENDER_CONVERT_DIR/input:/mender-convert/input \
+       -v $MENDER_CONVERT_DIR/output:/mender-convert/output \
```
kacf: > Simplify docker usage by moving the source code into the container image instead of bind mounting

I don't understand.. How does this simplify things? You are making the change in the `docker-mender-convert` file, which is in the repository. So the repository is definitely available when calling this. Why not just mount the whole thing? Wouldn't you expect the currently checked out version of `mender-convert` to be called?

lluiscampos: I think it is a workaround that @ryansch has found to work for him to avoid the U-boot build failure in case sensitive filesystems (Mac in his case). So copying instead of bind mounting.

But for us seems to create a conflict, as we actively decided to do the opposite. The reason to mount instead of copy was to have an easier development process, right?

IMO it would be okey with the copy approach, but I was not here when we decided to do otherwise, so I leave it in your hands.

kacf: Oh, so it's because of the rest of the source. Yeah, I did not see that link.

I don't think we should change `Dockerfile`, since we don't expect users to be actively rebuilding the container. However, I think there is a middle ground here. Instead of doing the copy in the `Dockerfile`, just make sure we copy the U-Boot source to some other location inside the container, before building it. Any non-volume location should be free of the case sensitivity issues.

ryansch: @kacf This feels similar to my PR on mender-artifact.  Docker is an incredibly useful tool for the end user.  In my case I don't have a linux box handy (that's not a VM) so it's wonderful to be able to say `docker run` to be able to use your software.  Should I attempt to package `mender-convert` differently?  How should I be able to run this on macOS?

estenberg: @ryansch I think @kacf means that we don't have to change the mount, but rather clone & build from in a different directory (not inside the mender-convert bindmount).

mender-convert$ grep clone *
convert-stage-5.sh:    git clone https://github.com/mendersoftware/grub-mender-grubenv.git $grubenv_dir >> "$build_log" 2>&1
convert-stage-5.sh:    git clone git://git.savannah.gnu.org/grub.git $grub_dir >> "$build_log" 2>&1
rpi-convert-stage-5.sh:    git clone https://github.com/mendersoftware/uboot-mender.git -b $branch >> "$build_log" 2>&1

Perhaps we can (sym/hard) link the clone destination directories to a non-volume location in the docker environment? This can be done before running the container I believe, so that it will work (i.e. not link) when running without a container.

Does that make sense?

kacf: Hmm, I'm actually not sure I understood @estenberg's comment... :-)

This is what I mean: With the existing mount, you have `/mender-convert`. The problem is that you cannot use it as is, because of the case insensitivity issues. So just copy it to `/mender-convert-local` as part of the Docker container entrypoint script, and then perform the operation there instead, and then copy the result back to `/mender-convert`. This gives you the best of both worlds: You avoid the issues, but to the user it looks like it executed inside his original folder.

estenberg: That would work but it might not be necessary to copy all of mender-convert. The case-sensitive issues seems to happen during U-Boot compilation, and currently U-Boot is cloned into a sub-directory of mender-convert.

Rather cloning U-Boot to a non-volume location inside the container should solve it, but it does require changes to the mender-convert scripts themselves, not just the Docker build/run scripts.

kacf: I see. Given the amount of writing done during a regular invocation of mender-convert, I don't think copying the directory is a big deal though.

oleorhagen:
Closed due to inactivity

PR #88, opened 2019-04-30
Port fixes from master

lluiscampos:


estenberg:
This worked well for me!
Unrelated to the backport but there is also a MENDER_ARTIFACT_VERSION in device-image-shell/Dockerfile.

lluiscampos:
@estenberg Yes. The update of the mender-artifact was done in master, so I thought it would be ok to have it here also.

PR #91, opened 2019-05-20
always run 'image_shrink_rootfs' in 'do_from_raw_disk_image'

mirzak:
There has been a lot of confusion lately and tpyically people are
confused by these type of messages:

    *** Data partition size set to default value: 128MB ***
    1/9 Repartitioning raw disk image...
     Detected raw disk image with 2 partition(s).
     Calculating partitions' sizes of the Mender image.
     Adjust Mender disk image size to the total storage size (2048MB).
     Defined total storage size of MB is too small.
     Minimal required storage is 60056MB. Aborting.

There are two things here that are hard to understand:

- Why does it not respect the 'storage-total-size-mb' setting of 2048MB?
- Where does the 60GB requirement come frome?

This normally happens if you have booted an image once on e.g Raspbian,
which will expand the rootfs to occupy the reamining blocks of the SD
card. And if you then do a "dd" dump of that image you will get that
expanded rootfs size. We normally recommend to run
"raw-disk-image-shrink-rootfs" on these type of images before
"from-raw-disk-image" command, which will ensure that the rootfs is
reduced in size.

To remedy the confusion here, simply run the
"raw-disk-image-shrink-rootfs" unconditionally when running
"from-raw-disk-image", this way it will not matter if the input image
has been "expanded" and the '--storage-total-size-mb' will always be
respected assuming that that the actual rootfs size does not overflow
the '--storage-total-size-mb' setting.

Because 'raw-disk-image-shrink-rootfs' is now a default, we also remove
the external command for this.

Changelog: shrink expanded rootfs when running "from_raw_disk_image"

### DIFF
--- a/mender-convert
+++ b/mender-convert
@@ -150,8 +150,8 @@ do_raw_disk_image_shrink_rootfs() {
   elif [[ $raw_disk_counts -eq 2 ]]; then
     sudo losetup $loopdevice $raw_disk_image -o $((${raw_disk_sizes[prootfs_start]} * $sector_size))
   else
-    log "Error: invalid/unsupported embedded raw disk image. Aborting."
-    return 1
+    log "Warning: invalid/unsupported embedded raw disk image. Skipping resize..."
+    return 0
   fi
 
   block_size=($(sudo dumpe2fs -h $loopdevice | grep 'Block size' | tr -s ' ' | cut -d ' ' -f3))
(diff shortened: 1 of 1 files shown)


mirzak:
Hm, this will not work for "x86" it seems

mirzak:
>  But maybe for the future. :)

Yes :)

PR #93, opened 2019-05-21
sustain 'shrinking raw image' as external command for 1.1.x

apodogrocki:
'raw-disk-image-shrink-rootfs' is kept externally available.

Issues: None

Changelog: None


lluiscampos:
The Travis error seems to be due to this PR pointing to 1.1.x. I think we can ignore it.

Let's wait for the acceptance tests, though

PR #97, opened 2019-06-17
Enable U-Boot to select between compressed and uncompressed Kernels

sensslen:
Changelog: Instruct U-Boot to be able to boot either a compressed or an uncompressed kernel. This is very useful e.g. when switching from Debian (Raspbian) created using mender-convert to a Yocto based environment that does not compress the Kernel by default.
Signed-off-by: Simon Ensslen<simon.ensslen@griesser.ch>

This is the same pull Request as before but this time complies to the commiting requirements

### DIFF
--- a/rpi-convert-stage-5.sh
+++ b/rpi-convert-stage-5.sh
@@ -66,8 +66,13 @@ build_uboot_files() {
 	fdt addr ${fdt_addr} && fdt get value bootargs /chosen bootargs
 	run mender_setup
 	mmc dev ${mender_uboot_dev}
-	load ${mender_uboot_root} ${kernel_addr_r} /boot/zImage
-	bootz ${kernel_addr_r} - ${fdt_addr}
+	if load ${mender_uboot_root} ${kernel_addr_r} /boot/zImage; then
+	    bootz ${kernel_addr_r} - ${fdt_addr}
+	elif load ${mender_uboot_root} ${kernel_addr_r} /boot/uImage; then
+	    bootm ${kernel_addr_r} - ${fdt_addr}
+	else
+	    echo "No bootable Kernel found."
+	fi
 	run mender_try_to_recover
 	EOF
 


sensslen:
I'm not sure what's wrong with my commit messages...

lluiscampos:
Hi @sensslen and thanks for your contribution.

First of all, you don't need to close and re-submit. You can always fix locally and force-push your branch.

I am not sure neither why the check is failing,\; I think the tools is getting confused due to the merge. Can you try by rebasing your branch on top of master and then force-push your changes?

Your Changelog message is a bit too detailed, I suggest to keep the short summary (something similar to the commit title) as a Changelog, and the longer detail explanation and justification only in the commit message.

Thanks

lluiscampos:
@mirzak Can you have a look at this PR?

rpi-convert-stage-5.sh
```diff
@@ -66,8 +66,16 @@ build_uboot_files() {
 	fdt addr ${fdt_addr} && fdt get value bootargs /chosen bootargs
 	run mender_setup
 	mmc dev ${mender_uboot_dev}
-	load ${mender_uboot_root} ${kernel_addr_r} /boot/zImage
-	bootz ${kernel_addr_r} - ${fdt_addr}
+	if [ -f /boot/zImage ]; then
```
mirzak: This scrip is running in U-boot hush environment and I would not expect above to work at all. Have you verified this?

You probably need to do something like this,

```
if load ${mender_uboot_root} ${kernel_addr_r} /boot/zImage; then
    bootz ${kernel_addr_r} - ${fdt_addr}
elif load ${mender_uboot_root} ${kernel_addr_r} /boot/uImage; then
    bootm ${kernel_addr_r} - ${fdt_addr}
fi
```

rpi-convert-stage-5.sh
```diff
@@ -66,8 +66,16 @@ build_uboot_files() {
...
+	  load ${mender_uboot_root} ${kernel_addr_r} /boot/zImage
+	  bootz ${kernel_addr_r} - ${fdt_addr}
+	elif [ -f /boot/uImage ]; then
+	  load ${mender_uboot_root} ${kernel_addr_r} /boot/uImage
+	  bootm ${kernel_addr_r} - ${fdt_addr}
+	else
+	  echo "No bootable Kernel found. Aborting."
+	  exit 1
```
mirzak: Do not exit the script here, as it should be possible to reach `mender_try_to_recover` to detect issues caused by an update.

rpi-convert-stage-5.sh
sensslen: Probably you're right - the code commited was proposed by a collegue. I presumed he tested it but he may not. I'll look into it and come back to you

sensslen: You were right - the code was not tested - sorry for the inconvenience and thanks a lot for making sure the quality of the mender software is good!

PR #98, opened 2019-06-17
Allow mender-convert to install without server specified

sensslen:
Changelog: Add an Option to Mender Convert that allows the installation of mender without enabling the mender service and thus not requiring either Server URL and Tenant Token nor Demo Host options

Signed-off-by: Simon Ensslen<simon.ensslen@griesser.ch>

Same Pull Request as before but this time complying to the rules

### DIFF
--- a/convert-stage-4.sh
+++ b/convert-stage-4.sh
@@ -154,8 +154,11 @@ install_files() {
   fi
 
   # Enable menderd service starting on boot.
-  sudo ln -sf /lib/systemd/system/mender.service \
-      ${primary_dir}/etc/systemd/system/multi-user.target.wants/mender.service
+  if [ -z "${standalone_operation}" ]; then
+    # Enable menderd service starting on boot.
+    sudo ln -sf /lib/systemd/system/mender.service \
+        ${primary_dir}/etc/systemd/system/multi-user.target.wants/mender.service
+  fi
 
   # By default production settings configuration is installed
   if [ -n "${demo}" ] && [ ${demo} -eq 1 ]; then
@@ -212,8 +215,8 @@ do_install_mender() {
 
   if [ -z "${server_url}" ] && [ -z "${demo_host_ip}" ] && \
      [ -z "${tenant_token}" ]; then
-    log "No server type specified. Aborting."
-    show_help
+    log "No Mender server configuration was provided, it will only be possible to update using standalone mode."
+    standalone_operation="true"
   fi
 
   if [ -n "${server_url}" ] && [ -n "${demo_host_ip}" ]; then


lluiscampos:
The error seems to be the same as in the other PR.

In this one the changelog message looks good to me, though.

convert-stage-4.sh
```diff
@@ -23,6 +23,7 @@ Usage: $0 [options]
         --server-cert       - Mender server certificate
         --tenant-token      - Mender tenant token
         --keep              - Keep intermediate files in output directory
+		--no_service        - Don't enable mender service during boot
```
lluiscampos: Tabs and spaces. Please align here

convert-stage-4.sh
```diff
@@ -154,8 +155,11 @@ install_files() {
   fi
 
   # Enable menderd service starting on boot.
-  sudo ln -sf /lib/systemd/system/mender.service \
-      ${primary_dir}/etc/systemd/system/multi-user.target.wants/mender.service
+  if ! [ -z "${no_service}" ]; then
```
lluiscampos: Is this correct? I assume you want to enable the service only if no_service not set (i.e. `test -z`).

sensslen: You're completely right

lluiscampos:
Thanks.

Please rebase your branch on top of master and squash your commits into a single change set.

sensslen:
Done...

lluiscampos:
Hello!

Okey, I finally figure out the problem with your commit.

Your Git user is `chensi1 <simon.ensslen@griesser.ch>` while your sign-off is `Simon Ensslen <simon.ensslen@griesser.ch>`.

You need to set the Git user to "Simon Ensslen" and then amend the commit with `--reset-author` option.

This applies to both of your PRs

sensslen:
OK, good point. Now the build succeeds.

mender-convert
```diff
@@ -729,6 +734,10 @@ while (( "$#" )); do
     -k | --keep)
       keep="-k"
       shift 1
+      ;;
+	--no_service)
```
lluiscampos: Alignment: spaces + tabs

convert-stage-4.sh
```diff
@@ -23,6 +23,7 @@ Usage: $0 [options]
         --server-cert       - Mender server certificate
         --tenant-token      - Mender tenant token
         --keep              - Keep intermediate files in output directory
+        --no_service        - Don't enable mender service during boot
```
lluiscampos: Can we name the parameter `--no-service`? Using dash instead of underscore is a more extended practice (and already in use in the current script

mirzak:
Sorry for being late to the party, but I was hoping that we could avoid yet another flag/option.

I described my thoughts for this here, https://tracker.mender.io/browse/MEN-2590

lluiscampos:
Hi @mirzak! Sorry I missed that ticket... Your suggestion in there makes sense and IMO it is cleaner that the current PR changset.

I think we should block this PR as is.

@sensslen would you like to rework your changes as suggested by @mirzak in the ticket? Else I am afraid we will close this PR and wait for a volunteer to do MEN-2590 or us to have time for it.

lluiscampos:
FYI, the acceptance tests failed also. So the changes are not safe anyway.

lluiscampos:
Spurious failure. Merging

sensslen:
I'll look into it...

mirzak:
Thanks @sensslen. This looks good now. I have started a final run of integration tests and if all is green we will merge.

PR #100, opened 2019-06-24
parameterize target architecture when creating Docker container

mirzak:
The intention is for the defaults to match the previous functionality, that is if you run:

```
docker-build
```

Without arguments the Mender client is built for arm (32-bit).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,7 +28,9 @@ RUN apt-get update && apt-get install -y \
 # to download mender-artifact
     wget \
 # to download mender-convert and U-Boot sources
-    git
+    git \
+# for arm64 support
+    gcc-aarch64-linux-gnu
 
 # Disable sanity checks made by mtools. These checks reject copy/paste operations on converted disk images.
 RUN echo "mtools_skip_check=1" >> $HOME/.mtoolsrc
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -61,13 +63,24 @@ RUN go get -d github.com/mendersoftware/mender
 WORKDIR $GOPATH/src/github.com/mendersoftware/mender
 RUN git checkout $MENDER_CLIENT_VERSION
 
-ENV CC "arm-buildroot-linux-gnueabihf-gcc"
+# Toolchain configuration
+ARG toolchain_host
+ENV TOOLCHAIN_HOST ${toolchain_host:-arm-buildroot-linux-gnueabihf}
```
estenberg: I don't know this syntax; looks like inline if statement to assign var, similar to C?).
Just wanted to make sure that the - in front of arm-... is right?

docker-build
```diff
@@ -6,4 +6,11 @@ IMAGE_NAME=mender-convert
 
 MENDER_CLIENT_VERSION="2.0.0"
 
-docker build . -t ${IMAGE_NAME} --build-arg mender_client_version=${MENDER_CLIENT_VERSION}
+DOCKER_ARGS="--build-arg mender_client_version=${MENDER_CLIENT_VERSION}"
+
+if [ "$1" = "arm64" ]; then
```
estenberg: we are doing this toolchain mapping (arm64 -> aarch64-linux-gnu) here which I think is fine -- makes it more usable.

But then I think you should move the "default" toolchain from the Dockerfile to here as well.

Dockerfile
mirzak: This is bash - Parameter Substitution :), and yeah the `-` should be there. 

But based on you second comment I will move this to the build script instead and make these arguments mandatory.

mirzak:
Updated

estenberg:
:+1:

docker-build
```diff
@@ -6,4 +6,16 @@ IMAGE_NAME=mender-convert
...
+    DOCKER_ARGS="${DOCKER_ARGS} --build-arg toolchain_host=aarch64-linux-gnu"
+    DOCKER_ARGS="${DOCKER_ARGS} --build-arg go_flags=GOARCH=arm64"
+else
+    DOCKER_ARGS="${DOCKER_ARGS} --build-arg toolchain_host=arm-buildroot-linux-gnueabihf"
+    DOCKER_ARGS="${DOCKER_ARGS} --build-arg go_flags=\"GOARM=6 GOARCH=arm\""
+fi
+
+echo $DOCKER_ARGS
```
lluiscampos: Debug leftovers

mirzak: nice, thanks

PR #103, opened 2019-07-01
initial support for Rockpro64 with Armbian image

mirzak:
Note that this integration relies on an "standard" U-boot integration,
but the binaries are not built in mender-convert and instead it
relies on fetching pre-built binaries.

Changelog: Support for RockPro64 board

### DIFF
--- a/rockpro64-convert-stage-2.sh
+++ b/rockpro64-convert-stage-2.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+output_dir=$1
+boot_mapping=$2
+embedded_rootfs_dir=$3
+uboot_backup_dir=${embedded_rootfs_dir}/opt/backup/uboot
+build_log=$output_dir/build.log
+
+boot_part_dev="/dev/mapper/${boot_mapping}"
+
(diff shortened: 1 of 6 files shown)


rockpro64-convert-stage-2.sh
```diff
@@ -0,0 +1,18 @@
...
+output_dir=$1
+boot_mapping=$2
+embedded_rootfs_dir=$3
+uboot_backup_dir=${embedded_rootfs_dir}/opt/backup/uboot
+build_log=$output_dir/build.log
+
+[ ! -f $output_dir/boot.vfat ] && \
+    { log "Error: extracted boot partition not found. Aborting."; exit 1; }
+[ ! -d "${embedded_rootfs_dir}" ] && \
+    { log "Error: embedded content not mounted."; exit 1; }
```
lluiscampos: It would be nice to check also that `/dev/mapper/${boot_mapping}` exists

mirzak: Fixed

lluiscampos:
Great! I already approved it earlier, anyway, so feel free to merge after the tests.

PR #107, opened 2019-08-07
install /etc/mender/script/version file

mirzak:
This file needs to be present for state-scripts to work as the format
version will be validated using the information provided in this file
and the version in the Mender Artifact.

"3" means that it will only accept state-scripts created using
Mender Artifact version 3.

Changelog: Title

### DIFF
--- a/convert-stage-4.sh
+++ b/convert-stage-4.sh
@@ -56,6 +56,8 @@ server_cert=
 tenant_token=
 # Mender tenant token.
 mender_tenant_token="dummy"
+# Mender state-script format version
+mender_state_scripts_version="3"
 
 declare -a mender_disk_mappings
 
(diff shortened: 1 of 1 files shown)


lluiscampos:
The Jenkins build failed, even though GitHub didn't got any status update. I have created a new ticket [QA-93](https://tracker.mender.io/browse/QA-93) and restarted the build.

convert-stage-4.sh
```diff
@@ -130,6 +132,9 @@ install_files() {
   sudo install -d -m 755 ${data_dir}/${dataconfdir}
   sudo install -d -m 755 ${data_dir}/${databootdir}
 
+  install -d -m 755 ${primary_dir}/${sysconfdir}/scripts/
+  echo -n "${mender_state_scripts_version}" > ${primary_dir}/${sysconfdir}/scripts/version
```
lluiscampos: Okey, so the error in Jenkins **is** related to this change.

When running the script outside from Docker, you need sudo to write here. So you need to replace this with something like:
```
echo -n "${mender_state_scripts_version}" | sudo tee ${primary_dir}/${sysconfdir}/scripts/version
```

mirzak: Indeed, good tests :)

convert-stage-4.sh
```diff
@@ -130,6 +132,9 @@ install_files() {
   sudo install -d -m 755 ${data_dir}/${dataconfdir}
   sudo install -d -m 755 ${data_dir}/${databootdir}
 
+  install -d -m 755 ${primary_dir}/${sysconfdir}/scripts/
```
lluiscampos: `sudo` on the install dir also :sweat_smile:

mirzak: Doh!

mirzak:
All green now except your approval :)

mirzak:
Thank you

PR #110, opened 2019-08-09
Fix commits check and Docker build in GitLab CI

lluiscampos:
It was broken since we modified check_commits.sh few weeks ago

Changelog: None


lluiscampos:
@mzedel Please have a look at commit b20b508

PR #111, opened 2019-08-12
configs: raspberrypi: Enable Full MKS graphics driver

mirzak:
See commit for more details and additional background information
can be found here:

    https://tracker.mender.io/browse/MEN-2685

Changelog: Fix HDMI output post 2019-07-10 Raspbian releases

### DIFF
--- a/rpi-convert-stage-5.sh
+++ b/rpi-convert-stage-5.sh
@@ -138,6 +138,24 @@ install_files() {
 
   sudo cp ${bin_dir_pi}/boot.scr ${boot_dir}
 
+  # Enable Full KMS (Kernel Mode Setting) graphics driver on Buster images.
+  # This is the the open-source GPU driver that is part of mesa, also known as
+  # VC4.
+  #
+  # This is done because it seems that the firmware GPU driver does currently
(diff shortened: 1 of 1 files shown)


lluiscampos:
Can you rebase from master? The GitLab build issues should be fixed now

estenberg:
Same here, important problem but I have no clue how to fix it :)

mirzak:
Holding of merge as I found an issue with this when testing older images based on Debian Stretch.

mirzak:
After some more testing, research and thought I no longer think that this is a good way of resolving it.

This will not work on Debian Stretch images, though it would be fairly easy to add a check in the conversion script which version we are converting and depending on that set this or not but there is another problem. If the intention is to run the device in headless mode it is not recommended to have this option enabled, relates to higher power consumption and devices getting unnecessary warm.

So instead I think we should just document this on MH, with a guide how to workaround this.

mirzak:
Closing this, the workaround is documented on [MH](https://hub.mender.io/t/raspberry-pi-3-model-b-b-raspbian/140)

PR #112, opened 2019-08-12
Add platform prefix to mender-artifact download URL

lluiscampos:
Current stable version 3.0.x should be the transition, as for future
versions we will be releasing the binary in a linux/darwin folder.

Changelog: None


lluiscampos:
The acceptance tests don't use the docker build. Stopping them.

lluiscampos:
I need this for the upcoming release. Merging without formal revision.

PR #120, opened 2019-10-01
scripts: run-tests: ensure that multi-config code path is executed

mirzak:
This is a follow-up to https://github.com/mendersoftware/mender-convert/pull/119.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ work
 rootfs_overlay_demo/*
 tests
 mender_local_config
+*.xml
+*.html
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -55,7 +55,7 @@ test_acceptance:
     - docker version
     # Install dependencies
     - apk --update --no-cache add bash wget git util-linux mtools python
-      py-pip gcc python2-dev libffi-dev libc-dev openssl-dev make
+      py-pip gcc python2-dev libffi-dev libc-dev openssl-dev make sudo
     - pip install "pytest>=3.0" "Fabric>=1.13.0, <2" pytest-html
     # Load image under test
     - export IMAGE_NAME=$DOCKER_REPOSITORY:pr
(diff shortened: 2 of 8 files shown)


lluiscampos:
Also, this other PR will fix the current GitLab failures (and remove Travis): https://github.com/mendersoftware/mender-convert/pull/121

mirzak:
Found a bug regarding configs which I fixed and updated the test script a bit more to cover additional cases.

mirzak:
Almost expected it to fail with:

```
./scripts/run-tests.sh: line 94: sudo: command not found
```

Is it ok to add this to the Gitlab CI image?

lluiscampos:
> Almost expected it to fail with:
> 
> ```
> ./scripts/run-tests.sh: line 94: sudo: command not found
> ```

Right. The reason for it not to fail and abort here is that the we have the "else" condition in the function call. It seems like bash will ignore the `set -e` inside the function and just propagate at the end the exit code.

```
convert_and_test "beaglebone" \
                 "release-1" \
                 "${BBB_DEBIAN_IMAGE_URL}" \
                 "${BBB_DEBIAN_IMAGE}" \
                 "${BBB_DEBIAN_IMAGE}.xz" || test_result=$?
```

From the lines above, removing `|| test_result=$?` will make it fail and abort in the `sudo` call. But then I don't have a good way to make the script continue to the next test case... Any ideas?

TBH, I didn't know about this feature bash, and I was rather surprised. I realized yesterday in a similar situation: `wget` was failing to download the Tinker image, but still the function proceeded.
 
> Is it ok to add this to the Gitlab CI image?

Yes, please add `sudo` in the dependencies list [here](https://github.com/mendersoftware/mender-convert/blob/next/.gitlab-ci.yml#L58).

mirzak:
Can a get an ack from you @lluiscampos that this is still ok, before I merge. There has been some changes since you approved the PR.

lluiscampos:
@mirzak Thanks for double checking. Looks good, go ahead and merge.

PR #121, opened 2019-10-01
Fix GitLAb pipeline and remove Travis script

lluiscampos:
```
commit 0845892795a165dfe13d0cc845b92be5d17b65d5
Author: Lluis Campos <lluis.campos@northern.tech>
Date:   Tue Oct 1 11:13:43 2019 +0200

    Allow for the user to override the IMAGE_NAME shell variable
    
    So that GitLab CI script (or any user, really) can customize the name of
    the resulting Docker image.

commit a49d08794cbe82196aec6331c8a2fc8441bfa39b
Author: Lluis Campos <lluis.campos@northern.tech>
Date:   Tue Oct 1 11:14:39 2019 +0200

    Remove Travis script, functionality is fully migrated to GitLab

commit b37daef50d50c5fffe3b320e2d5de5f6cfa3b2d1
Author: Lluis Campos <lluis.campos@northern.tech>
Date:   Tue Oct 1 14:48:35 2019 +0200

    Remove creation of obsolete output directory
```


docker-build
```diff
@@ -16,6 +16,6 @@
 
 set -e
 
-IMAGE_NAME=mender-convert
+IMAGE_NAME=${IMAGE_NAME:-mender-convert}
```
lluiscampos: This needs to be ported also to `docker-mender-convert`

.dockerignore
```diff
@@ -1,3 +1,4 @@
 deploy
 input
 work
+output*
```
mirzak: This is actually a ghost directory which is not used anymore in `v2`. It was changed to utilize `deploy` and `work`, the  `deploy` directory contains the final images while all the temporary files are store in `work` (which is cleaned when conversion completes or fails).

So this line should be dropped instead,

https://github.com/mendersoftware/mender-convert/blob/next/docker-mender-convert#L22

lluiscampos: Thanks. Fixed

PR #122, opened 2019-10-01
MEN-2785: Acceptance tests on Pipeline

lluiscampos:
Main change:

```
MEN-2785: Add acceptance tests to GitLab pipeline
    
Created a new test_acceptance stage that will execute in our private
runners. This stage will execute scripts/run-tests.sh and collect the
now produced xml/html reports for each device type.
    
Removed the build from the test script, so that we make sure that CI
tests the exact same image build in "build" stage.
```

Other minor changes:
- Remove unnecessary check for mender-artifact
- Use volume rather than mount on Docker invocation
- Avoid using the py.test wrapper and directly call python2 -m pytest

### DIFF
--- a/scripts/run-tests.sh
+++ b/scripts/run-tests.sh
@@ -75,10 +75,15 @@ convert_and_test() {
         --disk-image input/${image_name} \
         --config ${WORKSPACE}/test_config
 
+    if pip list | grep -q -e pytest-html; then
+        html_report_args="--html=${MENDER_CONVERT_DIR}/report_${device_type}.html --self-contained-html"
+    fi
+
     cd ${WORKSPACE}/mender-image-tests
 
     python2 -m pytest --verbose \
-            --junit-xml="${WORKSPACE}/results.xml" \
+            --junit-xml="${MENDER_CONVERT_DIR}/results_${device_type}.xml" \
+            ${html_report_args} \
             --test-conversion \
             --test-variables="${MENDER_CONVERT_DIR}/deploy/${device_type}-${artifact_name}.cfg" \
             --board-type="${device_type}" \
@@ -106,8 +111,6 @@ fi
 mkdir -p ${WORKSPACE}
 
 get_pytest_files
-
-./docker-build
 
 convert_and_test "qemux86_64" \
                  "release-1" \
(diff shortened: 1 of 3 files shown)


lluiscampos:
> Can not comment much on the `.gitlab-ci.yml` changes but the rest looks good to me. I also tested running `scripts/run-tests.sh` locally and that still works. Which makes me happy :)

Here but also in many other repos, our goal is always to be able to run test locally.

> 
> We should also update the instructions here,
> 
> https://github.com/mendersoftware/mender-convert/blob/next/scripts/README-run-tests.md
> 
> But as the instructions in above file where broken before the changes made in this PR, we can follow-up on this in a separate PR

I don't really see what is broken, really. The requirements could be a bit more fine tuned (we don't need pytest-xdist) and it also asumes Ubuntu, but I don't think that is a big deal for now.

One of the ideas I have for the start of next year (after full migration to Python3), is to align the test Python requirements across all our repos with Python tests. Then I will revisit mender-convert and clean this up.

lluiscampos:
Running Pipeline again after https://github.com/mendersoftware/mender-image-tests/pull/9 has been merged

scripts/run-tests.sh
```diff
@@ -112,29 +115,32 @@ mkdir -p ${WORKSPACE}
...
                  "${TINKER_IMAGE}.img" \
-                 "${TINKER_IMAGE}.zip"
+                 "${TINKER_IMAGE}.zip" || test_result=$?
 
 convert_and_test "beaglebone" \
                  "release-1" \
                  "${BBB_DEBIAN_IMAGE_URL}" \
                  "${BBB_DEBIAN_IMAGE}" \
-                 "${BBB_DEBIAN_IMAGE}.xz"
+                 "${BBB_DEBIAN_IMAGE}.xz" || test_result=$?
```
mirzak: This will not work right? If the final test succeeds but all previous have failed, this would still set the test_result to 0.

I think we need to OR the return values, to keep the `1`

kacf: Oh, nice catch! I missed that!

lluiscampos: It works: it only sets test_result to the exit code of the previous command if the command fails. On success it does nothing

I have tested it on the CI and locally.

If more than one failed, yes, you will be getting only the last errcode, but that is okey I think.

mirzak: Thanks for clarifying. Makes sense.

PR #126, opened 2019-10-11
Server types

drewmoseley:
This makes overlay setups for all Mender server types.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,10 +37,25 @@ unzip 2019-06-20-raspbian-buster-lite.zip && cd ..
 
 Bootstrap the demo rootfs overlay that is configured to connect to
 https://hosted.mender.io with polling intervals set appropriately for
-demonstration purposes:
+demonstration purposes.  There are three scripts here to support
+the Mender demo server, the Mender production server, and hosted
+Mender.
 
 ```
-./scripts/bootstrap-rootfs-overlay-demo.sh \
+./scripts/bootstrap-rootfs-overlay-demo-server.sh \
+    --output-dir ${PWD}/rootfs_overlay_demo \
+    --server-ip 192.168.1.1
+```
+
+```
+./scripts/bootstrap-rootfs-overlay-production-server.sh \
+    --output-dir ${PWD}/rootfs_overlay_demo \
+    --server-url https://foobar.mender.io \
+    [ --server-cert ~/server.crt ]
+```
+
+```
+./scripts/bootstrap-rootfs-overlay-hosted-server.sh \
     --output-dir ${PWD}/rootfs_overlay_demo \
     --tenant-token "Paste token from Hosted Mender"
 ```
--- a/scripts/bootstrap-rootfs-overlay-demo-server.sh
+++ b/scripts/bootstrap-rootfs-overlay-demo-server.sh
@@ -23,23 +23,21 @@ if [ "${root_dir}" != "${PWD}" ]; then
     exit 1
 fi
 
-# Do not actually paste it here, this is just the default value that will
-# end up in mender.conf if no token is specified using '--tenant-token'
-tenant_token="Paste your Hosted Mender token here"
+server_ip=""
 output_dir=""
 while (( "$#" )); do
   case "$1" in
-    -t | --tenant-token)
-      tenant_token="${2}"
-      shift 2
-      ;;
     -o | --output-dir)
       output_dir="${2}"
       shift 2
       ;;
+    -s | --server-ip)
+      server_ip="${2}"
+      shift 2
+      ;;
     *)
       echo "Sorry but the provided option is not supported: $1"
-      echo "Usage:  $(basename $0) --tenant-token"
+      echo "Usage:  $(basename $0) --output-dir ./rootfs_overlay_demo --server-ip <your server IP address>"
       exit 1
       ;;
   esac
@@ -49,16 +47,31 @@ if [ -z "${output_dir}" ]; then
     echo "Sorry, but you need to provide an output directory using the '-o/--output-dir' option"
     exit 1
 fi
+if [ -z "${server_ip}" ]; then
+    echo "Sorry, but you need to provide a server IP address using the '-s/--server-ip' option"
+    exit 1
+fi
 
 mkdir -p ${output_dir}/etc/mender
 cat <<- EOF > ${output_dir}/etc/mender/mender.conf
 {
   "InventoryPollIntervalSeconds": 5,
   "RetryPollIntervalSeconds": 30,
-  "ServerURL": "https://hosted.mender.io/",
-  "TenantToken": "${tenant_token}",
+  "ServerURL": "https://docker.mender.io",
+  "ServerCertificate": "/etc/mender/server.crt",
   "UpdatePollIntervalSeconds": 5
 }
 EOF
+cat <<- EOF > ${output_dir}/etc/hosts
+127.0.0.1	localhost
 
-echo "Configuration file written to: ${output_dir}/etc/mender"
+# The following lines are desirable for IPv6 capable hosts
+::1     localhost ip6-localhost ip6-loopback
+ff02::1 ip6-allnodes
+ff02::2 ip6-allrouters
+
+${server_ip} docker.mender.io s3.docker.mender.io
+EOF
+wget -q "https://raw.githubusercontent.com/mendersoftware/mender/master/support/demo.crt" -O ${output_dir}/etc/mender/server.crt
+
+echo "Configuration file for using Demo Mender Server written to: ${output_dir}/etc/mender"
--- a/scripts/bootstrap-rootfs-overlay-production-server.sh
+++ b/scripts/bootstrap-rootfs-overlay-production-server.sh
@@ -1,0 +1,80 @@
+#!/bin/bash
+#
+# Copyright 2019 Northern.tech AS
+#
+#  Licensed under the Apache License, Version 2.0 (the "License");
+#  you may not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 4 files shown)


README.md
```diff
@@ -37,10 +37,25 @@ unzip 2019-06-20-raspbian-buster-lite.zip && cd ..
...
+    --output-dir ${PWD}/rootfs_overlay_demo \
+    --server-ip 192.168.1.1
+```
+
+```
+./scripts/bootstrap-rootfs-overlay-production-server.sh \
+    --output-dir ${PWD}/rootfs_overlay_demo \
+    --server-url https://foobar.mender.io \
+    [ --server-cert ~/server.crt ]
+```
```
mirzak: I was really hoping to avoid covering these variations here ( and providing multiple scripts), and instead I was hoping that we would cover this in docs.mender.io instead. How rootfs overlay is utilized to apply configurations, and docs.mender.io should cover (in general) the different server configuration options instead of us duplicating it everywhere.

We are also moving away of providing multiple configurations, in other parts of the docs, e.g https://docs.mender.io/2.1/client-configuration/installing. This only covers Hosted Mender as well, and maybe this is the appropriate location to update on how to handle server variants.

We do have https://tracker.mender.io/browse/MEN-2786 to complete, and once this is finished I would think these parts would crystallize.

mirzak: I think it is ok if you add the scripts for now. But I would omit the update of the README mentioning all variants.

scripts/bootstrap-rootfs-overlay-demo-server.sh
```diff
@@ -0,0 +1,70 @@
...
+      output_dir="${2}"
+      shift 2
+      ;;
+    -s | --server-ip)
+      server_ip="${2}"
+      shift 2
+      ;;
+    *)
+      echo "Sorry but the provided option is not supported: $1"
+      echo "Usage:  $(basename $0) --tenant-token"
```
mirzak: Update usage.

scripts/bootstrap-rootfs-overlay-demo-server.sh
```diff
@@ -0,0 +1,70 @@
...
+  "RetryPollIntervalSeconds": 30,
+  "ServerURL": "https://docker.mender.io",
+  "ServerCertificate": "/etc/mender/server.crt",
+  "UpdatePollIntervalSeconds": 5
+}
+EOF
+cat <<- EOF > ${output_dir}/etc/hosts.mender-convert-append
+${server_ip} docker.mender.io s3.docker.mender.io
+EOF
+wget -q "https://raw.githubusercontent.com/mendersoftware/meta-mender/master/meta-mender-demo/recipes-mender/mender/files/server.crt" -O ${output_dir}/etc/mender/server.crt
```
mirzak: You can fetch this from,  https://raw.githubusercontent.com/mendersoftware/mender/master/support/demo.crt

scripts/bootstrap-rootfs-overlay-demo-server.sh
```diff
@@ -0,0 +1,70 @@
...
+{
+  "InventoryPollIntervalSeconds": 5,
+  "RetryPollIntervalSeconds": 30,
+  "ServerURL": "https://docker.mender.io",
+  "ServerCertificate": "/etc/mender/server.crt",
+  "UpdatePollIntervalSeconds": 5
+}
+EOF
+cat <<- EOF > ${output_dir}/etc/hosts.mender-convert-append
+${server_ip} docker.mender.io s3.docker.mender.io
```
mirzak: Update to provide full `/etc/hosts`

scripts/bootstrap-rootfs-overlay-production-server.sh
```diff
@@ -0,0 +1,80 @@
...
+
+if [ -n "${server_cert}" ] ; then
+cat <<- EOF >> ${output_dir}/etc/mender/mender.conf
+  "ServerCertificate": "/etc/mender/server.crt",
+EOF
+cp -f "${server_cert}" ${output_dir}/etc/mender/server.crt
+fi
+
+cat <<- EOF >> ${output_dir}/etc/mender/mender.conf
+  "UpdatePollIntervalSeconds": 5
```
mirzak: Ah here it is :)

mender-convert-modify
```diff
@@ -219,4 +219,11 @@ platform_modify
 for overlay in "${overlays[@]}"; do
   log_info "Applying rootfs overlay: ${overlay}"
   run_and_log_cmd "sudo rsync --archive --keep-dirlinks --verbose ${overlay}/ work/rootfs/"
+
+  for append in $(cd work/rootfs; find . -name '*.mender-convert-append'); do
+      sourcefile=$(echo $append | sed 's@.mender-convert-append$@@')
+      log_info "Applying rootfs overlay append to: ${sourcefile}"
+      run_and_log_cmd "sudo cat work/rootfs/${append} >> work/rootfs/${sourcefile}"
+      run_and_log_cmd "sudo rm -f work/rootfs/${append}"
+  done
```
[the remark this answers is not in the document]
drewmoseley: Please reconsider.  We want this to be easily applied to other operating systems and as current projects have shown, people will want to use the demo server with this. I'm concerned that blindly overwriting the hosts file will break things in different OSes and especially if the end user application somehow uses the /etc/hosts file.

I am ok limiting this to the hosts file rather than '*.mender-convert-append' but it's hardly a complicated loop.

README.md
drewmoseley: Again, why?  The main effort lately has been on ease of use and requiring extra steps through the docs complicates that.  We support 3 main methods of servers at the moment so it seems like this should mirror that.

scripts/bootstrap-rootfs-overlay-demo-server.sh
```diff
@@ -0,0 +1,70 @@
...
+EOF
+wget -q "https://raw.githubusercontent.com/mendersoftware/meta-mender/master/meta-mender-demo/recipes-mender/mender/files/server.crt" -O ${output_dir}/etc/mender/server.crt
+
+echo "Configuration file written to: ${output_dir}/etc/mender"
```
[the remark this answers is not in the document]
drewmoseley: Fixed locally. Will push soon.

scripts/bootstrap-rootfs-overlay-production-server.sh
```diff
@@ -0,0 +1,80 @@
...
+}
+EOF
+
+echo "Configuration file written to: ${output_dir}/etc/mender"
```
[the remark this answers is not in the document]
drewmoseley: Fixed locally. Will push soon.

scripts/bootstrap-rootfs-overlay-demo-server.sh
drewmoseley: Fixed.

scripts/bootstrap-rootfs-overlay-demo-server.sh
drewmoseley: Fixed

mender-convert-modify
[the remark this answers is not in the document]
mirzak: We can keep it as-is, if you feel strongly about this. No need to limit it to `/etc/hosts`. 

A couple of comments anyway,

> but it's hardly a complicated loop

No I do not think that actual code is complicated, but it is additional method that we add to rootfs overlay which you can use to modify the target filesystem that needs to documented maintained, tested and so on. If this will ever be used to only modify `/etc/hosts`, imo it is not worth its own feature.

> blindly overwriting the hosts file will break things in different OSes and especially if the end user application somehow uses the /etc/hosts file.

As this is only relevant for the Demo Server, I do not see a big problem with this.

scripts/bootstrap-rootfs-overlay-demo-server.sh
drewmoseley: Done

mender-convert-modify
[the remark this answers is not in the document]
drewmoseley: OK.  I just removed it.

drewmoseley:
All changes added and pushed, except for the README update removal. I feel we need that in there.

mender-convert-modify
```diff
@@ -219,11 +219,4 @@ platform_modify
 for overlay in "${overlays[@]}"; do
   log_info "Applying rootfs overlay: ${overlay}"
   run_and_log_cmd "sudo rsync --archive --keep-dirlinks --verbose ${overlay}/ work/rootfs/"
-
-  for append in $(cd work/rootfs; find . -name '*.mender-convert-append'); do
-      sourcefile=$(echo $append | sed 's@.mender-convert-append$@@')
-      log_info "Applying rootfs overlay append to: ${sourcefile}"
-      run_and_log_cmd "sudo cat work/rootfs/${append} >> work/rootfs/${sourcefile}"
-      run_and_log_cmd "sudo rm -f work/rootfs/${append}"
-  done
```
mirzak: If the intention is to drop this, I suggest to drop the first commit instead of removing it here.

scripts/bootstrap-rootfs-overlay-demo-server.sh
```diff
@@ -0,0 +1,78 @@
...
+  "InventoryPollIntervalSeconds": 5,
+  "RetryPollIntervalSeconds": 30,
+  "ServerURL": "https://docker.mender.io",
+  "ServerCertificate": "/etc/mender/server.crt",
+  "UpdatePollIntervalSeconds": 5
+}
+EOF
+cat <<- EOF > ${output_dir}/etc/hosts
+127.0.0.1	localhost
+127.0.1.1	beaglebone.localdomain	beaglebone
```
mirzak: Drop this line. that is,

```
127.0.1.1	beaglebone.localdomain	beaglebone
```

mender-convert-modify
drewmoseley: Yeah, I forgot I had done that as a separate commit. Fixed.

scripts/bootstrap-rootfs-overlay-demo-server.sh
drewmoseley: Dropped

PR #129, opened 2019-10-24
Merge master into 1.2.x, and update to Mender 2.2.0

kacf:
There is only bugfixes in master, so we can safely merge.


kacf:
@lluiscampos, @mirzak: I'm still testing this, but let me know if you think it looks alright.

PR #133, opened 2019-11-25
modify: Add an extra function user_local_modify for end users to popu…

drewmoseley:
…late.

Changelog: Title

### DIFF
--- a/mender-convert-modify
+++ b/mender-convert-modify
@@ -98,8 +98,8 @@ cd - > /dev/null 2>&1
 run_and_log_cmd "sudo rsync --archive --keep-dirlinks --verbose work/mender-deb/files/ work/rootfs/"
 
 if [ "${MENDER_ENABLE_SYSTEMD}" == "y" ]; then
-  run_and_log_cmd "sudo ln -sf /lib/systemd/system/mender.service \
-        work/rootfs/etc/systemd/system/multi-user.target.wants/mender.service"
+  run_and_log_cmd "sudo ln -sf /lib/systemd/system/mender-client.service \
+        work/rootfs/etc/systemd/system/multi-user.target.wants/mender-client.service"
 fi
 
 if [ "${MENDER_GRUB_EFI_INTEGRATION}" == "y" ]; then
--- a/mender-convert-modify
+++ b/mender-convert-modify
@@ -20,6 +20,10 @@ function platform_modify() {
     true
 }
 
+function user_local_modify() {
+    true
+}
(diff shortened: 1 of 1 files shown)


drewmoseley:
I use this with a local mender_local_config file to populate extra files needed for Raspbian headless configuration.  Very useful for things like enabling uart and providing Wifi credentials

drewmoseley:
Looks like the service file name changed upstream for some reason.  Need to rename it mender-client.

lluiscampos:
> Looks like the service file name changed upstream for some reason. Need to rename it mender-client.

Yes. We changed it in the last updates on the Debian package and we didn't follow up in this repo. Good catch.

mirzak:
@lluiscampos , any idea on the gitlab errors? Something has changed in the infrastructure it seems?

lluiscampos:
It seems to be an infra problem with the acceptance tests. I am looking into it.

lluiscampos:
> It seems to be an infra problem with the acceptance tests. I am looking into it.

Fixed independently in #134

lluiscampos:
> @lluiscampos , any idea on the gitlab errors? Something has changed in the infrastructure it seems?
> 
> Do we need to run with `python3` now?
> 
> https://github.com/mendersoftware/mender-convert/blob/next/scripts/run-tests.sh#L102

(sorry I missed your comment before, seems like we were typing at the same time).

What happened is that `meta-mender` master has been updated for Python 3 but `mender-convert` has been left behind. See JIRA [QA-120](https://tracker.mender.io/browse/QA-120)

mirzak:
Why do we not just run the tests using `python3`? As far as I understand it, there are no specific tests for mender-convert, and we are just pulling the existing tests used by meta-mender which should have been ported to python3 already?

Yes we are typing at the same time :)

lluiscampos:
> Why do we not just run the tests using `python3`? As far as I understand it, there are no specific tests for mender-convert, and we are just pulling the existing tests used by meta-mender which should have been ported to python3 already?

While looking at the ticket now I also had the same idea: QA-120 is way over estimated and the actual work is only to update the submodule and revert #134... I think I will just do it straight away! I'll rework the other PR :)

lluiscampos:
New PR to update tests to Python3. See #135

lluiscampos:
@drewmoseley Can you please rebase this PR on top of latest next? The Pipeline should now run properly.

drewmoseley:
Rebased

PR #134, opened 2019-11-26
Set meta-mender acceptance tests reference to a fixed commit

lluiscampos:
The main acceptance tests have recently been updated to Python3, but
mender-convert repo has been left behind. Therefore, point to the commit
previous to such change from where to get the pytest helpers.


lluiscampos:
We will do the update at once. See #135

PR #137, opened 2019-11-29
Add xml results to GitLab so that they are displayed in the GUI

lluiscampos:


lluiscampos:
It can be seen in action in latest build: https://gitlab.com/Northern.tech/Mender/mender-convert/pipelines/99562348/test_report

mirzak:
Nice :)

scripts/run-tests.sh
```diff
@@ -99,14 +99,19 @@ convert_and_test() {
 
     cd ${WORKSPACE}/mender-image-tests
 
+    # This is a trick to make pytest generate different junit reports
+    # for different runs: renaming the tests folder to tests_<testsuite>
+    cp -r tests tests_${device_type}
```
oleorhagen: Why this instead of `--junitxml` argument to pytest?

lluiscampos: Because it is not about the filename (which already is different for the different boards), is about the test name ids inside the xml. Previously, we were properly generating different reports but GitLab was merging them in its fancy UI.

This is a hack (hence the comment justifying it) but I didn't find any other way of doing it.

PR #139, opened 2019-11-29
Implement MENDER_COPY_BOOT_GAP feature

mirzak:
Verified this on a Odroid C-2 board using an Armbian image.

### DIFF
--- a/mender-convert-extract
+++ b/mender-convert-extract
@@ -102,5 +102,8 @@ fi
 # this is done because some images might contain the bootloader embedded here
 # and we might need to keep this area intact when we generate our custom
 # image.
-disk_extract_part "${disk_image}" \
-  1 $(( $(disk_get_part_value ${disk_image} 1 START) - 1)) work/boot-gap.bin
+if [ "${MENDER_COPY_BOOT_GAP}" == "y" ]; then
+  log_info "Extracting boot gap to work/boot-gap.bin"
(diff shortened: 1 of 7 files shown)


mender-convert-extract
```diff
@@ -102,5 +102,8 @@ fi
 # this is done because some images might contain the bootloader embedded here
 # and we might need to keep this area intact when we generate our custom
 # image.
-disk_extract_part "${disk_image}" \
-  1 $(( $(disk_get_part_value ${disk_image} 1 START) - 1)) work/boot-gap.bin
+if [ "${MENDER_COPY_BOOT_GAP}" == "y" ]; then
+  disk_extract_part "${disk_image}" \
+    1 $(( $(disk_get_part_value ${disk_image} 1 START) - 1)) work/boot-gap.bin
+  log_info "Extracting boot gap to work/boot-gap.bin"
```
lluiscampos: I suggest we have the log before the extract. So on failure the user knows what was the last thing attempted.

mirzak:
> I suggest to first merge #138, then update this PR with the MENDER_COPY_BOOT_GAP setting for raspberrypi0w.

Indeed. That was my plan as well. Sorry for not stating my intentions

mender-convert-extract
mirzak: Thanks. Fixed

PR #140, opened 2019-12-01
Bump Go to 1.13.4

x1ddos:
See https://golang.org/doc/devel/release.html for Go release notes.
I also suggest to subscribe to https://groups.google.com/group/golang-announce to stay up to date with the regular and out of schedule releases.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM ubuntu:18.04
 
 ARG MENDER_ARTIFACT_VERSION=3.2.0
-ARG GOLANG_VERSION=1.11.2
+ARG GOLANG_VERSION=1.13.4
 
 RUN apt-get update && apt-get install -y \
     kpartx \


oleorhagen:
Hi @x1ddos, and thank you!

We do try and stay up to date on our versions, however, 1.11, is still supported, and hence covered by security patches. That is why we're not frantic about upgrading.

Also, we will (most likely this week), move on to mender-convert version 2.0, which is a complete rewrite of the tool. So if you'd like, please have a look at the `next` branch, if  you want your contribution to live for a little while longer :)

-O

x1ddos:
From https://golang.org/doc/devel/release.html#policy

> Each major Go release is supported until there are two newer major releases. For example, Go 1.5 was supported until the Go 1.7 release, and Go 1.6 was supported until the Go 1.8 release. We fix critical problems, including critical security problems, in supported releases as needed by issuing minor revisions (for example, Go 1.6.1, Go 1.6.2, and so on). 

You are currently at 1.11.2. But say, hypothetically, even if it were supported, there have been many patch releases past 1.11.2, more precisely 1.11.13 being the latest: https://golang.org/doc/devel/release.html#go1.11.minor. Just because something is supported it doesn't help if you stay at the old version. :)

I'll take a look at the next branch.

mirzak:
Note that in the `next` branch we are using pre-built packages generated with,

https://github.com/mendersoftware/mender-dist-packages

oleorhagen:
Targeted deprecated branch. Closing

PR #145, opened 2019-12-09
MEN-2924: Publish Raspbian images

kacf:
MEN-2924: Publish Raspbian images

This is best reviewed commit by commit.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -61,9 +61,10 @@ test_acceptance:
     - aws s3 cp s3://mender-gitlab-tmp-storage/$CI_PROJECT_NAME/$CI_PIPELINE_ID/deploy.tar.gz deploy.tar.gz
     - tar xzf deploy.tar.gz
     # Extract converted Raspbian artifacts
-    - unxz deploy/raspberrypi-mender-raspbian.sdimg.xz
+    - unxz deploy/raspberrypi-${RASPBIAN_NAME}-mender-${MENDER_CLIENT_VERSION}.sdimg.xz
   script:
-    - ./scripts/test/run-tests.sh
+    - ./scripts/test/run-tests.sh --prebuilt-image raspberrypi ${RASPBIAN_NAME}-mender-${MENDER_CLIENT_VERSION}
+    - ./scripts/test/run-tests.sh --all
   artifacts:
     expire_in: 2w
     when: always
@@ -93,7 +94,7 @@ convert_raspbian:
 
   script:
     - echo "MENDER_CLIENT_VERSION=${MENDER_CLIENT_VERSION}" > mender_client_version_config
-    - env MENDER_ARTIFACT_NAME=mender-raspbian
+    - env MENDER_ARTIFACT_NAME=${RASPBIAN_NAME}-mender-${MENDER_CLIENT_VERSION}
       ./docker-mender-convert -d ${RASPBIAN_NAME}.img
       -c configs/raspberrypi3_config
       -c configs/images/raspberrypi3_raspbian_config
@@ -116,7 +117,7 @@ publish:s3:
 
   script:
     - echo "Publishing ${RASPBIAN_NAME}-mender-${MENDER_CLIENT_VERSION}.img.xz version to S3"
-    - aws s3 cp deploy/raspberrypi-mender-raspbian.sdimg.xz
+    - aws s3 cp deploy/raspberrypi-${RASPBIAN_NAME}-mender-${MENDER_CLIENT_VERSION}.sdimg.xz
         s3://$S3_BUCKET_NAME/${RASPBIAN_NAME}/arm/${RASPBIAN_NAME}-mender-${MENDER_CLIENT_VERSION}.img.xz
     - aws s3api put-object-acl --acl public-read --bucket $S3_BUCKET_NAME
         --key ${RASPBIAN_NAME}/arm/${RASPBIAN_NAME}-mender-${MENDER_CLIENT_VERSION}.img.xz
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -2,6 +2,11 @@ image: docker:git
 
 variables:
   DOCKER_REPOSITORY: mendersoftware/mender-convert
+  S3_BUCKET_NAME: mender-convert-images
+  MENDER_ARTIFACT_VERSION: master
+  MENDER_CLIENT_VERSION: master
+  RASPBIAN_URL: http://downloads.raspberrypi.org/raspbian_lite/images/raspbian_lite-2019-09-30/2019-09-26-raspbian-buster-lite.zip
+  RASPBIAN_NAME: 2019-09-26-raspbian-buster-lite
 
 include:
   - project: 'Northern.tech/Mender/mendertesting'
@@ -13,13 +18,15 @@ stages:
   - test
   - build
   - test_acceptance
+  - convert
+  - publish
 
 build:
   stage: build
   services:
     - docker:dind
   script:
-    - IMAGE_NAME=$DOCKER_REPOSITORY:pr ./docker-build
+    - IMAGE_NAME=$DOCKER_REPOSITORY:pr ./docker-build --build-arg MENDER_ARTIFACT_VERSION=${MENDER_ARTIFACT_VERSION}
     - docker save $DOCKER_REPOSITORY:pr > image.tar
   artifacts:
     expire_in: 2w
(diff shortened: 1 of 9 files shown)


kacf:
I think this is finally ready for review now.

configs/mender_convert_config
```diff
@@ -4,12 +4,18 @@
...
 # makes it easier to transfer them between e.g an build server and a local
 # machine, and obviously saves space.
-MENDER_COMPRESS_DISK_IMAGE=y
+#
+# Possible values are:
+# 'n'    - No compression
+# 'gzip' - Compress with gzip
+# 'lzma' - Compress with xz (LZMA)
+# 'y'    - Alias for 'gzip'
+MENDER_COMPRESS_DISK_IMAGE=gzip
```
mirzak: Do not know if we need backwards compatibility here as we have not released this yet, meaning maybe we can drop `y/n` and keep the interface similar to the `MENDER_ARTIFACT_COMPRESSION`

In case we drop `y` we need to update https://github.com/mendersoftware/mender-convert/blob/next/scripts/run-tests.sh

scripts/test/run-tests.sh
```diff
@@ -53,12 +50,12 @@ convert_and_test "qemux86_64" \
...
-  "${RASPBIAN_IMAGE_URL}" \
-  "${RASPBIAN_IMAGE}.img" \
-  "${RASPBIAN_IMAGE}.zip" \
-  "configs/raspberrypi3_config" || test_result=$?
+if [ -f deploy/raspberrypi-mender-raspbian.sdimg ]; then
+  run_tests "raspberrypi" "mender-raspbian" || test_result=$?
+else
+  echo "FAILED! This test needs a pre-converted Raspbian image. See the convert_raspbian job in the .gitlab-ci.yml file for how to generate, and then put the *uncompressed* image in the deploy folder"
+  test_result=1
+fi
```
mirzak: Not sure what is going on here. Why do we need pre-converted image here?

configs/mender_convert_config
```diff
@@ -127,6 +127,41 @@ MENDER_USE_BMAP="n"
...
+# and if the result is larger than IMAGE_ROOTFS_SIZE + IMAGE_ROOTFS_EXTRA_SPACE,
+# it will be used as the size of the rootfs instead of the other two variables.
+#
+# The actual free space will usually be lower than requested. See comment for
+# IMAGE_ROOTFS_EXTRA_SPACE.
+#
+# This variable directly mirrors the variable from the Yocto Project, which is
+# why it is missing a "MENDER_" prefix.
+IMAGE_OVERHEAD_FACTOR="1.5"
+
```
mirzak: This warrants an update to the docs as well, https://docs.mender.io/development/artifacts/debian-family/variables

configs/mender_convert_config
kacf: Good point, will fix this!

scripts/test/run-tests.sh
kacf: We want to test the exact image we are releasing, which is made in a different stage, and not using the `convert_and_test` call. This was mostly to give some heads up if people run this locally, though I don't expect many people to do that. Does that explain it?

I could wrap it in a special flag which we only call from `.gitlab-ci.yml`, to keep local tests passing, but I opted for being verbose about it instead. Let me know if you would like it changed.

configs/mender_convert_config
kacf: Aha! I will make that change too!

scripts/test/run-tests.sh
mirzak: Understood.

Probably not many people will try to run the tests locally, but at least I have appreciated to be able to do so and if not to much trouble it would be nice to continue doing it :)

.gitlab-ci.yml
```diff
@@ -16,6 +16,7 @@ stages:
 
 build:
   stage: build
+  image: docker:18-dind
```
lluiscampos: When missing, it is using the pipeline scope defined (in this case `docker:git`). Good to have it explicit in the job.

However, you don't need a docker:dind image when launching another docker:dind as a service, `docker:git` should be enough and it is the one we are using in every other Docker build job in other repos.

.gitlab-ci.yml
```diff
@@ -2,6 +2,11 @@ image: docker:git
 
 variables:
   DOCKER_REPOSITORY: mendersoftware/mender-convert
+  S3_BUCKET_NAME: mender-convert-images
+  MENDER_ARTIFACT_VERSION: master
+  MENDER_CLIENT_VERSION: master
+  RASPBIAN_URL: http://downloads.raspberrypi.org/raspbian_lite/images/raspbian_lite-2019-09-30/2019-09-26-raspbian-buster-lite.zip
+  RASPBIAN_NAME: 2019-09-26-raspbian-buster-lite
```
lluiscampos: Have we thought about how to support more than one at some point? Between major Debian releases we might be in the situation where we have to. This is long in the future so just something to think about.

.gitlab-ci.yml
```diff
@@ -58,3 +65,52 @@ test_acceptance:
...
+
+    - export IMAGE_NAME=$DOCKER_REPOSITORY:pr
+    - docker load -i image.tar
+
+    - wget -q ${RASPBIAN_URL}
+    - unzip ${RASPBIAN_NAME}.zip
+
+  script:
+    - echo "MENDER_CLIENT_VERSION=${MENDER_CLIENT_VERSION}" > mender_client_version_config
+    - env MENDER_ARTIFACT_NAME=mender-raspbian
```
lluiscampos: I think the Artifact name should include somehow the Raspbian version, maybe `mender-2019-09-26-raspbian-buster-lite`

scripts/test/run-tests.sh
lluiscampos: While working on the tool I also thought it was of great value to build/test locally, so I think we should keep it.

In this script, I would do if the pre-converted image exists, run_tests, else, convert_and_test (as it was before).

.gitlab-ci.yml
lluiscampos: And possibly the mender version also.

.gitlab-ci.yml
```diff
@@ -2,6 +2,13 @@ image: docker:git
 
 variables:
   DOCKER_REPOSITORY: mendersoftware/mender-convert
+  S3_BUCKET_NAME: mender-convert-images
+  MENDER_ARTIFACT_VERSION: master
+  MENDER_CLIENT_VERSION: master
```
lluiscampos: Once we release (say we branch out for 2.0.x), then this line will be modified to point to stable Mender client (2.2.0, most likely).

Or do you have other idea in mind?

.gitlab-ci.yml
kacf: Not deeply, but the method is extendable if we need it. New Debian versions are not the only reason to do this, but for example a Beaglebone image might also be introduced at some point. Right now, I prefer not to overcomplicate it as I think I would probably get it subtly wrong anyway.

.gitlab-ci.yml
kacf: Right, I will change that for clarity. But note that it is anyway renamed in the publication stage to the final name.

scripts/test/run-tests.sh
kacf: > In this script, I would do if the pre-converted image exists, run_tests, else, convert_and_test (as it was before).

I don't want to use a method that will silently pass if the image is missing (risk of false negative). If it's desired to convert using the old method, then I will base that on a flag instead (`--use-prebuild-image` or something), and still fail if the flag is specified and the image is missing.

.gitlab-ci.yml
kacf: Yes, but we need `mender setup`, don't we? I recall you mentioning something about needing the master branch until 2.3, but I can't recall the exact details.

lluiscampos: Discussed in person.

We are both referring to the same thing: Mender 2.3 which will include client 2.2.0

scripts/test/run-tests.sh
lluiscampos: Good point about the false positive. I like the extra flag suggestion, it is not that hard to implement right?

Maybe we just actually run the tests on this new prebuild image in addition to the build+test one, else we also risk the latter to fail the conversion or something down the road (for example if there is an problem with the link upstream or something)

scripts/test/test-utils.sh
```diff
@@ -1,37 +1,8 @@
-#!/bin/bash
```
lluiscampos: Not a big deal, but it is only used from `run-tests.sh` so it is not necessary to have split these functions.

.gitlab-ci.yml
kacf: Right, I will fix this.

scripts/test/test-utils.sh
kacf: True, but I like the separation. It makes `run-tests.sh` very easy to read.

scripts/test/run-tests.sh
kacf: Indeed, this is what I ended up doing. The script now requires a parameter, either `--all` or `--prebuilt-image ...`. So the prebuilt image is now an additional test, not a replacement test.

kacf:
Updated!

PR #147, opened 2019-12-16
MEN-2927: Make sure `.mender` file is uploaded for use in tutorial.

kacf:
Changelog: None

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -109,7 +109,10 @@ publish:s3:
   stage: publish
   image: debian:buster
   before_script:
-    - apt update && apt install -yyq awscli
+    - apt update && apt install -yyq awscli curl
+
+    - curl -Lo /usr/bin/mender-artifact https://d1b0l86ne08fsf.cloudfront.net/mender-artifact/${MENDER_ARTIFACT_VERSION}/linux/mender-artifact
+    - chmod ugo+x /usr/bin/mender-artifact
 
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -124,3 +127,9 @@ publish:s3:
         s3://$S3_BUCKET_NAME/${RASPBIAN_NAME}/arm/${RASPBIAN_NAME}-mender-${MENDER_CLIENT_VERSION}.img.xz
     - aws s3api put-object-acl --acl public-read --bucket $S3_BUCKET_NAME
         --key ${RASPBIAN_NAME}/arm/${RASPBIAN_NAME}-mender-${MENDER_CLIENT_VERSION}.img.xz
+
+    - mender-artifact modify -n release-1 deploy/raspberrypi-${RASPBIAN_NAME}-mender-${MENDER_CLIENT_VERSION}.mender
```
lluiscampos: I think we should do the actual conversion in `convert` stage, pass it as a build artifact, and here only upload it.

kacf: I thought about it, but this is tricky, because the tests test the name, and would fail if it was different. We also cannot rename during the mender-convert stage itself, because then they will be the same, which we're trying to avoid.

The renaming is specifically for our distribution and our demo, so I think it's alright to do it here. It's not how mender-convert should work in general.

lluiscampos: Ok. Agreed.

PR #153, opened 2020-01-14
MEN-2794: Adjust the expected S3 locations for the binaries

lluiscampos:
See also https://github.com/mendersoftware/mender-convert-integration-scripts/pull/3

The S3 path now follows the pattern:
`/mender/mender-convert/uboot/<dev-family>/<dev-config>-<uboot-ver>.tar`

For RPi and BBB these are automatically published form the CI Pipeline
of repo mender-convert-integration-scripts while for RockPro64 they have
been manually moved (as we expect this to be a one time integration).

Reworked a bit also the *_BINARIES variable so that we can keep the
u-boot version in the base config files and let the childs only define
the a config name.

Changelog: None

### DIFF
--- a/configs/beaglebone_black_debian_emmc_config
+++ b/configs/beaglebone_black_debian_emmc_config
@@ -1,8 +1,7 @@
 # Binaries generated with the following script:
-#
-#    https://github.com/drewmoseley/mender-convert-integration-scripts/blob/master/build-uboot-bbb.sh
-#
-BEAGLEBONE_BLACK_BINARIES="beaglebone-black-integration-debian-emmc-2018.07.004.tar"
+#    https://github.com/mendersoftware/mender-convert-integration-scripts/blob/master/build-uboot-bbb.sh
+
+BEAGLEBONE_BLACK_CONFIG="beaglebone_black_debian_emmc"
 source configs/beaglebone_black_base_config
 MENDER_STORAGE_TOTAL_SIZE_MB="3648"
 MENDER_DEVICE_TYPE="beaglebone-emmc"
--- a/configs/rockpro64_emmc_config
+++ b/configs/rockpro64_emmc_config
@@ -1,8 +1,4 @@
+ROCKPRO64_CONFIG="rockpro64_emmc"
 MENDER_STORAGE_DEVICE=/dev/mmcblk1p
-
-# Binaries generated with the following script:
-#
-#    https://d1b0l86ne08fsf.cloudfront.net/mender-convert/armbian/rockpro64/integration-scripts.tar.gz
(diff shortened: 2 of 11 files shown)


configs/rockpro64_emmc_config
```diff
@@ -1,8 +1,4 @@
+ROCKPRO64_CONFIG="emmc-boot"
```
mirzak: This this should be:

```
ROCKPRO64_CONFIG="rockpro64_emmc"
```

configs/rockpro64_sd_config
```diff
@@ -1,6 +1,3 @@
-# Binaries generated with the following script:
-#
-#    https://d1b0l86ne08fsf.cloudfront.net/mender-convert/armbian/rockpro64/integration-scripts.tar.gz
-ROCKPRO64_BINARIES="sd-boot-integration-2017.09.tar.gz"
+ROCKPRO64_CONFIG="sd-boot"
```
mirzak: This this should be:

ROCKPRO64_CONFIG="rockpro64_sd"

configs/beaglebone_black_debian_emmc_config
```diff
@@ -1,8 +1,7 @@
 # Binaries generated with the following script:
-#
-#    https://github.com/drewmoseley/mender-convert-integration-scripts/blob/master/build-uboot-bbb.sh
-#
-BEAGLEBONE_BLACK_BINARIES="beaglebone-black-integration-debian-emmc-2018.07.004.tar"
+#    https://github.com/mendersoftware/mender-convert-integration-scripts/blob/master/build-uboot-bbb.sh
+
+BEAGLEBONE_BLACK_CONFIG="beaglebone-black-debian-emmc"
```
mirzak: Also should this one follow the config name? You are now using `-` instead of underscore (`_`) which is used in the actual configuration name?

configs/rockpro64_emmc_config
lluiscampos: These I haven't changed to avoid modifying the scripts for RockPro64 (as they are manually updated to S3 and not checked in in any repo).

I agree in that we should follow the same convention, so I will then modify the script and re-upload both the script tarball and the binaries.

configs/beaglebone_black_debian_emmc_config
lluiscampos: Makes sense. I will rework it here and in https://github.com/mendersoftware/mender-convert-integration-scripts/pull/3

lluiscampos:
@mirzak I reworked here and in https://github.com/mendersoftware/mender-convert-integration-scripts/pull/3 to use `_` in the config names. For RockPro64, I manually modified the script, recreated the binaries, and re-upload them.

Please have a second look and consider approving the PRs ;)

PR #154, opened 2020-01-16
Set the acceptance test jobs to run in parallel on different machines

lluiscampos:
The current execution of the acceptance tests is too long, in some cases
reaching the GitLab timeout of this project of 2 hours.

Rework the Pipeline to use a template and rework the script to allow
running tests for only one device type.

Changelog: None

### DIFF
--- a/scripts/test/run-tests.sh
+++ b/scripts/test/run-tests.sh
@@ -3,7 +3,7 @@
 set -e
 
 usage() {
-  echo "$0 [--no-pull] <--all | --prebuilt-image DEVICE_TYPE IMAGE_NAME>"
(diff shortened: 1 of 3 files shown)


scripts/test/run-tests.sh
```diff
@@ -48,58 +48,61 @@ mkdir -p ${WORKSPACE}
 
 get_pytest_files
 
+if ! [ "$1" == "--all" -o "$1" == "--only" -a -n "$2" -o "$1" == "--prebuilt-image" -a -n "$3" ]; then
```
kacf: The argument parsing isn't very good in this script (depends on position), but it was preexisting.

PR #155, opened 2020-01-16
MEN-3051: Update the naming convention for the output files

oleorhagen:


mender-convert-package
```diff
@@ -194,7 +194,7 @@ mender_create_artifact "${device_type}" "${artifact_name}"
 
 log_info "Creating Mender compatible disk-image"
 
-sdimg_path=deploy/${image_name}.sdimg
+img_path=deploy/${image_name}.mender.img
```
oleorhagen: Outputs: `filename.mender.img.gz`, instead of `filename-mender.img.gz` like in the ticket. To me it made more sense. But your call @mirzak

mirzak: I think it should be `-mender.img` as was in the ticket.

oleorhagen: Done!

oleorhagen:
> Changes look good to me.
> 
> Though why is gitlab failing?

It's failing cause of the naming change, and the tests are expecting to work with sdimg's. 
Fixing now.

oleorhagen:
Should be all good now. Will leave here pending response on: https://tracker.mender.io/browse/MEN-3051?focusedCommentId=104089&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-104089 though.

PR #156, opened 2020-01-16
Fix Rasbperry Pi device type

psporysz:
I've built my custom Raspbian image using [pi-gen](https://github.com/RPi-Distro/pi-gen) project. By default, it generates random hostname but i am using other values too and it should not matter when it comes to determining the device type.

I found that device type of the generated image (and artifact) is taken from `/etc/hostname`. Since i do not understand what's the rationale for such source of default value for device type my proposition is to fix the device type in the config at least for Raspberry family which would work exactly the same as for oryginal Raspbian images.

### DIFF
--- a/configs/raspberrypi_config
+++ b/configs/raspberrypi_config
@@ -1,6 +1,9 @@
 # Binaries generated with the following script:
 #    https://github.com/mendersoftware/mender-convert-integration-scripts/blob/master/build-uboot-rpi.sh
 
+# Fix device type to avoid hostname as a source
+MENDER_DEVICE_TYPE="raspberrypi"
+
 # Raspberry Pi does not support GRUB bootloader integration, fallback to U-boot.
 MENDER_GRUB_EFI_INTEGRATION=n
 


lluiscampos:
Hi @psporysz and thanks for your contribution.

I don't have an strong opinion on this, so I let it to @mirzak to clarify his original intention on using `/etc/hostname` for the device type.

Meanwhile, please see our [Contributing guides for PR](https://github.com/mendersoftware/mender/blob/master/CONTRIBUTING.md#providing-pull-requests): you need to sign your commit with `-s` and to provide a `Changelog` tag in your commit message.

lluiscampos:
@psporysz Thanks. But the QA is still complaining because your git email and GPG email do not match:

```
commit 1ce8ccf7e887a88557b0ab08341203d889242d1b
Author: Przemyslaw Sporysz <przemyslaw.sporysz@euvic.pl>
Date:   Thu Jan 16 16:31:13 2020 +0100
    Fix Rasbperry Pi device type
    
    Prevent from using hostname as device type e.g. for custom built Raspbian
    images.
    
    Changelog: None
    
    Signed-off-by: Przemyslaw Sporysz <przemyslaw.sporysz@gmail.com>
```

mirzak:
> I found that device type of the generated image (and artifact) is taken from /etc/hostname. Since i do not understand what's the rationale for such source of default value for device type my proposition is to fix the device type in the config at least for Raspberry family which would work exactly the same as for oryginal Raspbian images.

`/etc/hostname` was used in lack of a better source for a default value. This normally works out if you are using "stock images" instead of images generated e.g using rpi-gen. This is primarily useful for testing purposes and on-boarding.

The recommend would be to override this value with something custom to your product/device as none of the defaults would "correct" for everyone.

To override the default values you can do the following:

```
echo 'MENDER_DEVICE_TYPE="raspberrypi"" >> custom_confg

# Now you can additional provide your custom configuration file
MENDER_ARTIFACT_NAME=release-1 ./docker-mender-convert \
   --disk-image input/$INPUT_DISK_IMAGE \
   --config configs/raspberrypi3_config \
   --config custom_config
   --overlay rootfs_overlay_demo/
```

For this reason I would NAK this change.

psporysz:
@mirzak thx for custom_config tip, it's ok, i've just seen that some config files already have the type (e.g. beaglebone or qemu) thus i had thought it was overlooked at some point.

The convert step takes a while and it is irritating at times when some images has proper type but some not, maybe instead of fixing it we could extend the doc with note about device type? It would had saved me some time when i invoked it first time.

mirzak:
> maybe instead of fixing it we could extend the doc with note about device type? It would had saved me some time when i invoked it first time.

Definitely, this should be made clearer in the docs.

mirzak:
@psporysz, after all a similar change was merged https://github.com/mendersoftware/mender-convert/commit/9217bf45ec189b2beebfded73c28bb1d8f565085.

Will close this one

PR #157, opened 2020-01-21
Switch to version based location in S3 bucket for GRUB EFI binaries

lluiscampos:
See related PR https://github.com/mendersoftware/mender-convert-integration-scripts/pull/4


lluiscampos:
@mirzak Do you have the chance to test this on hardware for a `x86_64` and `arm64` boards?

It works for `arm`, so if you don't have the time I think it is low risk to merge and see if there are bug reports.

mirzak:
I can do a test on `x86_64`. Reporting back when done

lluiscampos:
> Everything seems to be in order when testing x86_64 (qemu). Good job!

Thanks for testing! I didn't think about qemu, that would have been a way faster feedback loop for my testing...

PR #159, opened 2020-02-03
Men 3063

oleorhagen:


mender-convert-package
```diff
@@ -138,7 +138,12 @@ rootfs_part_sectors=$(disk_align_sectors ${rootfs_part_sectors} ${MENDER_PARTITI
 device_type=$(cat work/rootfs/data/mender/device_type | sed 's/[^=]*=//')
 artifact_name=$(cat work/rootfs/etc/mender/artifact_info | sed 's/[^=]*=//')
 
-image_name="${device_type}-${artifact_name}"
+# Get the name from the input disk_image
+temp_img_name=$(basename "$disk_image")
+
+# Strip the image suffix from temp_img_name and set the original image name as
+# the output image name
+image_name="${temp_img_name%.*}"
```
kacf: IMHO, we should still add `-mender` at the end to tell them apart.

mender-convert-package
```diff
@@ -143,7 +143,7 @@ temp_img_name=$(basename "$disk_image")
 
 # Strip the image suffix from temp_img_name and set the original image name as
 # the output image name
-image_name="${temp_img_name%.*}"
+image_name="${temp_img_name%.*}-mender"
```
kacf: Here you are doing exactly what I said, but is this a test commit..?

mender-convert-package
```diff
@@ -290,7 +290,7 @@ else
 fi
 distro_feature="${distro_feature} mender-convert"
 
-cat <<- EOF > deploy/${image_name}.cfg
+cat <<- EOF > deploy/${device_type}-${artifact_name}.cfg
```
kacf: Why are this one and the next one exceptions?

oleorhagen:
@kacf Ready for review

scripts/test/test-utils.sh
```diff
@@ -57,16 +57,16 @@ convert_and_test() {
     ${MENDER_CONVERT_EXTRA_ARGS}
 
   local ret=0
-  run_tests "${device_type}" "${artifact_name}" || ret=$?
+  run_tests "${device_type}" "${image_name%.img}-mender" || ret=$?
```
kacf: Could you rename this variable to `image_file`? It's quite confusing to read because in the other file, `image_name` is without suffix, here it is with suffix.

The same should probably be done with `image_name_compressed`, which doesn't appear in your patch, but which is related.

scripts/test/test-utils.sh
```diff
@@ -76,30 +76,33 @@ run_tests() {
...
-      mv $file "${file%-mender.img}.sdimg"
+      mv $file "${file%.img}.sdimg"
   done
 
   cd ${WORKSPACE}/mender-image-tests
 
   python3 -m pytest --verbose \
     --junit-xml="${MENDER_CONVERT_DIR}/results_${device_type}.xml" \
+    --pdb \
+    -x \
```
kacf: Leftovers?

scripts/test/test-utils.sh
```diff
@@ -57,16 +57,17 @@ convert_and_test() {
     ${MENDER_CONVERT_EXTRA_ARGS}
 
   local ret=0
-  run_tests "${device_type}" "${artifact_name}" || ret=$?
+  image_file = "${image_name%.img}-${device_type}-mender"
+  run_tests "${device_type}" "$image_file" || ret=$?
```
kacf: This also means you should be referring to `image_name` here, not `image_file`.

scripts/test/test-utils.sh
```diff
@@ -76,16 +77,16 @@ run_tests() {
 
   # Need to decompress images built with MENDER_COMPRESS_DISK_IMAGE=gzip before
   # running tests.
-  if [ -f deploy/${device_type}-${artifact_name}-mender.img.gz ]; then
+  if [ -f deploy/*.gz ]; then
     # sudo is needed because the image is created using docker-mender-convert
     # which sets root permissions on the image
-    sudo gunzip --force deploy/${device_type}-${artifact_name}-mender.img.gz
+    sudo gunzip --force deploy/*.gz
```
kacf: Why `*.gz`? This can easily break if you have more than one image in the directory while testing locally. This should be `${converted_image_name}.gz`

scripts/test/test-utils.sh
```diff
@@ -76,16 +77,16 @@ run_tests() {
...
   fi
 
   # MEN-3051: Rename the files back to .sdimg, as the sdimg extension has meaning
   # in the test-infrastructure.
   for file in ${MENDER_CONVERT_DIR}/deploy/*.img; do
```
kacf: This can probably also be changed to `${MENDER_CONVERT_DIR}/deploy/${converted_image_name}.img`.

scripts/test/test-utils.sh
```diff
@@ -53,20 +53,24 @@ convert_and_test() {
...
-    --disk-image input/${image_name} \
+    --disk-image input/${image_file} \
     ${MENDER_CONVERT_EXTRA_ARGS}
 
   local ret=0
-  run_tests "${device_type}" "${artifact_name}" || ret=$?
 
-  rm -f deploy/${device_type}-${artifact_name}*
+  # The output image name after conversion
+  image_name="${image_file%.img}-${device_type}-mender"
```
kacf: :+1:

PR #163, opened 2020-02-06
Fix DTB boot issue and switch back to GRUB 2.04

kacf:
```
commit 8d6902707187a8a683947d9c37457bfa7ba994d7
Date:   Thu Feb 6 11:01:57 2020

    Switch back to GRUB 2.02.
    
    2.04 does not work on Beaglebone Black.
    Ticket: https://tracker.mender.io/browse/MEN-2404
    
    Changelog: None
    
commit c321684a03764468877aee6b92b49835a6550990
Date:   Tue Jan 28 15:33:39 2020

    Make sure that DTB files are included on boot partition.
    
    This is important so that the bootloader (U-Boot) can use the DTB
    files when chainloading GRUB via UEFI. The logic for doing this is
    already present in `distro_bootcmd`, but because we were not placing
    the DTB files in the correct folder, this part was skipped.
    
    Changelog: Fix certain kernels hanging on boot. In particular, recent
    versions of Debian for Beaglebone was affected, but several other
    boards using UEFI may also have been affected.
    
```

### DIFF
--- a/configs/mender_grub_config
+++ b/configs/mender_grub_config
@@ -32,4 +32,4 @@ MENDER_GRUB_KERNEL_IMAGETYPE=""
 # if was not possible to auto-detect
 MENDER_GRUB_INITRD_IMAGETYPE=""
 
-MENDER_GRUB_BINARY_STORAGE_URL="${MENDER_STORAGE_URL}/mender-convert/grub-efi/2.04"
+MENDER_GRUB_BINARY_STORAGE_URL="${MENDER_STORAGE_URL}/mender-convert/grub-efi/2.02"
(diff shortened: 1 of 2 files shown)


kacf:
@mirzak may have some input as well:

> Should we get the rid of [BBB configs](https://github.com/mendersoftware/mender-convert/tree/master/configs)? If sou, we could also remove them from the `mender-convert-integration-scripts`.
> I actually checked the git history, and these were never used by the tests neither.
> 
> My point is to make a clear statement that when possible (i.e. BBB) we want GRUB/EFI and not u-boot for Mender integration.

Good point. We should start by adapting them, because there are a couple of fixes which are bootloader independent (most notably the udisks fix and the emmc storage location fix).

However, regarding keeping U-Boot support, I would definitely like to mark them non-standard and "recommend not using". However I can see at least one valid use case, and that's using the new mender-convert to upgrade an existing device to a new image, which already has U-Boot integration. That upgrade won't work with the GRUB integration. Maybe we can consider some `contrib` folder or something like that, for the U-Boot config?

Regardless I will merge this PR, and launch a followup PR for the above issue.

PR #164, opened 2020-02-10
MEN-3165: Make device_type specific to each Raspberry Pi model.

kacf:
This is a bit unfortunate, but our converted image does not work for
all of them, like the original.

Changelog: Title

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -76,8 +76,8 @@ test_acceptance_prebuilt_raspberrypi:
     - aws s3 cp s3://mender-gitlab-tmp-storage/$CI_PROJECT_NAME/$CI_PIPELINE_ID/deploy.tar.gz deploy.tar.gz
     - tar xzf deploy.tar.gz
     # Extract converted Raspbian artifacts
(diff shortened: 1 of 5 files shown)


kacf:
Thanks, updated!

>     * And we probably need a new config file for ubuntu conversion and test, as it was using [the same as raspberrypi3](https://github.com/mendersoftware/mender-convert/blob/master/scripts/test/run-tests.sh#L99-L104).

I think it's fine to use "raspberrypi3", since it is in fact a Raspberry Pi image, just from a different vendor. In any case, updated.

.gitlab-ci.yml
```diff
@@ -77,7 +77,7 @@ test_acceptance_prebuilt_raspberrypi:
     - tar xzf deploy.tar.gz
     # Extract converted Raspbian artifacts
     - unxz deploy/raspberrypi-${RASPBIAN_NAME}-mender-${MENDER_CLIENT_VERSION}-mender.img.xz
```
lluiscampos: There is still missing the update on the `unxz` command.

kacf: Doh! Thanks, updated!

PR #165, opened 2020-02-10
Use consistent compression and archive naming.

drewmoseley:

### DIFF
--- a/configs/beaglebone_black_base_config
+++ b/configs/beaglebone_black_base_config
@@ -10,14 +10,14 @@ MENDER_COPY_BOOT_GAP=n
 # 4MB alignment
 MENDER_PARTITION_ALIGNMENT="4194304"
 
-BEAGLEBONE_BLACK_BINARIES="${BEAGLEBONE_BLACK_CONFIG}-2018.07.tar"
+BEAGLEBONE_BLACK_BINARIES="${BEAGLEBONE_BLACK_CONFIG}-2018.07.tar.gz"
 BEAGLEBONE_BLACK_BINARIES_URL="${MENDER_STORAGE_URL}/mender-convert/uboot/beaglebone/${BEAGLEBONE_BLACK_BINARIES}"
 
 function platform_modify() {
   mkdir -p work/bbb/binaries
 
   run_and_log_cmd "wget -q ${BEAGLEBONE_BLACK_BINARIES_URL} -P work/bbb/binaries"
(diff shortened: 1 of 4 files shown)


drewmoseley:
Note that this will require an update of the stored binaries.  And I don't think we have the rockpro sources yet in uboot-mender.

drewmoseley:
I actually had the fixes to mender-convert-integration-scripts locally but forgot to submit them. See https://github.com/mendersoftware/mender-convert-integration-scripts/pull/7

drewmoseley:
Regarding ROCKPRO, before this commit they were extracted with "tar xvf" but now it is "tar xzvf" so I suspect that will fail if the binaries are not repackaged.

lluiscampos:
> Regarding ROCKPRO, before this commit they were extracted with "tar xvf" but now it is "tar xzvf" so I suspect that will fail if the binaries are not repackaged.

That is correct. I will manually update them as I said in `mender-convert-integration-scripts` PR

lluiscampos:
See also https://github.com/mendersoftware/mender-convert-integration-scripts/pull/7

lluiscampos:
The problem now is that once we merge `mender-convert-integration-scripts` PR, the already tagged mender-convert 2.0.0b1 will stop working, as some of the files use the same extension but now properly compressed.

I am thinking on waiting with these two PRs for one week more, and merge them just few days before final release.

What do you think @kacf and others?

kacf:
But it will not delete already uploaded binaries, will it?

lluiscampos:
> But it will not delete already uploaded binaries, will it?

Yes, it will. For example for [RPi u-boot](https://github.com/mendersoftware/mender-convert-integration-scripts/pull/7/files#diff-ec9076aed12a7e7bfd82c71de8cc2f80) or for [GRUB EFI](https://github.com/mendersoftware/mender-convert-integration-scripts/pull/7/files#diff-286f7969d4606fbe29e1ff6e9b0f5239), the files will have the same extension (same filename) but now properly compressed.

kacf:
Ah right, some even have the same name. Then I agree, let's merge this as late as we can basically. Make sure this is on the list of blockers or in the release checklist for 2.3.0!

lluiscampos:
> Ah right, some even have the same name. Then I agree, let's merge this as late as we can basically. Make sure this is on the list of blockers or in the release checklist for 2.3.0!

@kacf Good suggestion. I created a JIRA ticket for it, ref https://tracker.mender.io/browse/MEN-3202

@drewmoseley So the decision is to not merge this until end of next week. However this PR needs a rebase and conflict resolution. I can take over if you don't have the time.

drewmoseley:
@lluiscampos rebased.

PR #166, opened 2020-02-11
fix target path for initrd image in probe_initrd_in_boot_and_root

doudou:


doudou:
I have to admit that I'm really not sure about this one... Is it a rarely taken code path ? Or have I done something wrong ?

modules/probe.sh
```diff
@@ -218,7 +218,7 @@ probe_initrd_in_boot_and_root() {
   for boot in work/rootfs/boot work/boot; do
     initrd_image_path=$(probe_initrd_image ${boot})
     if [ -n "${initrd_image_path}" ] && [ "${boot}" == "work/boot" ]; then
-      sudo cp ${initrd_image_path} ${target_rootfs_dir}/boot
+      sudo cp ${initrd_image_path} work/rootfs/boot
```
kacf: It does look correct to me, especially since `target_rootfs_dir` does not occur anywhere else in this file. @mirzak, do you concur?

doudou:
It seems that the Ubuntu image you use in the acceptance tests is gone: https://gitlab.com/Northern.tech/Mender/mender-convert/-/jobs/436033048#L901

modules/probe.sh
mirzak: Yes this look good to me. It is a path code that you do not normally enter and for sure not one we enter in our conversion runs during testing, hence went by unnoticed.

kacf:
Test failure is because of the upstream link change, not related to this PR. Merging.

PR #167, opened 2020-02-11
add MENDER_SKIP_FIRST_PARTITION to allow ignoring the input image's first partition

doudou:


drewmoseley:
@doudou what is the reason for this change? Is there value in making this more general? ie the ability to skip arbitrary partitions?

doudou:
> Is there value in making this more general? ie the ability to skip arbitrary partitions?

Probably.

The use-case was an Ubuntu server install done in a VM, which has a first partition that simply not usable. mender-convert got confused over that one. Since Mender does generate a new boot partition, just skipping it was easier.

Generally speaking, it would make sense to just let the config file decide which partition is the boot partition and which is the root partition, including letting mender-convert generate the first boot partition.

I really don't have time for this right now (read: within the next month), unfortunately.

mender-convert-extract
```diff
@@ -93,9 +93,14 @@ done
 # but we do require it be present and hence we might need to generate
 # a filesystem image here.
 if [ ${nr_of_parts} -eq 1 ]; then
-  log_info "Generating boot partition (required, does not exist in original image)"
-  run_and_log_cmd "dd if=/dev/zero of=work/boot-generated.vfat count=${MENDER_BOOT_PART_SIZE_MB} bs=1M status=none"
-  run_and_log_cmd "${MKFS_VFAT} work/boot-generated.vfat"
```
kacf: The `boot-generated.vfat` name is used `modules/disk.sh`, so please don't change it unless you also change the logic there.

mender-convert-extract
```diff
@@ -93,9 +93,14 @@ done
...
+  run_and_log_cmd "mv work/part-1.fs work/part-2.fs"
+  MENDER_SKIP_FIRST_PARTITION=1
+fi
+
+if [ ${MENDER_SKIP_FIRST_PARTITION} -eq 1 ]; then
```
kacf: This should be documented at the very least in `mender_convert_config`, and preferably also in [the documentation](https://github.com/mendersoftware/mender-docs/blob/master/04.Artifacts/15.Debian-family/03.variables/docs.md) also.

doudou:
> I suspect you will have to adjust the signoff here as well, as in your other PR.

If you could be more specific as to what's wrong with the Signed-off-by line in the commit, I'd be glad. I have no idea why your scripts don't pick it up.

mender-convert-extract
doudou: True. The logic change here basically should make the special cases around `boot-generated.vfat` obsolete. Should I just remove the `disk_boot_part` and `disk_root_part` functions altogether ?

doudou:
> > I suspect you will have to adjust the signoff here as well, as in your other PR.
> 
> If you could be more specific as to what's wrong with the Signed-off-by line in the commit, I'd be glad. I have no idea why your scripts don't pick it up.

Forget it, saw your explanation in the other PR

mender-convert-extract
doudou: True. The logic change here basically should make the special cases around boot-generated.vfat obsolete. Should I just remove the disk_boot_part and disk_root_part functions altogether ?

kacf:
I'll take a look at this tomorrow. Just FYI, the failure now is due to an upstream link change and has nothing to do with your PR. Just ignore it.

mender-convert-extract
kacf: I *think* that should be ok. I'm not an expert on this code, so it would be nice if @mirzak could confirm this.

mirzak: I am refreshing my memory on this , hold on a sec

configs/mender_convert_config
```diff
@@ -102,6 +102,16 @@ MENDER_STORAGE_URL="https://d1b0l86ne08fsf.cloudfront.net"
 # Mender GitHub organization URL prefix
 MENDER_GITHUB_ORG="https://github.com/mendersoftware"
 
+# Whether the first partition of the input image should be ignored.
+#
+# For EFI targets, mender-convert assumes that the first partition of the input
```
mirzak: This is not isolated to EFI targets so can say `mender-convert assumes...`

configs/mender_convert_config
```diff
@@ -102,6 +102,16 @@ MENDER_STORAGE_URL="https://d1b0l86ne08fsf.cloudfront.net"
...
+# image should be used as a base for the boot partition. If it does not exist,
+# mender-convert will create one instead
+#
+# Setting this variable to 1 will force ignoring this partition, instead using
+# the auto-generated one.
```
mirzak: Suggestion:

Setting this variable to 1 will force mender-convert to generate a boot part instead of using the first partition of the input image

configs/mender_convert_config
```diff
@@ -102,6 +102,16 @@ MENDER_STORAGE_URL="https://d1b0l86ne08fsf.cloudfront.net"
...
+MENDER_SKIP_FIRST_PARTITION=0
```
mirzak: Can you please use,

```
MENDER_SKIP_FIRST_PARTITION="0"
```

That is a string value

kacf:
> I would also propose a different name for the config variable (`MENDER_SKIP_FIRST_PARTITION`), my proposal would be `MENDER_FORCE_GEN_BOOT_PART`

Is that the same thing though? The question is not whether you generate a boot partition, but whether you skip using the boot partition from the original image.

mirzak:
> Is that the same thing though? The question is not whether you generate a boot partition, but whether you skip using the boot partition from the original image.

Yeah, the difference is pretty small. `MENDER_FORCE_GEN_BOOT_PART` is more explicit imo while `MENDER_SKIP_FIRST_PARTITION` does say skip first partition but it does not tell you that it will generate a new one for you, and hence my proposal

kacf:
> Yeah, the difference is pretty small. `MENDER_FORCE_GEN_BOOT_PART` is more explicit imo while `MENDER_SKIP_FIRST_PARTITION` does say skip first partition but it does not tell you that it will generate a new one for you, and hence my proposal

Ok, but I think both names are a bit misleading then. This whole discussion makes me think that maybe this change isn't so great after all. @doudou, can you elaborate on what it was exactly that made the first partition unusable? Maybe we can detect it as a non-boot partition instead, and simply avoid this whole variable altogether.

doudou:
Honestly, don't know...

Also, I believe that trying to auto-detect whether the partition will always have corner-cases that are not working. In the end, we (users) have to configure mender-convert for our purposes, let us configure it.

As I said earlier in this thread:

> Generally speaking, it would make sense to just let the config file decide which partition is the 
> boot partition and which is the root partition, including letting mender-convert generate the first 
> boot partition.
>
> I really don't have time for this right now (read: within the next month), unfortunately.

A.k.a.: I think this is a stopgap measure. mender-convert should have a way to be told which boot & root partitions to use, including 'un-setting' the boot-partition variable to make mender-convert generate one (or have a MENDER_CONVERT_GEN_BOOT_PARTITION variable).

kacf:
What about some kind of mapping mechanism? Aka:

```
MENDER_BOOT_PARTITION_CREATION=auto | generate | keep
MENDER_INPUT_BOOT_PARTITION_NUMBER=1
MENDER_INPUT_ROOTFS_PARTITION_NUMBER=2
```

doudou:
> What about some kind of mapping mechanism? Aka:
> 
> ```
> MENDER_BOOT_PARTITION_CREATION=auto | generate | keep
> MENDER_INPUT_BOOT_PARTITION_NUMBER=1
> MENDER_INPUT_ROOTFS_PARTITION_NUMBER=2
> ```

That looks good

oleorhagen:
What is the status here?

mirzak:
I like proposal made by @kacf and we should create a task for this to follow up on as it might be out of scope of this PR.

This one might be related to this as well, https://tracker.mender.io/browse/MEN-3258

doudou:
My understanding is that this change will be implemented differently. I'm closing the PR.

PR #170, opened 2020-02-13
Fix broken link for Beaglebone U-Boot tools download.

kacf:
Changelog: None


kacf:
@mirzak: AFAICT this was just a typo. Does it make sense?

mirzak:
@drewmoseley would be better at answering this since ge did all the BBB stuff.

mirzak:
But looking at the history I would agree that it was a typo that was introduced here,

https://github.com/mendersoftware/mender-convert/commit/68b91cbeb286ffcadbd0ea7f8fcc098f96100859#diff-f7bf5b6fe54c41bff6066ed5f34184c7

drewmoseley:
Yes, I think this is correct.  It looks like @lluiscampos had some plans of versioning these tarballs but it seems that it was never implemented.  @lluiscampos can you comment?

kacf:
Thanks, @lluiscampos is off at the moment, but that means I at least have what I need.

kacf:
Cherry-picked to 2.0.x as well.

lluiscampos:
> Yes, I think this is correct. It looks like @lluiscampos had some plans of versioning these tarballs but it seems that it was never implemented. @lluiscampos can you comment?

Yep. I did have a plan but by the looks of it I didn't implemented it. The idea was to keep these 00x suffixes and then have the base one being a copy to the "latest".

Anyway, I think it is better to leave it as it is, as we most likely will have single "build" for each u-boot version.

Thank you all for solving this during my PTO ;)

PR #171, opened 2020-02-13
Remove container after conversion.

kacf:
Changelog: None


kacf:
@mirzak: Do you agree?

PR #173, opened 2020-02-24
Detect and warn about issues in MEN-2404

kacf:


modules/probe.sh
```diff
@@ -237,3 +242,40 @@ probe_initrd_in_boot_and_root() {
...
+
+  # Broken UEFI support in range v2018.09 - v2019.07 (see MEN-2404)
+  local regex='U-Boot 20(18\.(09|1[0-2])|19\.0[1-7])'
+
+  if egrep -qr "$regex" "$path"; then
+    local log_level=log_fatal
+    if [ "$MENDER_IGNORE_UBOOT_BROKEN_UEFI" = 1 ]; then
+      log_level=log_warn
+    fi
+    $log_level 'Detected a U-Boot version in the range v2018.09 - v2019.07. These U-Boot versions are known to have broken UEFI support, and therefore the MENDER_GRUB_EFI_INTEGRATION feature is unlikely to work. This only affects newly flashed devices using the partition image (extension ending in "img". The Mender artifact should still work to upgrade an existing, working device. There are two possible workarounds for this issue: 1) Use either an older or a newer image that works, and use a Mender artifact afterwards to up/down-grade it to the version you want. 2) If the device has a non-UEFI U-Boot port in mender-convert, use that (look for a board specific file in `configs`) . If you want to ignore this error and force creation of the image, set the MENDER_IGNORE_UBOOT_BROKEN_UEFI=1 config option.'
```
lluiscampos: Missing a parenthesis in `(extension ending in "img"`

kacf: Thanks, fixed!

kacf:
Fixed review comment, and also added one more commit to change our test image, since we were using one of the broken ones.

> I think we should update MEN-2404 description with the found conclusions (these two issues) so that any user that goes from these warnings/errors to JIRA can get the details there instead of following all the comments.

Good idea! I will make sure that gets done.

modules/probe.sh
```diff
@@ -237,3 +242,40 @@ probe_initrd_in_boot_and_root() {
...
+
+  # Broken UEFI support in range v2018.09 - v2019.07 (see MEN-2404)
+  local regex='U-Boot 20(18\.(09|1[0-2])|19\.0[1-7])'
+
+  if egrep -qr "$regex" "$path"; then
+    local log_level=log_fatal
+    if [ "$MENDER_IGNORE_UBOOT_BROKEN_UEFI" = 1 ]; then
+      log_level=log_warn
+    fi
+    $log_level 'Detected a U-Boot version in the range v2018.09 - v2019.07. These U-Boot versions are known to have broken UEFI support, and therefore the MENDER_GRUB_EFI_INTEGRATION feature is unlikely to work. This only affects newly flashed devices using the partition image (extension ending in "img"). The Mender artifact should still work to upgrade an existing, working device. There are two possible workarounds for this issue: 1) Use either an older or a newer image that works, and use a Mender artifact afterwards to up/down-grade it to the version you want. 2) If the device has a non-UEFI U-Boot port in mender-convert, use that (look for a board specific file in `configs`) . If you want to ignore this error and force creation of the image, set the MENDER_IGNORE_UBOOT_BROKEN_UEFI=1 config option.'
```
oleorhagen: Nitpick: Instead of this massive blob. Maybe split it up into multiple statements?
Example:

```bash
$log_level "some line"
$log_level "next line"
```

modules/probe.sh
```diff
@@ -237,3 +242,40 @@ probe_initrd_in_boot_and_root() {
...
+  case "$(probe_arch)" in
+    arm|aarch64)
+      # On ARM, as of version 2.04, GRUB can only boot kernels which have an EFI
+      # stub in them. See MEN-2404.
+      if ! file -k $kernel_path | fgrep 'EFI application'; then
+          local log_level=log_fatal
+          if [ "$MENDER_IGNORE_MISSING_EFI_STUB" = 1 ]; then
+              log_level=log_warn
+          fi
+          $log_level 'Detected a kernel which does not have an EFI stub. This kernel is not supported when booting with UEFI. Please consider using a U-Boot port if the board has one (look for a board specific file in `configs`), or find a kernel which has the CONFIG_EFI_STUB turned on. To ignore this message and proceed anyway, set the MENDER_IGNORE_MISSING_EFI_STUB=1 config option.'
```
oleorhagen: Same as above

modules/probe.sh
kacf: This was done on purpose. When grepping log lines you should get the whole context for that line. It doesn't look very nice in here, but when formatting log messages it's preferable to have the whole content.

For FWIW I tried breaking it into multiple concatenated strings, but this turned out to be quite difficult in shell, and ultimately made it look even more confusing.

PR #175, opened 2020-02-27
MEN-3172: [rpi] Fill all available space on the memory card

oleorhagen:
This commit adds a systemd script which on boot runs parted to fill the
remaining space on the memory card.

Changelog: Commit

### DIFF
--- a/configs/images/raspberrypi3_raspbian_config
+++ b/configs/images/raspberrypi3_raspbian_config
@@ -12,3 +12,37 @@ IMAGE_OVERHEAD_FACTOR=1.0
 
 # Best compression there is!
 MENDER_COMPRESS_DISK_IMAGE=lzma
+
+#
+# Resize the data partition to fill the remaining space, using parted, with systemd
+#
+function grow_data_partition() {
+  log_info "Adding systemd init script to run parted and resize the data partition on boot"
+  log_info "to fill all the available space on the storage media"
+  run_and_log_cmd "sudo mkdir -p work/rpi/etc/systemd/system/"
+  run_and_log_cmd "sudo mkdir -p work/rootfs/etc/systemd/system/data.mount.wants/"
+  cat <<-EOF > work/rpi/etc/systemd/system/mender-grow-data.service
+  [Unit]
+  Description=Mender service to grow data partition size
+  DefaultDependencies=no
+  Before=data.mount
+  Before=systemd-growfs@data.service
+
+  [Service]
+  Type=oneshot
+  User=root
+  Group=root
+  ExecStart=/sbin/parted --script /dev/mmcblk0 resizepart ${MENDER_DATA_PART_NUMBER} 100%
+
+  [Install]
+  WantedBy=data.mount
+EOF
+
+  # Install
+  run_and_log_cmd "cp work/rpi/etc/systemd/system/mender-grow-data.service \
+                      work/rootfs/etc/systemd/system/"
+  run_and_log_cmd "ln -sf work/rootfs/etc/systemd/system/mender-grow-data.service \
+                          work/rootfs/etc/systemd/system/data.mount.wants/"
+}
+
+PLATFORM_MODIFY_HOOKS+=(grow_data_partition)


configs/raspberrypi_config
```diff
@@ -95,4 +95,34 @@ function platform_modify() {
     run_and_log_cmd "sudo unlink work/rootfs/etc/rc3.d/S01resize2fs_once"
   fi
   run_and_log_cmd "sudo rm -f work/rootfs/etc/init.d/resize2fs_once"
+
```
oleorhagen: Review note:

Added to rasperrypi_config in general. Is this appropriate?

oleorhagen:
With PR: 

```
pi@raspberrypi:~$ df -h
Filesystem      Size  Used Avail Use% Mounted on
/dev/root       1.6G  1.2G  260M  83% /
devtmpfs        459M     0  459M   0% /dev
tmpfs           464M     0  464M   0% /dev/shm
tmpfs           464M  6.2M  457M   2% /run
tmpfs           5.0M  4.0K  5.0M   1% /run/lock
tmpfs           464M     0  464M   0% /sys/fs/cgroup
/dev/mmcblk0p1  253M   48M  205M  19% /uboot
/dev/mmcblk0p4  6.8G  3.0M  6.5G   1% /data
tmpfs            93M     0   93M   0% /run/user/1000
```

Without:
```
pi@raspberrypi:~$ df -h
Filesystem      Size  Used Avail Use% Mounted on
/dev/root       3.4G  1.3G  2.0G  39% /
devtmpfs        459M     0  459M   0% /dev
tmpfs           464M     0  464M   0% /dev/shm
tmpfs           464M  6.2M  457M   2% /run
tmpfs           5.0M  4.0K  5.0M   1% /run/lock
tmpfs           464M     0  464M   0% /sys/fs/cgroup
/dev/mmcblk0p1  253M   48M  205M  19% /uboot
/dev/mmcblk0p4  120M  1.6M  110M   2% /data

```

configs/raspberrypi_config
kacf: Good question... I think it's not very future proof. We expect that already in the next version of Raspbian, this will be solved, since the Systemd version should be high enough to take care of this. How about adding this to its own config file instead, and using that in the image that we produce for our downloads? The most important is that that image expands the data partition.

configs/raspberrypi_config
```diff
@@ -95,4 +95,34 @@ function platform_modify() {
...
+  Description=Mender service to grow data partition size
+  DefaultDependencies=no
+  Before=data.mount
+  Before=systemd-growfs@data.service
+
+  [Service]
+  Type=oneshot
+  User=root
+  Group=root
+  ExecStart=/sbin/parted --script /dev/mmcblk0 resizepart ${MENDER_DATA_PART_NUMBER} 100%
```
kacf: I think `/dev/mmcblk0` should not be hardcoded. But it doesn't look like mender-convert uses `MENDER_STORAGE_DEVICE`, which surprised me a bit. We only have `MENDER_STORAGE_DEVICE_BASE`. I suggest you use the value from `MENDER_STORAGE_DEVICE_BASE`, minus the `p`, if there is one (`/dev/sda` would not have it, for example).

configs/raspberrypi_config
lluiscampos: I also think it would be best with a separate config file.

configs/raspberrypi_config
lluiscampos: If we do this in a separate config file and is only going to be used for our converted image, I don't think it is an issue to hardcode `/dev/mmcblk0`

configs/raspberrypi_config
oleorhagen: Alright, goodie. I figured as much.

Next question. Should the config file then install this as it is sourced? Does anyone see any issues with this? 

Reason I'm asking, is it does not seem to me that adding multiple functions to the current hooks is possible? 

Meaning, that: In the new config file, this cannot be added to `platform_modify`, as this will override the hook added in `raspberrypi_config`? 

At least it seems to me to be this way. If yous agree, then my suggestion is to rewrite this into proper hooks, which one can add to?

configs/raspberrypi_config
oleorhagen: I think I agree with @lluiscampos here. This is very one-use-case specific

kacf: Actually, my idea was that it could be reusable outside RPi as well, but I agree that this is not important to have at all, and this simplifies it considerably. In that case I'm fine with hardcoding it.

But in that case I would move this into the `configs/images/raspberrypi3_raspbian_config` instead, since this is specific to our image.

configs/raspberrypi_config
kacf: Suggestion: Where it calls `platform_modify` today, instead change it into this:

```bash
for platform_modifier in $PLATFORM_MODIFY_HOOKS; do
    $platform_modifier
done
```

And then make sure `PLATFORM_MODIFY_HOOKS=platform_modify` is present in the base config file. Then you can add function names to this variable to call additional hooks.

oleorhagen: Yes, I've been playing around with this actually. But we'd have to change all the hook functions. If you are okay with this, I will go ahead and do it.

@mirzak any reservations?

oleorhagen: Hooks PR: https://github.com/mendersoftware/mender-convert/pull/176

If this is okayed, I will go ahead and rebase this PR on top of it.

oleorhagen:
:heavy_check_mark:  Checked and verified locally

configs/raspberrypi_config
oleorhagen: @lluiscampos what is your take on the `configs/images/raspberrypi3_raspbian_config` specification file and path?

lluiscampos: I think it belongs (temporary) to `raspberrypi3` but not necessary to our image only, so I would move it to an independent file but not in `../images`. For example `configs/raspberrypi3_data_expansion_config`.

If it is faster, let's discuss it in person and take a decision.

oleorhagen:
@kacf Checked [:heavy_check_mark: ], and verified on the device with the file in images, like discussed. Happy as a clam?

oleorhagen:
> This looks good to me.
> 
> I was hoping that we could extend this functionality to be more generally available and at least make an effort to enable this on any given system (if some conditions are met). Guess we can follow-up with this, can you create a separate ticket for this?

Yes, lets! By general, do you mean to more than Pi's I take it?

mirzak:
> Yes, lets! By general, do you mean to more than Pi's I take it?

Yes, any platform that meets a set of pre-conditions for this to work.

oleorhagen:
> > Yes, lets! By general, do you mean to more than Pi's I take it?
> 
> Yes, any platform that meets a set of pre-conditions for this to work.

Follow-up: https://tracker.mender.io/browse/MEN-3261

PR #176, opened 2020-02-28
Added hooks to mender-convert

oleorhagen:
This extends the current functionality of the platform_<stage> function
functionality into using hooks, so that each modification step can be called
from multiple configuration files.

The valid hooks are:

* PLATFORM_MODIFY_HOOKS
* PLATFORM_PACKAGE_HOOKS
* USER_LOCAL_MODIFY_HOOKS

and can be appended to as a regular bash array.

Changelog: Commit

### DIFF
--- a/mender-convert-modify
+++ b/mender-convert-modify
@@ -19,10 +19,12 @@
 function platform_modify() {
   true
 }
+PLATFORM_MODIFY_HOOKS=(platform_modify)
 
 function user_local_modify() {
   true
 }
+USER_LOCAL_MODIFY_HOOKS=(user_local_modify)
 
 function trap_exit() {
   echo "mender-convert-modify has finished. Cleaning up..."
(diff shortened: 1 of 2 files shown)


mender-convert-modify
```diff
@@ -19,10 +19,12 @@
 function platform_modify() {
   true
 }
+PLATFORM_MODIFY_HOOKS=(platform_modify)
```
oleorhagen: Review note: Decided to go with arrays instead of string lists. Cleaner imo

mender-convert-modify
```diff
@@ -233,10 +235,16 @@ ${MENDER_STORAGE_DEVICE_BASE}${MENDER_DATA_PART_NUMBER}   /data          auto
 EOF"
 
 log_info "Performing platform specific modifications (if any)"
-platform_modify
+for hook in "${PLATFORM_MODIFY_HOOKS[@]}"; do
+  log_info "Running hook: $hook"
+  eval $hook
```
oleorhagen: Review note: Strictly not necessary, but adds clarity I think.

oleorhagen:
> Is the `mender-image-tests` submodule update intentional?

Nope

oleorhagen:
And fixed! This was a weird one!

PR #179, opened 2020-03-04
Cherry men 3172

oleorhagen:


oleorhagen:
Hmm, I see. I was a little unsure, as we ended up putting it in the `configs/images` section, which I thought was mainly for building our own image for distribution.

lluiscampos:
> Hmm, I see. I was a little unsure, as we ended up putting it in the `configs/images` section, which I thought was mainly for building our own image for distribution.

Mmm, that is a very valid point! It would belong to the changelog of our own image, which we don't publish anywhere (and strictly speaking, we don't version). I think you are right and we should not add it to the tool's changelog. Thanks!

PR #182, opened 2020-03-05
COMFILE PI support

drewmoseley:
In the case of the COMFILE PI which is a customized system
with a RPI Compute Module, we need to add some gpio commands
into the boot.scr.


mirzak:
Changes look good to me, but you need `Changelog` tags on you commits.

drewmoseley:
Changelogs added.

PR #183, opened 2020-03-26
Rearrange the AWS credential logic to support project wide variables.

kacf:
The publishing variables have no prefix, and the tmp storage ones have
a TMP_STORAGE prefix, so adapt to that.

Changelog: None


kacf:
Cherry-picked to 2.0.x as well.

PR #184, opened 2020-04-01
MEN-2151: Add support for GPT partition tables

bakern:
Partition scheme can be forced by setting MENDER_PARTITION_SCHEME="gpt" or MENDER_PARTITION_SCHEME="dos", otherwise it will use the scheme from the input image.


mirzak:
Please have a look at https://github.com/mendersoftware/mender/blob/master/CONTRIBUTING.md.

Two formalities needed:
- You need to sign-off your commit
- You need to add a Changelog tag

Both of these are explained in the linked document

PR #185, opened 2020-04-01
MEN-2151: Add support for GPT partition tables

bakern:
Partition scheme can be forced by setting MENDER_PARTITION_SCHEME="gpt" or MENDER_PARTITION_SCHEME="dos", otherwise it will use the scheme from the input image.

Changelog: Add support for GPT partition tables

Signed-off-by: Nate Baker <bakern@gmail.com>

### DIFF
--- a/mender-convert-package
+++ b/mender-convert-package
@@ -223,9 +223,26 @@ rootfsb_end=$(( ${rootfsb_start} + ${rootfs_part_sectors} - 1 ))
 data_start=$(disk_align_sectors ${rootfsb_end} ${MENDER_PARTITION_ALIGNMENT} )
 data_end=$(( ${data_start} + ${data_part_sectors} - 1 ))
 
-# Create partition table. TODO: GPT support
-run_and_log_cmd "${PARTED} -s ${img_path} mklabel msdos"
-run_and_log_cmd "${PARTED} -s ${img_path} unit s mkpart primary fat32 ${boot_part_start} ${boot_part_end}"
+if [ -n "${MENDER_PARTITION_SCHEME}" ]; then
+  # Force partition scheme to the one from configs
+  partition_scheme="${MENDER_PARTITION_SCHEME}"
+  log_info "Forcing partition scheme defined by MENDER_PARTITION_SCHEME to: ${partition_scheme}"
+else
+  # By default use the partition scheme from the input image
+  partition_scheme=$(disk_get_part_value ${disk_image} 1 SCHEME)
+  log_info "Using input diskimage partition scheme (${partition_scheme})"
+fi
+
+# Create partition table.
+if [ "${partition_scheme}" == "gpt" ]; then
+  log_info "Writing GPT partition table"
+  run_and_log_cmd "${PARTED} -s ${img_path} mklabel gpt"
+  run_and_log_cmd "${PARTED} -s ${img_path} unit s mkpart ESP fat32 ${boot_part_start} ${boot_part_end}"
+else
+  log_info "Writing DOS (MBR) partition table"
+  run_and_log_cmd "${PARTED} -s ${img_path} mklabel msdos"
+  run_and_log_cmd "${PARTED} -s ${img_path} unit s mkpart primary fat32 ${boot_part_start} ${boot_part_end}"
+fi
 run_and_log_cmd "${PARTED} -s ${img_path} set 1 boot on"
 run_and_log_cmd "${PARTED} -s ${img_path} -- unit s mkpart primary ext2 ${rootfsa_start} ${rootfsa_end}"
 run_and_log_cmd "${PARTED} -s ${img_path} -- unit s mkpart primary ext2 ${rootfsb_start} ${rootfsb_end}"
--- a/mender-convert-package
+++ b/mender-convert-package
@@ -256,9 +256,13 @@ if [ "${MENDER_COPY_BOOT_GAP}" == "y" ]; then
 fi
 
 # Burn Partitions
+log_info "Writing boot partition image"
 disk_write_at_offset "${boot_part}" "${img_path}" "${boot_part_start}"
+log_info "Writing rootfsa partition image"
 disk_write_at_offset "${output_dir}/rootfs.img" "${img_path}" "${rootfsa_start}"
+log_info "Writing rootfsb partition image"
 disk_write_at_offset "${output_dir}/rootfs.img" "${img_path}" "${rootfsb_start}"
+log_info "Writing data partition image"
 disk_write_at_offset "${output_dir}/data.img" "${img_path}" "${data_start}"
 
 log_info "Performing platform specific package operations (if any)"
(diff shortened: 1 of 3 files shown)


mender-convert-package
```diff
@@ -223,9 +223,26 @@ rootfsb_end=$(( ${rootfsb_start} + ${rootfs_part_sectors} - 1 ))
...
-run_and_log_cmd "${PARTED} -s ${img_path} unit s mkpart primary fat32 ${boot_part_start} ${boot_part_end}"
+if [ -z "${MENDER_PARTITION_SCHEME}" ]; then
+  # By default use the partition scheme from the input image
+  partition_scheme=$(disk_get_part_value ${disk_image} 1 SCHEME)
+  log_info "Using input partition scheme (${partition_scheme})"
+else
+  # Force partition scheme to the one from configs
+  partition_scheme="${MENDER_PARTITION_SCHEME}"
+  log_info "Forcing partition scheme to ${partition_scheme}"
+fi
```
mirzak, line 234: Small nitpick, I would reverse this, e.g

```
if [ -n "${MENDER_PARTITION_SCHEME}" ]; then
  # Force partition scheme to the one from configs
  partition_scheme="${MENDER_PARTITION_SCHEME}"
  log_info "Forcing partition scheme to ${partition_scheme}"
else
# By default use the partition scheme from the input image
  partition_scheme=$(disk_get_part_value ${disk_image} 1 SCHEME)
  log_info "Using input partition scheme (${partition_scheme})"
fi
```

Makes it a bit easier to follow, no functional change. 

Would additionally update the log output to:

```
log_info "Forcing partition scheme defined by MENDER_PARTITION_SCHEME to: ${partition_scheme}"
```

and

```
log_info "Using input diskimage partition scheme (${partition_scheme})"
```

mender-convert-package
```diff
@@ -239,6 +256,7 @@ if [ "${MENDER_COPY_BOOT_GAP}" == "y" ]; then
 fi
 
 # Burn Partitions
+log_info "Writing boot, root, and data partitions"
```
mirzak, line 259: This change does not seem to be related, could you put this in a separate commit?

Additionally it would make sense to break this up, e.g

```
log_info "Writing boot partition image"
disk_write_at_offset "${boot_part}" "${img_path}" "${boot_part_start}"
log_info "Writing rootfsa partition image"
disk_write_at_offset "${output_dir}/rootfs.img" "${img_path}" "${rootfsa_start}"
....
```

modules/disk.sh
```diff
@@ -96,7 +96,7 @@ disk_align_sectors() {
 # $2 - destination file
 # $3 - offset in number of 512 sectors
 disk_write_at_offset() {
-  run_and_log_cmd "dd if=${1} of=${2} seek=${3} conv=sparse status=none"
+  run_and_log_cmd "dd if=${1} of=${2} seek=${3} conv=sparse,notrunc status=none"
```
mirzak, line 99: What is the motivation behind this change?

mender-convert-package
bakern: If you look in `mender-convert-modify` there are several instances of variables being checked and set this way, so I was just trying to follow those standards.  Basically "if this variable is not set then determine what it should be, otherwise use the set value."  For example `MENDER_DEVICE_TYPE` and `MENDER_GRUB_KERNEL_IMAGETYPE`.  I see your point in wanting the default option to be last though, so I can change it if you prefer.

Is it better to close this pull request and create a new one?  Sorry but I don't use GitHub that often, it seems like if I make a new commit it messes up the existing pull request (i.e. don't need both commits in there).

mender-convert-package
bakern: Yes I could make a different commit, it just seemed very minor.  While watching the output, if you're not writing the "boot gap" then the last log output you see is writing the partition table, so you get the feeling it's taking way too long without these extra statements.

modules/disk.sh
bakern: With GPT there should be a backup partition table at the end of the disk.  So before this change, `mender-convert-package` creates the output image file, writes the partition table, and when it writes the first partition data the file is truncated and the backup partition table is gone.

I don't think this change causes any worse performance, it leaves the backup partition table, and I think it makes troubleshooting easier.

An alternative would be to run `sgdisk -e` on the file after it's done (I haven't tested this on the image but it should work).  As a side note, if the output image is written to a disk that is larger than the image, the backup partition table will end up in the wrong place.  This can be fixed by running `sgdisk -e` and I have tested it.  The device will still boot without the backup partition table so I'm not sure if this important or not.  It might be good to either document this or make it automatic as part of any "first boot data partition resize" operations.  It doesn't seem like my images are resizing the data partition on first boot though, so I'm thinking maybe that is just for Yocto images?

bakern: I see now that the GROWFS option only works with systemd 242 and newer, so that's why it wasn't working for me (my test system has version 241).  Since that's part of systemd there's not really any script to tap into for this purpose.

I'm not sure if it would be better to just document this as an optional extra step when writing the image to the device, or if there's a good way to run the `sgdisk -e` command on first boot. Any suggestions?

mender-convert-package
mirzak: > For example MENDER_DEVICE_TYPE and MENDER_GRUB_KERNEL_IMAGETYPE. I see your point in wanting the default option to be last though, so I can change it if you prefer.

Up to you, a minor thing really.

mirzak: > Is it better to close this pull request and create a new one?

Normally you would just rebase and force push to the same branch. The PR would be updated accordingly.

mender-convert-package
mirzak: The change makes sense, just put it a different commit to keep the logically separated. You can have the commit in the same PR.

modules/disk.sh
mirzak: > I don't think this change causes any worse performance, it leaves the backup partition table, and I think it makes troubleshooting easier.

Makes sense, could you put this in a separate commit as well. I think your comments would be a good commit message

mirzak: > I see now that the GROWFS option only works with systemd 242 and newer

There was a discussions regarding this on MH today,

https://hub.mender.io/t/systemd-growfs-resize-to-128m-only/1446/7

mirzak:
The CI errors do not seem to be related to thus PR,

```
 upload failed: ./deploy.tar.gz to s3://mender-gitlab-tmp-storage/mender-convert/132410602/deploy.tar.gz An error occurred (InvalidAccessKeyId) when calling the CreateMultipartUpload operation: The AWS Access Key Id you provided does not exist in our records.
```

@lluiscampos, do you have any idea how to resolve this?

lluiscampos:
> The CI errors do not seem to be related to thus PR,
> 
> ```
>  upload failed: ./deploy.tar.gz to s3://mender-gitlab-tmp-storage/mender-convert/132410602/deploy.tar.gz An error occurred (InvalidAccessKeyId) when calling the CreateMultipartUpload operation: The AWS Access Key Id you provided does not exist in our records.
> ```
> 
> @lluiscampos, do you have any idea how to resolve this?

Something is not correct after our last key rotations. I pinged @kacf and we are looking into it :+1:

kacf:
Fixed! It was just a misconfiguration after the key rotation. Please restart the job.

PR #186, opened 2020-04-03
2.0.x cherry picks

mirzak:
@kacf @lluiscampos @drewmoseley , do you agree that we should cherry-pick all 3 here?

### DIFF
--- a/configs/mender_convert_config
+++ b/configs/mender_convert_config
@@ -65,6 +65,11 @@ MENDER_DATA_PART_GROWFS=y
 # input disk image
 MENDER_DEVICE_TYPE=""
 
+# The partition scheme to use for the converted image.
+# To override set this to "gpt" or "dos" (the scheme names used by fdisk and partx).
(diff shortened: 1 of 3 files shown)


drewmoseley:
Yeah, no issue from me.

kacf:
I think "Add support for GPT partition tables" should not be cherry-picked to 2.0.x. This is both a new feature, and a behavior change. I think we should save this for minor releases.

mirzak:
> This is both a new feature, and a behavior change.

My reasoning for including (even if I agree that it is a new feature) is that there is no behavioral change for the majority of the cases (MBR images) and it actually fixes a problem for hardware/firmware that requires the image to be GPT for it to boot.

But as we are probably releasing `2.1.0` and `2.0.1` at the same time, I guess it does not hurt to save this for the "minor release".

kacf:
I would argue for that, yes! :-)

PR #188, opened 2020-04-09
Ubuntu 18.04 support for x86-64

nandra:
This PR add support for converting x86-64 Ubuntu 18.04 installed on Giada VM23 (32 G eMMC storage) and then converted using mender-convert and flashed back. This is more less RFC as 3rd commit is IMO bit hack but cannot find better way how to override grub installation name (maybe some runtime stuff can be implemented). Thanks.

### DIFF
--- a/configs/ubuntu_x86-64_hdd_config
+++ b/configs/ubuntu_x86-64_hdd_config
@@ -1,0 +1,26 @@
+# This configuration can be used to run Ubuntu on a any x86-64 machine.
+#
+# This has been tested on images generated with the following command:
+#
+# Boot live USB and copy the data from internal storage (hdd, eMMC) e.g. for hdd:
+#
+# 	dd if=/dev/sda of=/media/ubuntu/ubuntu.img bs=1M count=10000 status=progress
+#
+# Converted with the following command:
+#
+#	MENDER_ARTIFACT_NAME=release-1 ./docker-mender-convert \
+#		--disk-image input/ubuntu.img \
+#		--config configs/ubuntu_x86-64_hdd_config \
+#		--overlay rootfs_overlay_demo/
+#
+# and image must be then copied back after conversion to internal storage (when booted again from live USB) e.g. for eMMC:
+#
+#	zcat /media/ubuntu/ubuntu-x86_64-mender.img.gz | sudo dd of=/dev/sda bs=1M status=progress
+
+MENDER_STORAGE_DEVICE_BASE=/dev/sda
+MENDER_DEVICE_TYPE="x86_64"
+
+MENDER_STORAGE_TOTAL_SIZE_MB=16000
+
+# Nothing to copy
+MENDER_COPY_BOOT_GAP="n"


kacf:
For the install path and the target name, I'm interested in knowing the motivation. Does the original not work, or is it just to follow convention?

nandra:
@kacf original doesn't work. Ubuntu in EFI partition contains directory ubuntu where are grub related stuff (also grub name isn't bootx64.efi but grubx64.efi). So that was the motivation. I know it's not nice and also e.g. for other OS like fedora or so will not work.

Original installed ubuntu image looks like this:
```
├── EFI
│   ├── BOOT
│   │   ├── BOOTX64.EFI
│   │   └── fbx64.efi
│   └── ubuntu
│       ├── BOOTX64.CSV
│       ├── fw
│       ├── fwupx64.efi
│       ├── grub.cfg
│       ├── grubx64.efi
│       ├── mmx64.efi
│       └── shimx64.efi
```

Thanks.

kacf:
Is Secure Boot enabled on this device? I wonder if it could be related to [this post](https://askubuntu.com/questions/342365/what-is-the-difference-between-grubx64-and-shimx64/342382#342382). Maybe we can fix this in a more future proof way by instead scanning for where `shimx64.efi` is, and if we find it, use that folder and name the bootloader `grubx64.efi`?

@drewmoseley, @mirzak: Thoughts?

nandra:
Yes EFI was enabled.

mirzak:
> Maybe we can fix this in a more future proof way by instead scanning for where shimx64.efi is, and if we find it, use that folder and name the bootloader grubx64.efi?

I think trying to determinate this dynamically would be nice and over time we would find all the edge cases. I would expect that most mature distributions have something similar as they are prepared for Secure Boot. 

But do we need to scan for the `shimx64.efi`? Why not just got for `grubx64.efi` directly? 

@nandra, I would recommend to take a look [modules/probe.sh](https://github.com/mendersoftware/mender-convert/blob/master/modules/probe.sh) which contains other functions that try to find files, e.g the Linux kernel image. What we propose is to do something similar to this for the `grubx64.efi` file as well.

nandra:
@mirzak ok thanks. I'll take a look and extend [this method](https://github.com/mendersoftware/mender-convert/blob/b79a279cc1a952a0db5bcc896c63ad66e53c3b74/modules/probe.sh#L103) for x86 (just to have it in one place). Thanks.

mirzak:
> @mirzak ok thanks. I'll take a look and extend this method for x86 (just to have it in one place).

Just note that you need two things:
1. Destination directory
2. File name

The linked method will give you 2. so you probably need to a create something new for 1.

kacf:
> But do we need to scan for the `shimx64.efi`? Why not just got for `grubx64.efi` directly?

Sure, even simpler!

drewmoseley:
To clarify, EFI is enabled here but not secure boot. (At least there is no requirement for it to be enabled).

Using shim is one option for SB but that is only needed if you are using the standard Microsoft Keys.  If you are using custom keys you can load grub directly.  If possible it would be nice if we dynamically scanned for both.

nandra:
@drewmoseley we're anyway replacing only grubx64.efi why to scan also for shim? Do I miss something? (I'm EFI/SB newbie though ;) ) Thanks.

nandra:
Added updated handling. Please comment if you have any objections. Many thanks.

modules/probe.sh
```diff
@@ -94,6 +94,50 @@ probe_debian_arch_name() {
...
 }
 
+# For x86 distros like e.g. Ubuntu try to find grubx64.efi
+# file and return it's path otherwise return empty path
+#
+# No input parameters and these work on the assumption that boot and root parts
+# are mounted at work/boot
+find_grub_efi_path() {
+  grub_path=$(sudo find work/boot -name grubx64.efi)
+  echo "${grub_path}"
```
kacf, line 104: `grub_path` is entirely redundant here, you can just output `find` directly.

nandra: :+1: updated and added changelogs

modules/probe.sh
```diff
@@ -94,6 +94,49 @@ probe_debian_arch_name() {
   echo "${deb_arch}"
 }
 
+# For x86 distros like e.g. Ubuntu try to find grubx64.efi
+# file and return it's path otherwise return empty path
+#
+# No input parameters and these work on the assumption that boot and root parts
+# are mounted at work/boot
+find_grub_efi_path() {
+  echo "$(sudo find work/boot -name grubx64.efi)"
```
kacf, line 103: I'm nitpicking a lot now, but my point was that `echo` is entirely unnecessary, you can just call `find` directly. :-)

nandra: I'm using this method in 2 places that is the reason why I put it to function ;)

kacf: You can still do that. This is what I mean:

```bash
find_grub_efi_path() {
  sudo find work/boot -name grubx64.efi
}
```

nandra: ok I get it now. Will update ;)

nandra:
@mirzak ok sure np -> I'll rename and fix it. It was pleasure to add this small code to mender-convert -> nice work :+1:

modules/probe.sh
nandra: Updated

nandra:
@mirzak Updated. Thanks.

bakern:
> Original installed ubuntu image looks like this:
> 
> ```
> ├── EFI
> │   ├── BOOT
> │   │   ├── BOOTX64.EFI
> │   │   └── fbx64.efi
> │   └── ubuntu
> │       ├── BOOTX64.CSV
> │       ├── fw
> │       ├── fwupx64.efi
> │       ├── grub.cfg
> │       ├── grubx64.efi
> │       ├── mmx64.efi
> │       └── shimx64.efi
> ```

@nandra can you show how the /boot/efi folder looks in the converted image?  What happens when it tries to boot?

I'm just wondering because in my debian source image it also uses grubx64.efi instead of bootx64.efi, and it works fine.  The only difference is my source image doesn't have the `EFI/BOOT` folder, it only has the `EFI/debian` folder (instead of ubuntu).

nandra:
@bakern what do you mean by - it works fine? You have x86 debian converted image (without those changes) and it works fine? Thanks for clarification.

After conversion using changes in this PR directory looks like:
```
├── EFI
│   ├── BOOT
│   │   ├── BOOTX64.EFI
│   │   ├── fbx64.efi
│   │   ├── grub.cfg
│   │   ├── mender_grubenv1
│   │   │   ├── env
│   │   │   ├── lock
│   │   │   └── lock.sha256sum
│   │   └── mender_grubenv2
│   │       ├── env
│   │       ├── lock
│   │       └── lock.sha256sum
│   └── ubuntu
│       ├── BOOTX64.CSV
│       ├── fw
│       ├── fwupx64.efi
│       ├── grub.cfg
│       ├── grubx64.efi
│       ├── mmx64.efi
│       └── shimx64.efi
```
Basically grubx64.efi is replaced with mender one and this one point then to /EFI/BOOT where grub.cfg + grubenv is stored. I verified it and converted image boots without issues.

mirzak:
@lluiscampos , any idea what is going on with the tests?

lluiscampos:
> @lluiscampos , any idea what is going on with the tests?

Fails locally also. It seems like `disco` have been removed from http://archive.ubuntu.com/ubuntu/dists/ completely! I mean, it is EOL, but I haven't seen it before that Ubuntu removes it from the archive so shortly after

Anyway, launching a new PR with updated base image. See #189

mirzak:
Thanks for looking in to it @lluiscampos

lluiscampos:
@nandra Could you please rebase this PR on top of latest master to get the CI fixes?

nandra:
@lluiscampos sure thing.

bakern:
> @bakern what do you mean by - it works fine? You have x86 debian converted image (without those changes) and it works fine? Thanks for clarification.

@nandra yes, exactly.  I'm not suggesting there's anything wrong with your PR, just trying to understand what the differences are.  It seems like it should either work for both or not work for both, but I must be missing something.

This is my original image:
 ```
 ├── EFI
 │   └── debian
 │       ├── BOOTX64.CSV
 │       ├── fbx64.efi
 │       ├── grub.cfg
 │       ├── grubx64.efi
 │       ├── mmx64.efi
 │       └── shimx64.efi
 ```

This is after mender-convert without your changes:
 ```
 ├── EFI
 │   ├── BOOT
 │   │   ├── bootx64.efi
 │   │   ├── grub.cfg
 │   │   ├── mender_grubenv1
 │   │   │   ├── env
 │   │   │   ├── lock
 │   │   │   └── lock.sha256sum
 │   │   └── mender_grubenv2
 │   │       ├── env
 │   │       ├── lock
 │   │       └── lock.sha256sum
 │   └── debian
 │       ├── BOOTX64.CSV
 │       ├── fbx64.efi
 │       ├── grub.cfg
 │       ├── grubx64.efi
 │       ├── mmx64.efi
 │       └── shimx64.efi
 ```

As you can see my original didn't have bootx64.efi either.  I don't want to distract you from this fix, just thought I'd mention it.

Thanks,
Nate

nandra:
I think for Debian conversion mender-convert create BOOT directory and system use this folder instead of debian directory for booting and that is the reason why it is working. I tried same but system won't boot (looks like it start always in /BOOT/ubuntu folder). So this could be difference. If I find some time I can play with Debian image. Do you have some config I can try? Thanks.

bakern:
> I think for Debian conversion mender-convert create BOOT directory and system use this folder instead of debian directory for booting and that is the reason why it is working. I tried same but system won't boot (looks like it start always in /BOOT/ubuntu folder).

I'm still new to Mender, so that's one thing I'm trying to grasp yet.  Doesn't mender-convert need to put its own Grub and EFI files in `EFI/BOOT`, and the system should boot from that?  If the system uses the files from your original `/ubuntu` folder, will it do what is needed?

With your converted image, when it boots do you see output like this?

```
Welcome to GRUB!

lock: OK
lock: OK
```

> Do you have some config I can try?

The config I'm using is pretty basic.  I did have to force the partition scheme to be GPT to boot on one device I'm working on though.  Here is the config:

```
MENDER_GRUB_EFI_INTEGRATION=y
MENDER_STORAGE_DEVICE=/dev/sda
MENDER_STORAGE_DEVICE_BASE=/dev/sda
MENDER_STORAGE_TOTAL_SIZE_MB=15272
MENDER_COPY_BOOT_GAP="n"
```

nandra:
> > I think for Debian conversion mender-convert create BOOT directory and system use this folder instead of debian directory for booting and that is the reason why it is working. I tried same but system won't boot (looks like it start always in /BOOT/ubuntu folder).
> 
> I'm still new to Mender, so that's one thing I'm trying to grasp yet. Doesn't mender-convert need to put its own Grub and EFI files in `EFI/BOOT`, and the system should boot from that? If the system uses the files from your original `/ubuntu` folder, will it do what is needed?
Yes this is right but for ubuntu it's seems bit different. It's described in details here: https://unix.stackexchange.com/questions/565615/efi-boot-bootx64-efi-vs-efi-ubuntu-grubx64-efi-vs-boot-grub-x86-64-efi-grub-efi

I tried manually copy mender grub to /EFI/BOOT/BOOTX64.efi but device doesn't boot anymore. 
> 
> With your converted image, when it boots do you see output like this?
> 
> ```
> Welcome to GRUB!
> 
> lock: OK
> lock: OK
> ```
> 
> > Do you have some config I can try?
> 
> The config I'm using is pretty basic. I did have to force the partition scheme to be GPT to boot on one device I'm working on though. Here is the config:
Thanks I'll give a try.
> 
> ```
> MENDER_GRUB_EFI_INTEGRATION=y
> MENDER_STORAGE_DEVICE=/dev/sda
> MENDER_STORAGE_DEVICE_BASE=/dev/sda
> MENDER_STORAGE_TOTAL_SIZE_MB=15272
> MENDER_COPY_BOOT_GAP="n"
> ```

@mirzak pls wait a bit with merging just want to verify if I didn't mix up something. Will post update when test ready.

nandra:
I've tested it in way that I replace BOOTX64.efi with mender grub and device cannot boot. Also tried to boot mmc option only (see image below) and it just shows '_' and doesn't boot. Only option works was mmc ubuntu.

IMG_0861

I tried to convert debian image without this changes and on my board it doesn't boot (only when convert using this PR). Puzzled even more ;). Ideas?

bakern:
> I tried to convert debian image without this changes and on my board it doesn't boot (only when convert using this PR). Puzzled even more ;). Ideas?

If the Mender devs have an opinion I will defer to them, I'm just getting started :)

I would be interested to see the partition scheme of the input and converted images though.  Can you share the partition tables?

```
fdisk -l mender-convert/input/ubuntu.img
fdisk -l mender-convert/deploy/ubuntu-mender.img (after unzipping)
```
Could you also show the files in the boot partition of the converted image without your changes?  This page shows how to do it: https://docs.mender.io/2.3/artifacts/modifying-a-disk-image.  Just do the same for the boot partition instead, and then share the file tree.

I'll try to convert an ubuntu image to test as soon as I get time.  I think many small/embedded systems have UEFI that don't behave the same, so you might need to add the boot option and browse to the bootx64.efi put there by mender-convert.

nandra:
I'll try tomorrow on IntelNUC. This is Giada VM23. Maybe I spot something. Thanks.

nandra:
Right. So tried with IntelNUC and finding are:
- I was able to run mender converted image with changes in this PR
- as @bakern pointed out that it should work also without it's true. I just took config and run mender_convert as usual and image boots also fine

- on Giada VM23 it doesn't boot without changes in this PR

@mirzak @drewmoseley ideas? I can drop search/replace commit and add only defconfig commit and then can investigate other issue. Thanks.

drewmoseley:
@nandra I'm not completely following.  Can you summarize the issue?  What is the downside of leaving that commit in, if it resolves issues on some platforms?

nandra:
@drewmoseley 
2 test setup:
- Intel NUC
** image generated by conversion using by this PR/ and without this PR boot fine

- Giada VM23
** image generated by conversion using by this PR boot fine
** image generated by conversion without this PR doesn't boot

So if we merge only first commit from this PR it will work fine on most of x86 systems (except Giada and this can be investigated later). 

Hope it's more clear now ;). Thanks.

drewmoseley:
Thanks. Sounds reasonable to me.

kacf:
@nandra: So shall I merge only the first commit?

nandra:
@kacf yes but there is also necessary  small change from mmcblk0 to sda (s for Giada which use emmc it doesn't work ;) ) If you like I can push updated version. Thanks.

kacf:
Yes, please do!

nandra:
Updated

mirzak:
> @mirzak: You agree?

Yep!

kacf:
Thanks, @nandra!

PR #191, opened 2020-04-16
Fix incorrect use of partition number

purushnayak:
The current code enumerates the partitions and uses the index of this
enumeration to query partition parameters rather than the actual
partition number that is reported and expected by partx. This might not
work in certain cases.

Example:
/dev/sda1
/dev/sda2
/dev/sda5

The partition number of the last partition is 5 but current code will
use 3. This PR fixes this so that the actual number reported by partx is
used for all operations rather than the enumerated index.

### DIFF
--- a/mender-convert-extract
+++ b/mender-convert-extract
@@ -65,7 +65,8 @@ if [ ! -f ${MKFS_VFAT} ]; then
   MKFS_VFAT="/sbin/mkfs.vfat"
 fi
 
-declare -i nr_of_parts=$(disk_get_nr_of_parts ${disk_image})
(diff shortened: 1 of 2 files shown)


lluiscampos:
Hi @purushnayak and thank you for your contribution!

The changes look good.

However, there are few formalities with your commit that are missing: you need to sign the commit and add a Changelog tag. See [contributing readme](https://github.com/mendersoftware/mender/blob/master/CONTRIBUTING.md#providing-pull-requests). Please fix this and update your PR, so that all the automated tests can run.

purushnayak:
Re-creating separate pull request after updating commit message.

lluiscampos:
Re-submitted in #194

PR #194, opened 2020-04-22
Fix incorrect use of partition number

purushnayak:
The current code enumerates the partitions and uses the index of this enumeration to query partition parameters rather than the actual partition number that is reported and expected by partx. This might not work in certain cases.

Example:
/dev/sda1
/dev/sda2
/dev/sda5

The partition number of the last partition is 5 but current code will use 3. This PR fixes this so that the actual number reported by partx is used for all operations rather than the enumerated index.

ChangeLog: Fix error when partitions numbers are not sequential

Signed-off-by: Purushotham Nayak <purushotham.m.nayak@gmail.com>

### DIFF
--- a/mender-convert-extract
+++ b/mender-convert-extract
@@ -65,7 +65,8 @@ if [ ! -f ${MKFS_VFAT} ]; then
   MKFS_VFAT="/sbin/mkfs.vfat"
 fi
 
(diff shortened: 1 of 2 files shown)


lluiscampos:
Re-submitted at #218

PR #197, opened 2020-04-26
Setting version now works in docker if project added as git submodule

dellgreen:
- Issue: #MEN-3475

Changelog: Title

Signed-off-by: Dell Green <dell.green@ideaworks.co.uk>


dellgreen:
https://tracker.mender.io/browse/MEN-3475

PR #199, opened 2020-04-27
MEN-3271: Add upload of RaspberryPi 4 image.

kacf:


.gitlab-ci.yml
```diff
@@ -156,21 +183,34 @@ publish:s3:
     - export AWS_SECRET_ACCESS_KEY=$TMP_STORAGE_AWS_SECRET_ACCESS_KEY
     - apt update && apt install -yyq awscli
     # Fetch artifacts from temporary S3 bucket
-    - aws s3 cp s3://mender-gitlab-tmp-storage/$CI_PROJECT_NAME/$CI_PIPELINE_ID/deploy.tar.gz deploy.tar.gz
-    - tar xzf deploy.tar.gz
+    - aws s3 cp s3://mender-gitlab-tmp-storage/$CI_PROJECT_NAME/$CI_PIPELINE_ID/raspberrypi3.tar.gz raspberrypi3.tar.gz
+    - tar xzf raspberrypi3.tar.gz
```
lluiscampos, line 187: Missing fetch/extract for raspberry4 tarball.

.gitlab-ci.yml
```diff
@@ -77,15 +120,32 @@ build:
...
+  variables:
+    RASPBERRYPI_PLATFORM: raspberrypi3
+
+test_acceptance_prebuilt_raspberrypi4:
+  <<: *test_acceptance_prebuilt_raspberrypi
+  dependencies:
+    - build
+    - convert_raspbian_raspberrypi4
+  variables:
+    RASPBERRYPI_PLATFORM: raspberrypi4
```
lluiscampos, line 148: Really nice use of two-level template :+1:

.gitlab-ci.yml
```diff
@@ -156,21 +183,34 @@ publish:s3:
...
+    - tar xzf raspberrypi3.tar.gz
   script:
     # Prepare high privilege S3 keys (the TMP_STORAGE keys are for the tmp storage only)
     - export AWS_ACCESS_KEY_ID=$PUBLISH_AWS_ACCESS_KEY_ID
     - export AWS_SECRET_ACCESS_KEY=$PUBLISH_AWS_SECRET_ACCESS_KEY
-    - echo "Publishing ${RASPBIAN_NAME}-mender-${MENDER_CLIENT_VERSION}.img.xz version to S3"
-    - aws s3 cp deploy/${RASPBIAN_NAME}-raspberrypi3-mender.img.xz
-        s3://$S3_BUCKET_NAME/${RASPBIAN_NAME}/arm/${RASPBIAN_NAME}-mender-${MENDER_CLIENT_VERSION}.img.xz
+
+    - echo "Publishing ${RASPBIAN_NAME}-raspberrypi4-mender-${MENDER_CLIENT_VERSION}.img.xz version to S3"
```
lluiscampos, line 193: `raspberrypi3` here

lluiscampos: I guess the best would be to have the string `${RASPBIAN_NAME}-raspberrypi4-mender-${MENDER_CLIENT_VERSION}.img.xz` into an env variable, so that we are certain we echo and cp the same thing

.gitlab-ci.yml
kacf: Good catch, fixing!

.gitlab-ci.yml
kacf: Good point. I took the opportunity to make this even better, and turned both sections in the publish stage into loops. You can actually make loops across several list entries in the yaml. I think that wasn't possible in the past, I can swear I tried it a while ago. But it works now, so take a look!

.gitlab-ci.yml
```diff
@@ -156,21 +183,30 @@ publish:s3:
...
-    - echo "Publishing ${RASPBIAN_NAME}-mender-${MENDER_CLIENT_VERSION}.mender version to S3"
-    - aws s3 cp deploy/${RASPBIAN_NAME}-raspberrypi3-mender.mender
-        s3://$S3_BUCKET_NAME/${RASPBIAN_NAME}/arm/${RASPBIAN_NAME}-mender-${MENDER_CLIENT_VERSION}.mender
-    - aws s3api put-object-acl --acl public-read --bucket $S3_BUCKET_NAME
-        --key ${RASPBIAN_NAME}/arm/${RASPBIAN_NAME}-mender-${MENDER_CLIENT_VERSION}.mender
+
+    - for RASPBERRYPI_PLATFORM in raspberrypi3 raspberrypi4; do
+    -   PUBLISH_NAME=${RASPBIAN_NAME}-${RASPBERRYPI_PLATFORM}-mender-${MENDER_CLIENT_VERSION}.img.xz
+    -   echo "Publishing ${PUBLISH_NAME} version to S3"
+    -   aws s3 cp ${RASPBERRYPI_PLATFORM}/${RASPBIAN_NAME}-${RASPBERRYPI_PLATFORM}-mender.img.xz
```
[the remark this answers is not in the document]
kacf: No, this is correct. The original name as coming from mender-convert is not the same as the name we publish it under.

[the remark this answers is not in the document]
lluiscampos: You are right, I can see it now in the old code.

kacf:
Merged. I will keep an eye on the publish stage.

PR #200, opened 2020-04-27
configs: Added centos x86-64 hdd defconfig

nandra:
Changelog: Title

Setup for converting CentOS Linux 8 (test device was used Intel NUC)
with 128G SSD.


nandra:
Just thinking if won't be better to create one generic x86 defconfig. This config is same just image name is different. Open to suggestions. Thanks.

kacf:
Yes, I'm thinking the same thing. x86 should be pretty uniform.

nandra:
@kacf updated. Thanks.

kacf:
Thanks, @nandra!

PR #201, opened 2020-05-08
Update to new download source for grub-efi, and update grub-mender-grubenv.

kacf:
From grub-mender-grubenv changelog:
Changelog: add 'rootwait' to bootargs
Changelog: grubenv: Handle debug command prompt when running as EFI app.
Changelog: utilize regexp to dynamically set mender_grub_storage_device

### DIFF
--- a/configs/mender_grub_config
+++ b/configs/mender_grub_config
@@ -3,6 +3,10 @@
 # Only disable this if you know what you are doing
 MENDER_GRUB_EFI_INTEGRATION=y
 
+# Version of GRUB to use. Note that there needs to be a precompiled version
+# available at the MENDER_STORAGE_URL download source.
+GRUB_VERSION=2.04
+
 # Specific Linux kernel boot arguments
 #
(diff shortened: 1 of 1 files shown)


configs/mender_grub_config
```diff
@@ -13,7 +17,7 @@ MENDER_GRUB_EFI_INTEGRATION=y
 MENDER_GRUB_KERNEL_BOOT_ARGS=""
 
 # grub-mender-grubenv is the Mender integration for the GRUB bootloader
-MENDER_GRUBENV_VERSION="1.3.0"
+MENDER_GRUBENV_VERSION="6ffc280e64dd7b6bbc52f7747609f11fbdd1a057"
```
lluiscampos, line 20: Same here, I think this checksum is outdated.

lluiscampos: Okey... it is already fixed in https://github.com/mendersoftware/mender-convert/pull/203/ :)

PR #202, opened 2020-05-08
Remove kernel_devicetree from EFI path.

kacf:
This information is not used when loading via UEFI, instead it is
queried directly from the UEFI provider.

Changelog: Commit


kacf:
Indeed!

PR #204, opened 2020-05-15
MEN-3527: Fix regressions introduced during refactoring

oleorhagen:
This fixes misc minor regressions introduced during my refactoring

* Declare the required env variables to empty string
* Create the work-dir before running any conversion commands
* Remove MENDER_CONVERT_VERSION definition from mender-convert
* Improve the error message on unsupported image types

Also, some of these only shows up when not running this in the docker container, which we are not testing at all.

I guess we should start testing this also. Question is then: What exactly? Just one of the boards as a smoke test? I guess all of them is a little over the top.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ rootfs_overlay_demo/*
 mender_local_config
 *.xml
 *.html
+convert.log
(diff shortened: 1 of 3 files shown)


dellgreen:
have just tested iit using the same build process that was previously not working and it now all executes correctly. many thanks

oleorhagen:
> have just tested it using the same build process that was previously not working and it now all executes correctly. many thanks

Great to hear!

Thanks for reporting

PR #206, opened 2020-05-18
Misc changes for Yocto

drewmoseley:
I am attempting to use mender-convert on a pre-built Yocto image for intel.  I needed these changes as a start to support the existing intel-corei7-64 builds.

### DIFF
--- a/configs/generic_x86-64_hdd_config
+++ b/configs/generic_x86-64_hdd_config
@@ -38,3 +38,14 @@ MENDER_STORAGE_TOTAL_SIZE_MB=16000
 
 # Nothing to copy
 MENDER_COPY_BOOT_GAP="n"
+
+function platform_modify() {
+    #
+    # Make sure /lib64 exists since the Mender binary requires it.
+    # Some systems put everything under /lib (ie Yocto) and a simple
+    # symlink is enough to find everything Mender needs.
+    #
+    if [ ! -e work/rootfs/lib64 ]; then
+        run_and_log_cmd "ln -s /lib work/rootfs/lib64"
+    fi
+}
--- a/modules/decompressinput.sh
+++ b/modules/decompressinput.sh
@@ -28,7 +28,7 @@ function compression_type () {
   fi
   local -r disk_image="${1}"
   case "${disk_image}" in
-    *.img | *.sdimg)
+    *.img | *.sdimg | *.wic | *.rpi-sdimg)
       echo "none"
       ;;
     *.gz)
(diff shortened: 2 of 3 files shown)


modules/decompressinput.sh
```diff
@@ -28,7 +28,7 @@ function compression_type () {
   fi
   local -r disk_image="${1}"
   case "${disk_image}" in
-    *.img | *.sdimg)
+    *.img | *.sdimg | *.ext4 | *.wic | *.rpi-sdimg)
```
oleorhagen, line 31: Never seen the `rpi-sdimg` before. What's the difference there?

configs/generic_x86-64_hdd_config
```diff
@@ -38,3 +38,14 @@ MENDER_STORAGE_TOTAL_SIZE_MB=16000
...
 MENDER_COPY_BOOT_GAP="n"
+
+function platform_modify() {
+    #
+    # Make sure /lib64 exists since the Mender binary requires it.
+    # Some systems put everything under /lib (ie Yocto) and a simple
+    # symlink is enough to find everything Mender needs.
+    #
+    if [ ! -e work/rootfs/lib64 ]; then
+        run_and_log_cmd "ln -s /lib work/rootfs/lib64"
```
oleorhagen, line 49: It's early morning here, but will this not link `/lib` locally to the work directory. Is this right?

mirzak: Yes, but also to `/lib` on the device where the link is stored. :)

oleorhagen: I was almost sure I was wrong here. Just had to double check :p

mirzak: Clarification, the link will be `lib64 -> /lib` and the link is stored in `work/rootfs/` so on the device it should be `/lib64 -> /lib`. So I think this is correct.

But question is, why do we need this?

kacf: It's because of [this bug](https://tracker.mender.io/browse/MEN-2799).

modules/decompressinput.sh
mirzak: `wic` and `rpi-sdimg` are image types. Just suffix that that is different, but they are disk images.

But `ext4` stands out to me. This is not an disk image, and probably should not be used for "disk image" input.

drewmoseley: I thought we could function on an ext4 image.  Is that not correct?

mirzak: No, we currently do not have support for that.

drewmoseley: OK.  I removed ext4.

drewmoseley:
@mirzak ok to merge?

PR #207, opened 2020-05-28
Switch Docker container from 19.10 (EOL) to 20.04.

kacf:
This also requires us to use noninteractive install environment
because of new timezone selection in tzdata package.

Changelog: Fix 404 download errors when trying to run `docker-build`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # Build pxz in separate image to avoid big image size
-FROM ubuntu:19.10 AS build
-RUN apt-get update && apt-get install -y \
+FROM ubuntu:20.04 AS build
+RUN apt-get update && env DEBIAN_FRONTEND=noninteractive apt-get install -y \
     build-essential \
     git \
     liblzma-dev
@@ -9,11 +9,11 @@ RUN apt-get update && apt-get install -y \
 RUN git clone https://github.com/jnovy/pxz.git /root/pxz
 RUN cd /root/pxz && make
 
-FROM ubuntu:19.10
+FROM ubuntu:20.04
 
 ARG MENDER_ARTIFACT_VERSION=3.2.1
 
-RUN apt-get update && apt-get install -y \
+RUN apt-get update && env DEBIAN_FRONTEND=noninteractive apt-get install -y \
 # For 'ar' command to unpack .deb
     binutils \
     xz-utils \


kacf:
I thought this was already fixed in master, but apparently not.

mirzak:
> I thought this was already fixed in master, but apparently not.

It was with below,

https://github.com/mendersoftware/mender-convert/commit/e170109121eb34dc649f1f5d5eb354b80c8d6056

Above is cherry-picked to `2.0.x` as well.

But I guess an update to 20.04 is better.

Dockerfile
```diff
@@ -1,6 +1,6 @@
 # Build pxz in separate image to avoid big image size
-FROM ubuntu:19.10 AS build
-RUN apt-get update && apt-get install -y \
+FROM ubuntu:20.04 AS build
+RUN apt-get update && env DEBIAN_FRONTEND=noninteractive apt-get install -y \
```
oleorhagen, line 3: Why not `Dockerfile` `ENV` syntax?

Dockerfile
```diff
@@ -9,11 +9,11 @@ RUN apt-get update && apt-get install -y \
 RUN git clone https://github.com/jnovy/pxz.git /root/pxz
 RUN cd /root/pxz && make
 
-FROM ubuntu:19.10
+FROM ubuntu:20.04
```
oleorhagen, line 12: Thought:
Why not just go with `latest` here? There's not really much that could break I think.

Dockerfile
kacf: Yes, I suppose that is better, if a new apt command is added later. Will change it.

Dockerfile
mirzak: Sounds risky, `2020.04` is an LTS, so I guess we should be fine for a while.

```
```

kacf: We should stick to stable releases I think, and 20.04 is an LTS. That's one of the reasons for using a container in the first place. `mkfs.ext4` has actually broken several times in the past.

oleorhagen: haha, I knew this was gonna get downvoted to hell. So risk averse!

kacf:
> It was with below,
> 
> [e170109](https://github.com/mendersoftware/mender-convert/commit/e170109121eb34dc649f1f5d5eb354b80c8d6056)
> 
> Above is cherry-picked to `2.0.x` as well.
> 
> `19.04` is EOL, not `19.10`
> 
> But I guess an update to 20.04 is better.

Yes, I think so. I think 19.10 is also EOL, or at least very soon will be.

oleorhagen:
https://wiki.ubuntu.com/Releases

Dockerfile
kacf: On second thought, maybe not. The Dockerfile reference [explicitly recommends not setting `DEBIAN_FRONTEND=noninteractive` using `ENV`](https://docs.docker.com/engine/reference/builder/#env). We are doing this in a lot of other places though, but I guess we can at least follow the advice here.

oleorhagen: Sure. Still, we don't really have interactive users of our image though

oleorhagen: Interesting that this actually found a place in the documentation :p

Dockerfile
kacf: Working too long in release management! ;-)

mirzak: So you are basically just trolling @oleorhagen ?

oleorhagen: Just a reminder that I am yet young

PR #211, opened 2020-06-05
Misc changes from Reliagate Integration

drewmoseley:


modules/probe.sh
```diff
@@ -195,7 +195,12 @@ probe_kernel_in_boot_and_root() {
   done
 
   if [ -z "${kernel_imagetype_path}" ]; then
-    log_warn "Unfortunately we where not able to find the Linux kernel image."
+    log_warn "Unfortunately we were not able to find the Linux kernel image."
+    log_fatal "Please specify the image name using MENDER_GRUB_KERNEL_IMAGETYPE"
+  fi
+
+  if [ "$(echo "${kernel_imagetype_path}" | wc -w)" -gt "1" ]; then
+    log_warn "Unfortunately we were not able to find a unique Linux kernel image."
```
mirzak, line 203: There is a separate [PR](https://github.com/mendersoftware/mender-convert/pull/212) to fix this, so would just drop this changeset

modules/disk.sh
```diff
@@ -96,7 +96,7 @@ disk_align_sectors() {
 # $2 - destination file
 # $3 - offset in number of 512 sectors
 disk_write_at_offset() {
-  run_and_log_cmd "dd if=${1} of=${2} seek=${3} conv=sparse,notrunc status=none"
+  run_and_log_cmd "dd if=${1} of=${2} seek=${3} conv=notrunc status=none"
```
mirzak, line 99: I do not understand how this would cause problems? Can you give an example?

configs/generic_x86-64_hdd_config
```diff
@@ -39,6 +39,9 @@ MENDER_STORAGE_TOTAL_SIZE_MB=16000
 # Nothing to copy
 MENDER_COPY_BOOT_GAP="n"
 
+# Not using UBoot.
+MENDER_GRUB_EFI_INTEGRATION="y"
+
```
mirzak, line 44: This is the default value, as specified in,

https://github.com/mendersoftware/mender-convert/blob/7b6d957170236f139e5022d0d7d74a726ad6bae3/configs/mender_convert_config#L175

Do not see a huge benefit of being explicit about it here.

mender-convert-modify
```diff
@@ -120,7 +120,7 @@ if [ "${MENDER_GRUB_EFI_INTEGRATION}" == "y" ]; then
   if [ -z "${MENDER_GRUB_INITRD_IMAGETYPE}" ]; then
```
mirzak, line 120: Missing `Changelog:`

modules/probe.sh
drewmoseley: Yeah, I saw that over the weekend and like the other solution better.

configs/generic_x86-64_hdd_config
drewmoseley: Dang. My wrapper script didn't include that config file.  Agreed this is unneeded.

modules/disk.sh
drewmoseley: I didn't understand it either but it definitely did make a difference.  Last week when I tested it. ;-)
Now for some reason it doesn't seem to be needed. Who knows.

drewmoseley:
Updated as discussed.

PR #219, opened 2020-07-03
MEN-3762 chmod 600 on mender.conf

merlin-northern:
ChangeLog:title
Signed-off-by: Peter Grzybowski <peter@northern.tech>

### DIFF
--- a/mender-convert-modify
+++ b/mender-convert-modify
@@ -193,6 +193,7 @@ cat <<- EOF > work/mender.conf.data
 EOF
 
 run_and_log_cmd "sudo cp work/mender.conf.data work/rootfs/data/mender/mender.conf"
+run_and_log_cmd "sudo chmod 600 work/rootfs/data/mender/mender.conf"
 
 if [ -z "${MENDER_DEVICE_TYPE}" ]; then
   # Observed systems who do not have this file, e.g images generated with mkosi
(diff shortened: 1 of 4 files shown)


mender-convert-modify
```diff
@@ -193,6 +193,7 @@ cat <<- EOF > work/mender.conf.data
 EOF
 
 run_and_log_cmd "sudo cp work/mender.conf.data work/rootfs/data/mender/mender.conf"
+run_and_log_cmd "chmod 600 work/rootfs/data/mender/mender.conf"
```
kacf, line 196: I think you need `sudo` here.

merlin-northern: > I think you need `sudo` here.

ah, crap, copy-pasted from the line above, and removed sudo by mistake. [thanks](https://github.com/mendersoftware/mender-convert/pull/219/commits/564d263f53be0cc4b0b69e3f45f6153541462ae3)

merlin-northern:
> Thanks!
> 
> I realized now though, that this only changes one of the two `mender.conf` files. This is the one in `/data`, you also have [several different variants in `/etc`](https://github.com/mendersoftware/mender-convert/search?q=%22%2Fetc%2Fmender%2Fmender.conf%22&unscoped_q=%22%2Fetc%2Fmender%2Fmender.conf%22).

yeah I checked those. but it is redirection, it should not change the existing conf file mode. or do you think it is possible that those redirections create the file?

kacf:
They are definitely creating the file, since the input image is an image without Mender.

merlin-northern:
> They are definitely creating the file, since the input image is an image without Mender.

ok! [fixed](https://github.com/mendersoftware/mender-convert/pull/219/commits/198e22767a7410c96eda0ba07cf605daf2224c5c)

PR #220, opened 2020-07-03
Partition UUID support added for gpt/dos partition tables for deterministic booting

dellgreen:
Partition UUID support added for gpt and dos partition tables for deterministic booting
    
[Issue: MEN-3725](https://tracker.mender.io/browse/MEN-3725)    

The following new config variables have been added which closely mirror those used in meta-mender
to give fine grained controll over partition device paths. If partuuid support is enabled then
these partition variables must be used as format doesnt fit device/number variables scheme.
If partuuid support is disabled then partition variables are optional liek in meta-mender.
    

- MENDER_ENABLE_PARTUUID
- MENDER_BOOT_PART
- MENDER_ROOTFS_PART_A
- MENDER_ROOTFS_PART_B
- MENDER_DATA_PART

```
Example:
MENDER_BOOT_PART="/dev/disk/by-partuuid/26445670-f37c-408b-be2c-3ef419866620" #gpt
MENDER_BOOT_PART="/dev/disk/by-partuuid/26445670-01" #dos
MENDER_BOOT_PART="/dev/sda1" #partuuid support disabled
```

### DIFF
--- a/modules/disk.sh
+++ b/modules/disk.sh
@@ -178,3 +178,186 @@ disk_root_part() {
   fi
   echo "${root_part}"
 }
+
+
+# Check if supplied argument is valid partuuid device path.
+# Supports both dos and gpt paths
+#
+# $1 - partuuid device path
+disk_is_valid_partuuid_device() {
+  disk_is_valid_partuuid_gpt_device "$1" || disk_is_valid_partuuid_dos_device "$1"
+}
+
+
+# Check if supplied argument is valid gpt partuuid device path.
+#
+# Example: /dev/disk/by-partuuid/26445670-f37c-408b-be2c-3ef419866620
+#
+# $1 - gpt partuuid device path
+disk_is_valid_partuuid_gpt_device() {
+  echo "${1}" | grep -qE '^/dev/disk/by-partuuid/([0-9a-f]{8}(-[0-9a-f]{4}){3}-[0-9a-f]{12})$'
+}
+
+
+# Check if supplied argument is valid dos partuuid device path.
+#
+# Example: /dev/disk/by-partuuid/26445670-01
+#
+# $1 - dos partuuid device path
+disk_is_valid_partuuid_dos_device() {
+  echo "${1}" | grep -qE '^/dev/disk/by-partuuid/[0-9a-f]{8}-[0-9a-f]{2}$'
+}
+
+
+# Get partuuid from supplied device path.
+# Supports both dos and gpt paths
+#
+# $1 - partuuid device path
+disk_get_partuuid_from_device() {
+  if ! disk_is_valid_partuuid_device "${1}"; then
+    log_fatal "Invalid partuuid device: '${1}'"
+  fi
+  echo "${1}" | sed "s:/dev/disk/by-partuuid/::"
+}
+
+
+# Get dos disk identifier from supplied device path.
+#
+# $1 - dos compatible partuuid device path
+disk_get_partuuid_dos_diskid_from_device() {
+  if ! disk_is_valid_partuuid_dos_device "${1}"; then
+    log_fatal "Invalid dos partuuid device: '${1}'"
+  fi
+  partuuid=$(disk_get_partuuid_from_device "${1}")
+  echo "$partuuid" | cut -d- -f1
+}
+
+
+# Get dos partuuid number from supplied device path.
+#
+# $1 - dos compatible partuuid device path
+disk_get_partuuid_dos_part_number() {
+  if ! disk_is_valid_partuuid_dos_device "${1}"; then
+    log_fatal "Invalid dos partuuid device: '${1}'"
+  fi
+  partuuid=$(disk_get_partuuid_from_device "${1}")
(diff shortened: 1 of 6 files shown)


modules/disk.sh
```diff
@@ -178,3 +178,182 @@ disk_root_part() {
...
+        echo $dev_base
+    fi
+}
+
+
+# Conditionally redefine partition number/device if fine grained variable set.
+#
+# $1 - variable name
+# $2 - variable value
+disk_overide_partition_variable() {
```
kacf, line 348: Nitpick: `overide -> override`

modules/disk.sh
```diff
@@ -178,3 +178,182 @@ disk_root_part() {
...
+        /dev/[sh]d[a-z][1-9]* )
+            dev_part=${1##*d[a-z]}
+            ;;
+        ubi*_* )
+            dev_part=$(echo $1 | cut -d_ -f2)
+            ;;
+        /dev/disk/by-partuuid/* )
+            if disk_is_valid_partuuid_dos_device "$1";then
+              dev_part=$(disk_get_partuuid_dos_part_from_device "$1")
+              dev_part=$((16#${dev_part}))
```
kacf, line 299: I'm a bit puzzled by this one. What is the result here? Is it supposed to return the full UUID string? If so, how does `$((16#${dev_part}))` cope with the hyphens in the string? And isn't this supposed to return the partition index in the table, not the UUID?

dellgreen: The following gets the partition number only from the path:
` dev_part=$(disk_get_partuuid_dos_part_from_device "$1")`
  
The following converts partition number from hex to integer
` dev_part=$((16#${dev_part}))`

on closer inspection i can see the comment is misleading on this function and needs to be changed

dellgreen:
hopefully these recent changes make it a bit clearer

modules/disk.sh
kacf: A little bit, but the part I don't get is the partition number itself. Let's take an example:

1. Take a random partuuid device, `/dev/disk/by-partuuid/2b706a90-c021-11ea-b531-e3e6704f570c`.
2. `dev_part=$(disk_get_partuuid_dos_part_number "$1")` will return `c021`
3. `dev_part=$((16#${dev_part}))` will return `49185`.

But why is this the partition number? It doesn't seem correct, it's just a sub part extracted from the UUID. Are both `sgdisk` and `grub-mender-grubenv` expected to recognize this and treat it correctly?

dellgreen: Based on my understanding of the kernel docs:

"1. Take a random partuuid device, /dev/disk/by-partuuid/2b706a90-c021-11ea-b531-e3e6704f570c."
A) The format used for a dos/mbr partuuid is not the same as the gpt format shown above, as its not embedded in the partition table, the kernel uses the dos disk identifier followed by a hypen and then the partition index in hex to generate a pseudo-partuuid at runtime.

kacf: This may very well be the information I was missing. But where did you find that? This is new to me.

dellgreen: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=d33b98fc82b0908e91fb05ae081acaed7323f9d2
https://wiki.archlinux.org/index.php/Persistent_block_device_naming
https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-core/classes/mender-part-images.bbclass#L217

dellgreen: to be honest i didn't know this existed either until I looked into it. :)

kacf: Thanks, that helped a lot! I see now that I missed that `disk_is_valid_partuuid_dos_device` checks for this format, so I could have looked there. Spotting small differences in regexes is not my strong suit! :wink:

I suggest that you put an example of a UUID of both types in the `disk_is_valid_partuuid_gpt_device` and `disk_is_valid_partuuid_dos_device` functions, respectively, so that it's a bit clearer which format is accepted by them. This should make it more obvious that the trailing part of the dos variant is the partition index.

kacf: I think that other than what I just mentioned, the PR is ready to go.

dellgreen: Ok great, i have made the recommended changes. 
Looks like the beaglebone gitlab pipeline build is failing with connection timeouts to debian.beagleboard.org repos

kacf:
Thanks @dellgreen! It would be greatly appreciated if you could also insert some lines about the new variables in [the doc section](https://github.com/mendersoftware/mender-docs/blob/master/04.Artifacts/15.Debian-family/03.variables/docs.md)!

dellgreen:
> Thanks @dellgreen! It would be greatly appreciated if you could also insert some lines about the new variables in [the doc section](https://github.com/mendersoftware/mender-docs/blob/master/04.Artifacts/15.Debian-family/03.variables/docs.md)!

@kacf awesome, see below for documentation pull request

https://github.com/mendersoftware/mender-docs/pull/1078

PR #221, opened 2020-07-14
RFC: mender-convert-modify: Check is selinux is configured in enforce …

nandra:
…mode and force rootfs-relabel

Changelog: Title

CentOS 8 converted image cannot boot to login shell:
user: no shell: permission denied

It turns out that issue was caused by selinux. Idea for relabeling rootfs
was inspired by this post:
https://forums.centos.org/viewtopic.php?t=48714

Removing of /.autorelabel after action is done is not necessary.

### Description

Thank you!

### DIFF
--- a/mender-convert-modify
+++ b/mender-convert-modify
@@ -270,6 +270,19 @@ ${boot_part_device}   ${boot_part_mountpoint}          auto       defaults,sync 
 ${data_part_device}   /data          auto       ${MENDER_DATA_PART_FSTAB_OPTS}      0  0
 EOF"
 
+#
+# Make sure to re-label rootfs when selinux is in enforcing mode
+# e.g. CentOS8 after conversion cannot start login shell due selinux
+# inspired by: https://forums.centos.org/viewtopic.php?t=48714
+#
+if [ -f work/rootfs/etc/selinux/config ]; then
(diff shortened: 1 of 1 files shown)


drewmoseley:
In principle I'm fine with this but I wonder if this is specific to x86 or should this be something more broadly applicable?

nandra:
@drewmoseley you're right. I'll try to install and convert RPI3 image with CentOS and verify that. Thanks.

nandra:
CentoOS 8 for RPI have also syslinux enforce mode enabled so @drewmoseley you're right we need to solve it other way. @mirzak any ideas where to put this handling? Thanks.

mirzak:
I think it can go directly in mender-convert-modify since it is ha check if SELinux is enabled and will only affect system that have it. I assume that checking for `work/rootfs/etc/selinux/config` should be enough? I do not have much experience with SELinux

e.g somewhere at the end, https://github.com/mendersoftware/mender-convert/blob/master/mender-convert-modify#L272

nandra:
@mirzak @drewmoseley updated according proposal. Tested on device after autorelabel + reboot works perfectly fine.

lluiscampos:
Merged. Thanks @nandra!

PR #224, opened 2020-07-20
RFC: Misc changes

drewmoseley:
Some misc changes for review.
The first one I am fairly confident with.
The second one (ie the rpi kernel copy) I'm not so sure is the best fix.  Essentially this is required if we support using pre-converted sdimg files as input.  I have not tested that on other platforms so there may be issues there.

### DIFF
--- a/configs/raspberrypi_config
+++ b/configs/raspberrypi_config
@@ -71,8 +71,12 @@ function platform_modify() {
     RASPBERRYPI_BOOTLOADER_IMAGE="${RASPBERRYPI_KERNEL_IMAGE}"
   fi
 
-  # Extract Linux kernel and install to /boot directory on rootfs
-  run_and_log_cmd "sudo cp work/boot/${RASPBERRYPI_KERNEL_IMAGE} work/rootfs/boot/${MENDER_KERNEL_IMAGETYPE}"
+  # Extract Linux kernel and install to /boot directory on rootfs if it doesn't already exist.
+  # If it already exists, the input image is likely already converted so we just want to use
+  # the existing kernel image.
+  if [ ! -f "work/rootfs/boot/${MENDER_KERNEL_IMAGETYPE}" ]; then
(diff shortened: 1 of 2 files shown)


mender-convert-modify
```diff
@@ -196,7 +196,7 @@ EOF
     # folder, then any subfolders in versioned order.
     for candidate in work/rootfs/boot/dtbs $(find work/rootfs/boot/dtbs/ -maxdepth 1 -type d | sort -V -r); do
       if [ $(find $candidate -maxdepth 1 -name '*.dtb' | wc -l) -gt 0 ]; then
-        run_and_log_cmd "cp -r $candidate work/boot/dtb"
+        run_and_log_cmd "sudo cp -r $candidate work/boot/dtb"
```
kacf, line 199: I'm surprised this is needed, as the boot partition is usually vfat, which doesn't need sudo to copy. But doesn't hurt.

configs/raspberrypi_config
```diff
@@ -71,8 +71,12 @@ function platform_modify() {
...
   fi
 
-  # Extract Linux kernel and install to /boot directory on rootfs
-  run_and_log_cmd "sudo cp work/boot/${RASPBERRYPI_KERNEL_IMAGE} work/rootfs/boot/${MENDER_KERNEL_IMAGETYPE}"
+  # Extract Linux kernel and install to /boot directory on rootfs if it doesn't already exist.
+  # If it already exists, the input image is likely already converted so we just want to use
+  # the existing kernel image.
+  if [ ! -f "work/rootfs/boot/${MENDER_KERNEL_IMAGETYPE}" ]; then
+    run_and_log_cmd "sudo cp work/boot/${RASPBERRYPI_KERNEL_IMAGE} work/rootfs/boot/${MENDER_KERNEL_IMAGETYPE}"
+  fi
```
kacf, line 79: Hmm, re-converting already converted images? I would expect that to fail for a variety of other reasons too. But I think this fix is valid nevertheless. We should not overwrite already existing files unless it's something that needs changing from our side. The copy is primarily to make sure it's present, not to update it.

PR #228, opened 2020-08-27
Changes to allow custom options for filesystems

drewmoseley:

### Description

I added options to allow for the customization of mount options and mkfs option.  These are required to properly support read-only systems with mender-binary-delta updates.

### DIFF
--- a/configs/mender_convert_config
+++ b/configs/mender_convert_config
@@ -32,8 +32,15 @@ MENDER_ARTIFACT_COMPRESSION="gzip"
 # this.
 MENDER_ENABLE_SYSTEMD=y
 
+
+# Custom mkfs options for creating the rootfs partition
+MENDER_ROOT_PART_MKFS_OPTS=""
+
 # Set the fstab options for mounting the data partition
 MENDER_DATA_PART_FSTAB_OPTS="defaults"
+
+# Custom mkfs options for creating the data partition
+MENDER_DATA_PART_MKFS_OPTS=""
 
 # The file system will be grown to occupy the full block device. If the file
 # system is already at maximum size, no action will be performed.
--- a/configs/mender_convert_config
+++ b/configs/mender_convert_config
@@ -36,6 +36,11 @@ MENDER_ENABLE_SYSTEMD=y
 # Custom mkfs options for creating the rootfs partition
 MENDER_ROOT_PART_MKFS_OPTS=""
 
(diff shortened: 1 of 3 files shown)


configs/mender_convert_config
```diff
@@ -32,9 +32,21 @@ MENDER_ARTIFACT_COMPRESSION="gzip"
...
 MENDER_ENABLE_SYSTEMD=y
 
+
+# Custom mkfs options for creating the rootfs partition
+MENDER_ROOT_PART_MKFS_OPTS=""
+
+# Set the fstab options for mounting the root partition
+MENDER_ROOT_PART_FSTAB_OPTS="defaults"
+# Set to 0 to disable fsck; needed for read-only rootfs support
+MENDER_ROOT_PART_FS_PASSNO="1"
```
mirzak, line 42: This can be one configuration option, e.g:

```
MENDER_ROOT_PART_FSTAB_OPTS = "defaults 1 1"
```

mirzak (changes requested):

modules/disk.sh
```diff
@@ -114,21 +114,27 @@ disk_create_file_system_from_folder() {
...
+  esac
+
   case "$4" in
     "ext4")
       MKFS_EXT4="/usr/bin/mkfs.ext4"
       if [ ! -f ${MKFS_EXT4} ]; then
         MKFS_EXT4="/sbin/mkfs.ext4"
       fi
-      run_and_log_cmd "${MKFS_EXT4} -q -F ${2}"
+      run_and_log_cmd "${MKFS_EXT4} -q -F ${2} -i 4096 ${EXTRA_OPTS}"
```
kacf, line 129: The `-i 4096` was snuck into the same commit. Why? If it's intentional I think this should be in a separate commit.

configs/mender_convert_config
kacf: Yes, but in Yocto the same variable is used only for the first part, so it's probably better to keep them split for consistency. The dump and pass number aren't configurable in Yocto ATM.

kacf (changes requested):

modules/disk.sh
drewmoseley: Rebase failure.  It's fixed now

kacf (approved):
Thanks, looks good!

@mirzak?

PR #229, opened 2020-09-08
Update versions for 2.2.0 release

lluiscampos:
Changelog: Update RaspiOS image to 2020-08-24
Changelog: Update mender-artifact to 3.4.0


kacf (approved):
> Changelog: Update mender client to 2.4.0-build1

I would not include this in the changelog, only the final version should be listed there. Everything else looks good, so pre-approving.

PR #231, opened 2020-09-09
MEN-3912: Warn user when converting read-only fs breaking Delta updates

lluiscampos:
Changelog: Warn user when converting read-only file systems that would
result in unstable checksums, making the image incompatible with Mender
Delta updates.

### DIFF
--- a/configs/mender_convert_config
+++ b/configs/mender_convert_config
@@ -38,7 +38,8 @@ MENDER_ROOT_PART_MKFS_OPTS=""
 
 # Set the fstab options for mounting the root partition
 MENDER_ROOT_PART_FSTAB_OPTS="defaults"
-# Set to 0 to disable fsck; needed for read-only rootfs support
+
+# Set to 0 to disable fsck; recommended for read-only rootfs
 MENDER_ROOT_PART_FS_PASSNO="1"
 
 # Set the fstab options for mounting the data partition
--- a/mender-convert-modify
+++ b/mender-convert-modify
@@ -261,6 +261,14 @@ if [ "${MENDER_DATA_PART_GROWFS}" == "y" ]; then
   MENDER_DATA_PART_FSTAB_OPTS="${MENDER_DATA_PART_FSTAB_OPTS},x-systemd.growfs"
 fi
 
+if echo ${MENDER_ROOT_PART_FSTAB_OPTS} | tr ',' '\n' | egrep -q "^ro$"; then
+  if ! echo "${MENDER_ROOT_PART_MKFS_OPTS}" | fgrep -q -- "-O ^64bit"; then
+    log_warn "EXT4 64bits feature is known to create unstable checksums on read-only file systems, add \"-O ^64bit\" to MENDER_ROOT_PART_MKFS_OPTS to remove it"
+  fi
+  if ! echo "${MENDER_ROOT_PART_MKFS_OPTS}" | fgrep -q -- "-O ^has_journal"; then
+    log_warn "EXT4 journal feature is known to create unstable checksums on read-only file systems, add \"-O ^has_journal\" to MENDER_ROOT_PART_MKFS_OPTS to remove it"
+  fi
+fi
 
 log_info "Using boot partition device in fstab: $boot_part_device"
 log_info "Using data partition device in fstab: $data_part_device"


configs/mender_convert_config
```diff
@@ -32,6 +32,8 @@ MENDER_ARTIFACT_COMPRESSION="gzip"
 # this.
 MENDER_ENABLE_SYSTEMD=y
 
+# Set to 'y' for read-only rootfs
+READ_ONLY_ROOTFS=n
```
mirzak, line 36: Do we need a new variable here?

If you look at https://github.com/mendersoftware/mender-convert/blob/master/mender-convert-modify#L268-L275, we set the `/etc/fstab` for `/dev/root`.

So if `MENDER_ROOT_PART_FSTAB_OPTS` contains `ro` we could trigger these warnings. Unless it is to big of a hassle to parse in stable way.

lluiscampos: Makes sense to me, that is a good idea. I was following the suggested fix in the JIRA issue.

@kacf were you thinking otherwise or just an overlook?

kacf: Hmm, I had not considered that, but I *think* it is a good idea. We are not expecting the read-only property of the system to be defined anywhere else than in `/etc/fstab`, are we? Like setting kernel arguments to `ro` is not enough, the user also needs `fstab`, right?

@drewmoseley, you reported the ticket, does it sound good to you as well?

drewmoseley: As far as I know that is correct.  I think the kernel command line "ro" may get overridden by whatever is in fstab so I think relying on fstab is the way to go.

lluiscampos: Cool! Good discussion. I'll rework soon

drewmoseley (approved):

mender-convert-modify
```diff
@@ -261,6 +261,17 @@ if [ "${MENDER_DATA_PART_GROWFS}" == "y" ]; then
...
 
+if echo ${MENDER_ROOT_PART_FSTAB_OPTS} | tr ',' '\n' | egrep -q "^ro$"; then
+  if ! echo "${MENDER_ROOT_PART_MKFS_OPTS}" | fgrep -q -- "-O ^64bit"; then
+    log_warn "EXT4 64bits feature is known to create unstable checksums on read-only file systems, add \"-O ^64bit\" to MENDER_ROOT_PART_MKFS_OPTS to remove it"
+  fi
+  if ! echo "${MENDER_ROOT_PART_MKFS_OPTS}" | fgrep -q -- "-O ^has_journal"; then
+    log_warn "EXT4 journal feature is known to create unstable checksums on read-only file systems, add \"-O ^has_journal\" to MENDER_ROOT_PART_MKFS_OPTS to remove it"
+  fi
+  if [ "${MENDER_ROOT_PART_FS_PASSNO}" != "0" ]; then
+    log_warn "fsck enabled on a read-only file system, set MENDER_ROOT_PART_FS_PASSNO to 0 to disable it "
```
kacf, line 272: No strong opinion, but I think I would vote for `log_info` level here. I don't consider it wrong to use fsck on a read-only filesystem. This is still useful in cases where hardware storage errors have occurred, or when the filesystem has become read-write mounted because of some manual action.

mender-convert-modify
```diff
@@ -261,6 +261,17 @@ if [ "${MENDER_DATA_PART_GROWFS}" == "y" ]; then
   MENDER_DATA_PART_FSTAB_OPTS="${MENDER_DATA_PART_FSTAB_OPTS},x-systemd.growfs"
 fi
 
+if echo ${MENDER_ROOT_PART_FSTAB_OPTS} | tr ',' '\n' | egrep -q "^ro$"; then
```
kacf, line 264: Nice `tr` trick!

kacf (approved):

configs/mender_convert_config
```diff
@@ -32,7 +32,6 @@ MENDER_ARTIFACT_COMPRESSION="gzip"
 # this.
 MENDER_ENABLE_SYSTEMD=y
 
-
```
mirzak, line 35: Nit, unrelated change

mirzak (approved):

mender-convert-modify
lluiscampos: Then I will remove it (I don't think is a point to "log" a recommendation) and rephrase [comment here](https://github.com/mendersoftware/mender-convert/blob/master/configs/mender_convert_config#L41).

PR #235, opened 2020-09-16
[test-utils.sh] Clean the pip list + grep command

lluiscampos:
So that we clean the deprecation warning and the multiple errors when
calling this command from CI.

### DIFF
--- a/scripts/test/test-utils.sh
+++ b/scripts/test/test-utils.sh
@@ -92,7 +92,7 @@ run_tests() {
 
   converted_image_name="${converted_image_file%.img}"
 
-  if pip3 list | grep -q -e pytest-html; then
+  if pip3 list --format=columns | grep pytest-html; then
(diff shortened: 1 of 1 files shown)


lluiscampos:
See the current warning/erros in latest run from master [here](https://gitlab.com/Northern.tech/Mender/mender-convert/-/jobs/736357295).

oleorhagen (approved):

lluiscampos:
The warning disappeared but the [errors are still there](https://gitlab.com/Northern.tech/Mender/mender-convert/-/jobs/741013715). @oleorhagen do you have any suggestion?

oleorhagen:
The issue is that `grep` exits immediately upon finding a match when run with `-q`, I think :)

lluiscampos:
> The issue is that `grep` exits immediately upon finding a match when run with `-q`, I think :)

Thanks. Trying that :)

PR #236, opened 2020-09-21
master

nishad1092:
So I  just have one small change here in this file 
https://github.com/mendersoftware/mender/blob/master/support/mender-device-identity

So, It will be great to have this edit before  building. 

the change I need is in line 60 of the mender-device-identity file, replace that line/ comment that line and add a new line which is :-----  
echo "hostname=$(cat /etc/hostname)"

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,4 @@
 deploy
 input
 work
+rootfs_overlay*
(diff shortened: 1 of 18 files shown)


drewmoseley:
@nishad1092 this doesn't seem to have any changes from you.  All the commits here are from our existing branches.  Can you clean up and rebase on the current master branch?

lluiscampos:
This PR is attempting to merge `2.2.x` into `master`, which is not possible. Closing.

Feel free to open a new one with actual code changes to be reviewed :)

PR #237, opened 2020-09-21
MEN-3952: beaglebone: Implement workaround for broken U-Boot and kernel.

kacf:
Several smaller fixes, but this is the main one:

```
commit 8c629c0612f43b36e2c53848b2b5a6e834f5968c
Date:   Mon Sep 21 19:54:03 2020

    MEN-3952: beaglebone: Implement workaround for broken U-Boot and kernel.
    
    Originally the intention was to fix the U-Boot integration for
    Beaglebone, but this seems to have been broken for a long time, almost
    a year, and no one has complained about it. The problem appears to be
    connected to this commit [1], but it's not entirely clear to me what
    the fix should be, and there are a lot of patches for our U-Boot fork.
    
    However, since the U-Boot integration is supposed to be the backup
    solution, rather than dig into U-Boot I decided to just fix the UEFI
    path instead. This method doesn't require any patching, but there is
    a problem with the specific image which is available for download at
    the time of writing. So just grab the boot loader and kernel pieces
    from a later, yet unbuilt, image, and it all works again.
    
    Changelog: Title
    
    Changelog: beaglebone: Remove U-Boot integration, which has not worked
    for a long time. U-Boot will still be used for booting, but GRUB will
    be used for integration with Mender, by chainloading via UEFI.
    
    [1] https://github.com/mendersoftware/uboot-mender/commit/e88d5dbf01dc8c3822c695267b43d6000d9b2a19
    
```

### DIFF
--- a/configs/beaglebone_black_base_config
+++ b/configs/beaglebone_black_base_config
@@ -1,24 +1,64 @@
-# Binaries generated with the following script:
-#    https://github.com/mendersoftware/mender-convert-integration-scripts/blob/master/build-uboot-bbb.sh
-
-# There are reported issues with GRUB bootloader integration, fallback to U-boot.
-MENDER_GRUB_EFI_INTEGRATION=n
-
-# We will write a modified bootloader
-MENDER_COPY_BOOT_GAP=n
-
-# 4MB alignment
-MENDER_PARTITION_ALIGNMENT="4194304"
-
-BEAGLEBONE_BLACK_BINARIES="${BEAGLEBONE_BLACK_CONFIG}-2018.07.tar.gz"
+# At the time of writing, the binaries below were uploaded manually. However,
+# they are simply a vanilla U-Boot build of v2020.07, from Ubuntu 18.04, using
+# these commands:
+#   make CROSS_COMPILE=arm-linux-gnueabihf- am335x_boneblack_vboot_defconfig
+#   make CROSS_COMPILE=arm-linux-gnueabihf-
+#   make CROSS_COMPILE=arm-linux-gnueabihf- envtools
+BEAGLEBONE_BLACK_BINARIES="beaglebone_black_uefi-2020.07.tar.gz"
 BEAGLEBONE_BLACK_BINARIES_URL="${MENDER_STORAGE_URL}/mender-convert/uboot/beaglebone/${BEAGLEBONE_BLACK_BINARIES}"
 
-function platform_modify() {
+# At the time of writing, the binaries below were uploaded manually. They are
+# simply a dump of the deb package from the link below, with the addition of the
+# initrd file, which is not part of the package, but which is generated when
+# installed on a live Beaglebone device.
+# Original deb URL: http://repos.rcn-ee.com/debian/pool/main/l/linux-upstream/linux-image-4.19.94-ti-r51_1buster_armhf.deb
+BEAGLEBONE_EFI_COMPATIBLE_KERNEL="linux-image-4.19.94-ti-r51_1buster_armhf-deb-dump.tar.xz"
+BEAGLEBONE_EFI_COMPATIBLE_KERNEL_URL="${MENDER_STORAGE_URL}/mender-convert/beaglebone/kernel/${BEAGLEBONE_EFI_COMPATIBLE_KERNEL}"
+BEAGLEBONE_EFI_COMPATIBLE_KERNEL_VERSION=4.19.94-ti-r51
+
+MENDER_IGNORE_UBOOT_BROKEN_UEFI=1
+MENDER_IGNORE_MISSING_EFI_STUB=1
+
+function beaglebone_fix_broken_uefi_uboot() {
+  log_warn "Attempting to work around broken UEFI support in U-Boot by forcing a newer U-Boot version."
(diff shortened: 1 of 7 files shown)


kacf:
@mirzak, @drewmoseley: What do you think about this PR? Essentially both boot methods are broken, so I opted to fix UEFI, and simply delete the U-Boot integration for now. Primarily this is motivated by the fact that no one has noticed it was broken in almost a whole year, so Beaglebone is simply not a popular board. My thinking now is that it serves primarily as a vehicle for us to verify that UEFI works for at least one non-QEMU board, and we don't need to worry about compatibility. But I'd like your thoughts on this.

lluiscampos (approved):
This is a great solution, @kacf I am very happy we are back to the GRUB/EFI integration for BBB. I think it is very important to have it as part of the default conversions and tests in `mender-convert`.

I couldn't spot anything wrong (rather than the tests fail due to config files being renamed), so this is a go from me :)

mender-convert-modify
```diff
@@ -196,18 +196,18 @@ EOF
...
-    # folder, then any subfolders in versioned order.
-    for candidate in work/rootfs/boot/dtbs $(find work/rootfs/boot/dtbs/ -maxdepth 1 -type d | sort -V -r); do
-      if [ $(find $candidate -maxdepth 1 -name '*.dtb' | wc -l) -gt 0 ]; then
-        run_and_log_cmd "sudo cp -r $candidate work/boot/dtb"
-        break
-      fi
-    done
-  fi
+# Copy dtb directory to the boot partition for use by the bootloader.
+if [ -d work/rootfs/boot/dtbs ]; then
```
drewmoseley, line 202: I saw you did something similar to this in meta-mender and was going to ask about it.  Doesn't this present a risk of them being out of date after full rootfs updates are processed?  Do we want to setup a state script to update them?

configs/beaglebone_black_base_config
```diff
@@ -1,39 +1,70 @@
-# Binaries generated with the following script:
-#    https://github.com/mendersoftware/mender-convert-integration-scripts/blob/master/build-uboot-bbb.sh
-
-# There are reported issues with GRUB bootloader integration, fallback to U-boot.
-MENDER_GRUB_EFI_INTEGRATION=n
+# At the time of writing, the binaries below were uploaded manually. However,
+# they are simply a vanilla U-Boot build of v2020.07, from Ubuntu 18.04, using
+# these commands:
```
drewmoseley, line 3: Should we update https://github.com/mendersoftware/uboot-mender along with this?

drewmoseley (approved):
Just a few questions that shouldn't hold up merging of this.  Looks good to me.

mirzak (approved):

mender-convert-modify
kacf: It does. But this is how the UEFI support works in U-Boot; it loads the dtb files from the UEFI partition. How the dtb files are loaded falls under the whole `distro_bootcmd` framework, and therefore it will be hard to "menderize" it by loading them from a different partition. I guess it makes sense from their perspective, you don't want a UEFI loader to look *outside* of the UEFI partition. Quite a tricky issue though....

Using a state script is a good idea, but it comes with the same caveats as updating the bootloader: There is no rollback if it breaks. So I don't think we should enable it by default, since I reckon having an old dtb file is better than having a broken one.

Ideally we should think about some way of solving this, but I think it's ok to leave it as it is for now. It's not worse than it already was, and if you're integrating with U-Boot, the DTB files will end up in both places, so you can still make DTB files upgradeable by making sure your U-Boot commands are picking them from the rootfs partition.

configs/beaglebone_black_base_config
kacf: What do you mean? To upload the binaries? I don't think it's worth the effort. This is a one-shot upload which only applies to this specific image which is available today. Everything we need is already upstream, it's just not in the image yet.

drewmoseley: Do we need the ability to rebuild this reliably?

kacf: No, this is a one time thing, we want to use the U-Boot image which is already in the image. The only reason it is done for this specific image is because their image is old and out of date.

kacf:
@drewmoseley: Currently the tests are failing because the dual rootfs configuration will not fit on the EMMC, not by a long shot. You set it to 3648MiB [some time ago](https://github.com/mendersoftware/mender-convert/commit/5e04bbc3334f0e2bc822873d149b2841b064645a). Do you remember which image you tested? I cannot find an image on [the Beaglebone download page](https://beagleboard.org/latest-images) which will allow splitting into the dual rootfs configuration and still fit. I could be that with the current Debian we have to insist on using sdcard.

drewmoseley:
I don't recall but I do have _bone-debian-10.1-console-armhf-2019-09-15-1gb.img_ in my input directory so that's likely the one.

kacf:
Thanks, found it! This image appears to be less official, it's not on their homepage. But I found it, and later ones, at https://elinux.org/Beagleboard:BeagleBoneBlack_Debian. I will adjust the download link in the test.

PR #239, opened 2020-10-06
2.2.x cherry-picks: Remove BBB U-Boot support and replace with UEFI

kacf:
This is a pretty severe change to bring into a patch release, but since the U-Boot integration is completely broken, I still think we should do it, since there is no other working solution. And it also makes our own testing easier,


lluiscampos (approved):
I agree on bringing these to `2.2.x`. The commit list looks good.

kacf:
Thanks!

PR #240, opened 2020-10-19
[tests, CI, dependabot] Remove --no-pull, update tests, add dependabot

lluiscampos:
Remove all the logic for the special handling of this submodule, and
treat it instead a regular Git submodule.

This commit updates the submodule to latest, removes the --no-pull
functionality, and removes the ignore of the submodule to make
developers aware

To ease maintenance, add also dependabot to pay attention to the future
updates of the submodule.

### DIFF
--- a/scripts/test/test-utils.sh
+++ b/scripts/test/test-utils.sh
@@ -136,6 +136,7 @@ run_tests() {
 get_pytest_files() {
   wget -N ${MENDER_ACCEPTANCE_URL}/pytest.ini -P $WORKSPACE/mender-image-tests
   wget -N ${MENDER_ACCEPTANCE_URL}/common.py -P $WORKSPACE/mender-image-tests
+  wget -N ${MENDER_ACCEPTANCE_URL}/helpers.py -P $WORKSPACE/mender-image-tests
   wget -N ${MENDER_ACCEPTANCE_URL}/conftest.py -P $WORKSPACE/mender-image-tests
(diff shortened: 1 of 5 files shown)


lluiscampos:
See related PRs https://github.com/mendersoftware/mender-qa/pull/432 and https://github.com/mendersoftware/meta-mender/pull/1137

scripts/test/test-utils.sh
```diff
@@ -136,6 +136,7 @@ run_tests() {
 get_pytest_files() {
   wget -N ${MENDER_ACCEPTANCE_URL}/pytest.ini -P $WORKSPACE/mender-image-tests
   wget -N ${MENDER_ACCEPTANCE_URL}/common.py -P $WORKSPACE/mender-image-tests
+  wget -N ${MENDER_ACCEPTANCE_URL}/helpers.py -P $WORKSPACE/mender-image-tests
```
mzedel, line 139: from the commit/ PR description I'm not entirely sure, but I assume this is the update to the tests, right?

mzedel (approved):

scripts/test/test-utils.sh
lluiscampos: The latest master of the submodule does `from helpers import ...`, which will fail.

This should be failing in master right now the same way, but no builds have been done in the last days.

kacf (approved):

PR #241, opened 2020-10-20
Do not compress output image for uncompressed input image

lluiscampos:
The two main commits are:

* Support overriding default image compression
The default behavior is not changing, but splitting the arguments in
prepend/append allows the user to override the output image compression.

* Do not compress output image for uncompressed input image
Respecting this way always the input type (compressed or not) by
default, allowing the user to override behaviour if desired.
For the tests, removed the unnecessary duplication on rerunning all
tests on uncompressed versions, and added instead an explicit test only
for qemu-x86_64.

### DIFF
--- a/mender-convert
+++ b/mender-convert
@@ -24,7 +24,8 @@ touch work/convert.log
 ###############################################################################
 #                Declaration of important variables for this file             #
 ###############################################################################
-declare override_extraargs="" # Override arguments passed to the sub-scripts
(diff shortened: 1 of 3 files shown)


oleorhagen (approved):
Really really nice!

Elegant solution. I was a bit curious how you we're gonna solve this without hacking too much. This is way nicer than what I had in mind!

:medal_sports:  for you

PR #242, opened 2020-10-21
MEN-2208: Run test_update tests for meta-mender QEMU image

lluiscampos:
The main commit is:

MEN-2208: Run test_update tests for meta-mender QEMU image
This commit adds all necessary bits to be able to execute runtime tests
for mender-convert'ed QEMU x86_64 image. Namely:
* Install SSH key as an authorized one on the device using an overlay,
and pass the private one to be use by the test framework.
* Create a QEMU launcher wrapper, similar to the one in meta-mender, for
the test framework to use in order to start/stop the virtual device.
* Exclude the test_update tests for all configurations but Ubuntu QEMU
one using a pytest filter 'not test_upate'.
* Copy from meta-mender the keys required for test_signed_updates.
* Install mender-artifact in CI to be used in tests.


lluiscampos:
Probably some adjustments will still be required to make CI happy, but I don't foresee design changes so the PR can be reviewed.

scripts/test/test-utils.sh
```diff
@@ -116,16 +128,19 @@ run_tests() {
   echo "Converted image name: "
   echo "$converted_image_name"
 
-  python3 -m pytest --verbose \
+  eval python3 -m pytest \
```
lluiscampos, line 131: I added the `eval` here to be able to properly quote the filter, so that `pytest_extra_args` resolves to `-k 'not test_update'`.

However this doesn't seem to work on CI and I am running out of ideas, I need help from a shell ninja. It works locally on my machine, and CI is using bash, so I don't know where is the issue really.

mender-convert-modify
```diff
@@ -143,7 +143,9 @@ initrd_imagetype=initrd
 EOF
 
   run_and_log_cmd "ln -s ${kernel_imagetype} work/rootfs/boot/kernel"
-  run_and_log_cmd "ln -s ${initrd_imagetype} work/rootfs/boot/initrd"
+  if [ "${initrd_imagetype}" != "" ]; then
+    run_and_log_cmd "ln -s ${initrd_imagetype} work/rootfs/boot/initrd"
+  fi
```
lluiscampos, line 148: @mirzak This I had to add once I upgraded the Ubuntu image. The only changes in the image is adding packages `openssh-server` and `dhcpcd5`, so I don't know how could that have changed the kernel to not use an initrd now.

Do you think this is an incorrect fix?

(the qemu image boots, the tests actually pass)

lluiscampos: I moved the related commits to #243. Let's discuss there how should this be fixed.

scripts/test/test-utils.sh
lluiscampos: Ignore my previous comment. It actually works but I got confused with a previous run (not my last force push).

scripts/test/mender-convert-qemu
```diff
@@ -0,0 +1,41 @@
...
+set -e
+
+if [ -z "$DISK_IMG" ]; then
+    echo "ERROR: DISK_IMG is not set"
+    exit 1
+fi
+
+# Try common locations for OVMF
+ovmf_file=""
+for maybe_ovmf_file in /usr/share/OVMF/OVMF_CODE.fd /usr/share/ovmf/x64/OVMF_CODE.fd; do
```
merlin-northern, line 12: minor: `find /usr -type f -and -name OVMF_CODE.fd | head -1` in case we need to really look for it.

scripts/test/mender-convert-qemu
```diff
@@ -0,0 +1,41 @@
...
+fi
+
+qemu-system-x86_64 \
+    -enable-kvm \
+    -nographic \
+    -m 256 \
+    -net user,hostfwd=tcp::8822-:22 \
+    -net nic,macaddr=52:54:00$(od -txC -An -N3 /dev/urandom|tr \  :) \
+    -bios ${ovmf_file} \
+    -drive format=raw,file=${DISK_IMG} &
```
merlin-northern, line 30: do we need some output redirection here? (as in: do we need to preserve something qemu will output to the stdout?)

scripts/test/mender-convert-qemu
```diff
@@ -0,0 +1,41 @@
...
+
+qemu_pid=$!
+
+trap qemu_cleanup HUP KILL TERM
+
+qemu_cleanup() {
+    kill -s TERM $qemu_pid
```
merlin-northern, line 37: technically it is possible for qemu to hang or otherwise not respond to TERM, I would sleep here for 2 or 4 seconds and then `kill -0 $qemu_pid 2>/dev/null && kill -9 $qemu_pid; sleep 2; kill -0 $qemu_pid 2>/dev/null && echo "qemu pid $qemu_pid still present";`

scripts/test/mender-convert-qemu
```diff
@@ -0,0 +1,41 @@
...
+    -nographic \
+    -m 256 \
+    -net user,hostfwd=tcp::8822-:22 \
+    -net nic,macaddr=52:54:00$(od -txC -An -N3 /dev/urandom|tr \  :) \
+    -bios ${ovmf_file} \
+    -drive format=raw,file=${DISK_IMG} &
+
+qemu_pid=$!
+
+trap qemu_cleanup HUP KILL TERM
```
merlin-northern, line 34: should we also add INT and CHLD here?

scripts/test/run-tests.sh
```diff
@@ -83,7 +92,10 @@ else
     #                 "release-1" \
     #                 "${TINKER_IMAGE_URL}" \
     #                 "${TINKER_IMAGE}.img" \
-    #                 "${TINKER_IMAGE}.zip" || test_result=$?
+    #                 "${TINKER_IMAGE}.zip" \
+    #                 -- \
+    #                 "-k" "'not test_update'" \
+    #                 || test_result=$?
```
merlin-northern, line 98: is there a reason we are keeping those lines commented-out?

scripts/test/test-utils.sh
```diff
@@ -154,5 +156,18 @@ get_pytest_files() {
...
+  wget -N ${MENDER_ACCEPTANCE_URL}/files/test-public-RSA.pem -P $WORKSPACE/mender-image-tests/files
 }
 
+prepare_ssh_keys() {
+  if [ "$(stat -c %U tests/ssh-public-key-overlay/root)" != "root" ]; then
+    sudo chown -R root:root tests/ssh-public-key-overlay/root
+  fi
+  if [ "$(stat -c %a tests/ssh-priv-key/key)" != "600" ]; then
+    chmod 600 tests/ssh-priv-key/key
+  fi
```
merlin-northern, line 171: suggestion: nice ifs, but I would just drop them and run chown and chmod. 
doesn't chmod need sudo as well?

scripts/test/test-utils.sh
```diff
@@ -154,5 +156,18 @@ get_pytest_files() {
   wget -N ${MENDER_ACCEPTANCE_URL}/helpers.py -P $WORKSPACE/mender-image-tests
   wget -N ${MENDER_ACCEPTANCE_URL}/conftest.py -P $WORKSPACE/mender-image-tests
   wget -N ${MENDER_ACCEPTANCE_URL}/fixtures.py -P $WORKSPACE/mender-image-tests
+  mkdir -p $WORKSPACE/mender-image-tests/files
+  wget -N ${MENDER_ACCEPTANCE_URL}/files/test-private-EC.pem -P $WORKSPACE/mender-image-tests/files
+  wget -N ${MENDER_ACCEPTANCE_URL}/files/test-private-RSA.pem -P $WORKSPACE/mender-image-tests/files
+  wget -N ${MENDER_ACCEPTANCE_URL}/files/test-public-EC.pem -P $WORKSPACE/mender-image-tests/files
+  wget -N ${MENDER_ACCEPTANCE_URL}/files/test-public-RSA.pem -P $WORKSPACE/mender-image-tests/files
```
merlin-northern, line 162: suggested refactoring:
```bash
for a in EC RSA; do
  for t in public private; do
    wget -N ${MENDER_ACCEPTANCE_URL}/files/test-${t}-${a}.pem -P $WORKSPACE/mender-image-tests/files;
  done;
done;
```

scripts/test/mender-convert-qemu
```diff
@@ -0,0 +1,41 @@
...
+if [ -z "${ovmf_file}" ]; then
+    echo "ERROR: could not find OVMF file"
+    exit 1
+fi
+
+qemu-system-x86_64 \
+    -enable-kvm \
+    -nographic \
+    -m 256 \
+    -net user,hostfwd=tcp::8822-:22 \
```
merlin-northern, line 27: minor: just for clarity, maybe lets create `MENDER_SSH_PORT=8822` variable?

scripts/test/run-tests.sh
```diff
@@ -38,14 +38,18 @@ mkdir -p ${WORKSPACE}
...
   usage
 fi
 
 test_result=0
 
 if [ "$1" == "--prebuilt-image" ]; then
-  run_tests "$2" "$3" || test_result=$?
+  run_tests "$2" "$3" \
+            "-k" "'not test_update'" \
+            || test_result=$?
```
merlin-northern, line 52: minor: do we really need to split this line? it is less than 70 characters long.

tests/ssh-priv-key/key
```diff
@@ -0,0 +1,51 @@
+-----BEGIN RSA PRIVATE KEY-----
```
merlin-northern, line 1: these keys are generated for the purpose of this PR?

merlin-northern (commented):
> Copy from meta-mender the keys required for test_signed_updates

Which keys are those?

scripts/test/mender-convert-qemu
lluiscampos: Ok, thanks!

scripts/test/mender-convert-qemu
lluiscampos: Yes, we want all the output in the terminal. This is important when tests fail to be able to inspect the GRUB boot output.

scripts/test/mender-convert-qemu
lluiscampos: This only happens once (at the end of all the tests) and if it does not respond the test will finish anyway and the VM be teared down, right? So I don't think it is that important, and I'd rather keep it simple.

scripts/test/mender-convert-qemu
lluiscampos: I add these to (almost) follow the [original script in meta-mender](https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-qemu/scripts/mender-qemu#L20). And actually, we know that is only TERM signal being sent (test code [here](https://github.com/mendersoftware/meta-mender/pull/1145/files#diff-a0376fd2164bc856f75068a78c3eb3e26a5232e47bdaa1417da392b538cb540fL175)), so is it really important to add more?

scripts/test/test-utils.sh
lluiscampos: When running locally, it was very annoying to get asked for sudo password often, that is why I added the ifs. I think I will leave it.

That specific chmod is done on the private key, which is owned by whoever cloned the repo, therefore does not require sudo.

NOTE: This was actually the problem for the tests to fail (incorrect permissions), the last force-push added more `chmod`s here that made the tests finally pass.

scripts/test/test-utils.sh
lluiscampos: Nice!

I originally wanted to refactor and move these keys into the `mender-image-tests` repository, but run into trouble with meta-mender and decided to just copy. I kind of "liked" that it was explicit as to point out to someone in the future that this is very ugly to do and we should refactor instead...

Anyway, I'll do your suggested rework.

scripts/test/mender-convert-qemu
lluiscampos: This is hardcoded in the test framework and not possible to change that easily. The meta-mender script does it [the same way](https://github.com/mendersoftware/meta-mender/blob/master/meta-mender-qemu/scripts/mender-qemu#L128).

scripts/test/run-tests.sh
lluiscampos: Yeah, but I did it to be consistent with all the others below (the ones calling `convert_and_test`).

tests/ssh-priv-key/key
lluiscampos: I copied them from [mender-test-containers](https://github.com/mendersoftware/mender-test-containers/tree/master/docker/ssh-keys).

lluiscampos:
> > Copy from meta-mender the keys required for test_signed_updates
> 
> Which keys are those?

The ones we are discussing here: https://github.com/mendersoftware/mender-convert/pull/242#discussion_r510687679

scripts/test/test-utils.sh
lluiscampos: Thinking about it again, sorry but your suggestion is technically longer (5 lines) and adds complexity. So I'll leave it plain and simple for the reason stated above (that we know this full `get_pytest_files` idea is a horrible one).

scripts/test/mender-convert-qemu
merlin-northern: if we are sure they never get sent it is all right. for the same reasons no harm in adding them.

merlin-northern:
ok, looks good.

merlin-northern (approved):

lluiscampos:
Last force-push updates submdoule to latest master (after mendersoftware/mender-image-tests#17).

PR #243, opened 2020-10-22
[mender-convert-modify] Allow images with no initrd

lluiscampos:
Extracting these two commits from #242 to merge them independently, as tests as currently failing for master.


mirzak (approved):

tranchitella (approved):
The change in the bash script is syntactically correct, this is all I can check.

PR #244, opened 2020-10-23
QA-214: Package 'latest' Mender client debian package by default

oleorhagen:
This cannot be merged until https://github.com/mendersoftware/mender-dist-packages/pull/34 is in, and a latest package is merged obviously.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -5,7 +5,7 @@ variables:
   # These variables are present elsewhere in the repository too. Make sure to
   # search for and change them too.
   MENDER_ARTIFACT_VERSION: master
-  MENDER_CLIENT_VERSION: master
+  MENDER_CLIENT_VERSION: latest
   # Make sure to update the link in mender-docs to the new one when changing
   # this.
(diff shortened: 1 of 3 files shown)


lluiscampos (changes requested):
I don't think this will work, as the pipeline will override with master (see [here](https://github.com/mendersoftware/mender-convert/blob/master/.gitlab-ci.yml#L8)).

But also, I am not sure that we want this to be mender latest [tag] instead of master for mender-convert/master. We need to discuss this once more.

lluiscampos:
> I don't think this will work, as the pipeline will override with master (see [here](https://github.com/mendersoftware/mender-convert/blob/master/.gitlab-ci.yml#L8)).
> 
> But also, I am not sure that we want this to be mender latest [tag] instead of master for mender-convert/master. We need to discuss this once more.

As discussed in the call, this is correct but we also have to change the pipeline to use `latest`.

lluiscampos (approved):

oleorhagen:
2.2.x-cherry: https://github.com/mendersoftware/mender-convert/pull/253

PR #248, opened 2020-10-30
[pipeline] Switch from dependencies to needs

lluiscampos:
This new keyword of GitLab allows for jobs in further stages start as
soon as their dependencies are done.

This should speed-up this pipeline, starting most of the acceptance
testing without finishing to the raspberrypi conversions.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -30,6 +30,7 @@ stages:
 
 build:
   stage: build
+  needs: []
   image: docker:git
   services:
     - docker:19.03.5-dind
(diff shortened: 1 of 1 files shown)


alfrunes (approved):

oleorhagen (approved):
Nice!

Let's exploit the DAG where we can

.gitlab-ci.yml
```diff
@@ -136,17 +139,21 @@ convert_raspbian_raspberrypi4:
 
 test_acceptance_prebuilt_raspberrypi3:
   <<: *test_acceptance_prebuilt_raspberrypi
-  dependencies:
-    - build
-    - convert_raspbian_raspberrypi3
+  needs:
+    - job: build
```
oleorhagen, line 143: Question:
I would have guessed that these relations are transitive, so that, since the convert job already `needs` build, is this needed here then?

kacf (approved):

.gitlab-ci.yml
lluiscampos: I see. I think you are right, but I don't dare to try. I will merge as-is and learn more about this feature in my playground repo next value add day. Thanks!

PR #249, opened 2020-10-30
[tests/mender-image-tests] Set to previous version

lluiscampos:
After reverting the fixture rename in meta-mender/master.


lluiscampos:
This is a follow up of https://github.com/mendersoftware/meta-mender/pull/1160.

PR #251, opened 2020-11-02
Remove the 'only' tag from the publish job

oleorhagen:
It is already manual, so really no need for the 'only' tag.

Changelog: None

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -210,6 +210,3 @@ publish:s3:
     -   aws s3api put-object-acl --acl public-read --bucket $S3_BUCKET_NAME
           --key ${RASPBIAN_NAME}/arm/${PUBLISH_NAME}
     - done
-
-  only:
-    - /^(master|[0-9]+\.[0-9]+\.x)$/


kacf (approved):

lluiscampos:
I disagree, I don't think this is a good idea.

As the deliverable here is identified only by the mender client version, with this change we basically allow for any dev branch to override production deliverables.

@oleorhagen Could you elaborate in why did you need this with the ongoing release?

oleorhagen:
> I disagree, I don't think this is a good idea.
> 
> As the deliverable here is identified only by the mender client version, with this change we basically allow for any dev branch to override production deliverables.
> 
> @oleorhagen Could you elaborate in why did you need this with the ongoing release?

Because I wanted to run this with the release candidate: `<release>-build'n'`.
Another option could be to add this to the only tag.

Are we completely off here?

lluiscampos:
@oleorhagen I see. But `MENDER_VERSION` can still be `<release>-build'n'` (in the GitLab variables and/or in the code) while `mender-convert` is `2.2.x`, hence still being published from a verified and reviewed branch. This is how I have done it previously.

lluiscampos:
@oleorhagen Alternatively, I would have extended the current only regex to add the `-buildXX` suffix. In any case what I want to avoid is publishing from arbitrary branches.

oleorhagen:
> @oleorhagen I see. But `MENDER_VERSION` can still be `<release>-build'n'` (in the GitLab variables and/or in the code) while `mender-convert` is `2.2.x`, hence still being published from a verified and reviewed branch. This is how I have done it previously.

Hmm, I see. But how did the release-job show up then? It didn't with me.


> @oleorhagen Alternatively, I would have extended the current only regex to add the `-buildXX` suffix. In any case what I want to avoid is publishing from arbitrary branches.

This is a fine compromise for me. I'll leave this as is for the day though, and revert, and extend when I'm done releasing.

lluiscampos:
> > @oleorhagen I see. But `MENDER_VERSION` can still be `<release>-build'n'` (in the GitLab variables and/or in the code) while `mender-convert` is `2.2.x`, hence still being published from a verified and reviewed branch. This is how I have done it previously.
> 
> Hmm, I see. But how did the release-job show up then? It didn't with me.

Building from mender-convert `2.2.x` or from `2.2.1-build3`?

> 
> > @oleorhagen Alternatively, I would have extended the current only regex to add the `-buildXX` suffix. In any case what I want to avoid is publishing from arbitrary branches.
> 
> This is a fine compromise for me. I'll leave this as is for the day though, and revert, and extend when I'm done releasing.

:ok_hand:

oleorhagen:
> Building from mender-convert `2.2.x` or from `2.2.1-build3`?

Ahh, I see. So it would have been okay building from the branch here? (I know they're the same sha).

I think I already know the answer here

oleorhagen:
@lluiscampos this also?

lluiscampos:
@oleorhagen :+1:

PR #252, opened 2020-11-02
Remove the 'only' tag from the publish job

oleorhagen:
It is already manual, so really no need for the 'only' tag.

Changelog: None

(cherry picked from commit 23d169a4b291354ea0505dbc6f7622e2babab2bc)


kacf (approved):

oleorhagen:
@lluiscampos do you still want this reverted?

lluiscampos:
@oleorhagen yes, we have to. As painfully verified by yourself, we don't "protect" tags so this job can only be run from the listed branches there.

PR #262, opened 2020-11-25
Package image tests

oleorhagen:
Review as a whole, will be one commit in the end.

### DIFF
--- a/scripts/test/test-utils.sh
+++ b/scripts/test/test-utils.sh
@@ -124,7 +124,7 @@ run_tests() {
       mv $file "${file%.img}.sdimg"
   done
 
-  cd ${WORKSPACE}/mender-image-tests
+  cd ${WORKSPACE}
 
   echo "Converted image name: "
   echo "$converted_image_name"
@@ -138,9 +138,9 @@ run_tests() {
     --board-type="${device_type}" \
     --mender-image="${converted_image_name}.sdimg" \
     --sdimg-location="${MENDER_CONVERT_DIR}/deploy" \
-    --ssh-priv-key="../ssh-priv-key/key" \
-    --qemu-wrapper="../../scripts/test/mender-convert-qemu" \
-    tests \
+    --ssh-priv-key="./ssh-priv-key/key" \
+    --qemu-wrapper="../scripts/test/mender-convert-qemu" \
+    mender-image-tests \
     ${pytest_extra_args}
 
   exitcode=$?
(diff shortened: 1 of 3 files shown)


scripts/test/test-utils.sh
```diff
@@ -151,11 +151,11 @@ run_tests() {
...
-  wget -N ${MENDER_ACCEPTANCE_URL}/conftest.py -P $WORKSPACE/mender-image-tests
-  wget -N ${MENDER_ACCEPTANCE_URL}/fixtures.py -P $WORKSPACE/mender-image-tests
+  # wget -N ${MENDER_ACCEPTANCE_URL}/pytest.ini -P $WORKSPACE/mender-image-tests
+  # wget -N ${MENDER_ACCEPTANCE_URL}/common.py -P $WORKSPACE/mender-image-tests
+  # wget -N ${MENDER_ACCEPTANCE_URL}/helpers.py -P $WORKSPACE/mender-image-tests
+  # wget -N ${MENDER_ACCEPTANCE_URL}/conftest.py -P $WORKSPACE/mender-image-tests
+  # wget -N ${MENDER_ACCEPTANCE_URL}/fixtures.py -P $WORKSPACE/mender-image-tests
   mkdir -p $WORKSPACE/mender-image-tests/files
   wget -N ${MENDER_ACCEPTANCE_URL}/files/test-private-EC.pem -P $WORKSPACE/mender-image-tests/files
   wget -N ${MENDER_ACCEPTANCE_URL}/files/test-private-RSA.pem -P $WORKSPACE/mender-image-tests/files
```
lluiscampos, line 161: We don't need to do it now, but these keys files should also be moved to `mender-image-tests` repo. I didn't do it because it would have been a mess with the paths, but now that all the framework code is in the submodule, it should be fine.

tests/conftest.py
```diff
@@ -0,0 +1,16 @@
...
+import os
+import pytest
+
+# Dirty hack for adding the module discoverable (for now)
+import sys
+
+cwd = os.getcwd()
+sys.path.append(
+    cwd + "/mender-image-tests/tests"
+)
```
lluiscampos, line 13: Alternatively, something like this should also work:
```suggestion
sys.path.append(
    os.path.join(os.path.dirname(__file__), "mender-image-tests/tests")
)
```

lluiscampos (changes requested):
This looks promising :climbing:

A part from my minor comments below, I also had a look at the test failures.

* Several platforms fail on `test_partition_alignment`. I suspect this is related to one (or more) build variables from mender-convert not being correctly imported by the new test framework
* All qemu tests fail. This I am happy to say that I know the issue. We have removed the [pytest.ini](https://github.com/mendersoftware/meta-mender/blob/master/tests/acceptance/pytest.ini), which adds `-s` option required to run pytest and fabric together. See also this better documented [pytest.ini](https://github.com/mendersoftware/mender-dist-packages/blob/master/tests/pytest.ini). You can fix this by having such file in the directory from where the tests are run (this is important) or by just adding `-s` in the pytest call. I personally prefer the former, as it will work too on local runs of pytest.

tests/conftest.py
oleorhagen: Nice, this is better

scripts/test/test-utils.sh
oleorhagen: Yes, I'll add it as a follow up. But for now, let's keep the moving parts to a minimum. Only one test left to fix now :-)

lluiscampos: Makes sense.

oleorhagen:
> This looks promising climbing
> 
> A part from my minor comments below, I also had a look at the test failures.
> 
>     * Several platforms fail on `test_partition_alignment`. I suspect this is related to one (or more) build variables from mender-convert not being correctly imported by the new test framework

This I have now resolved (at least locally), and it was due to a logic error in the if test (yours truly).
> 
>     * All qemu tests fail. This I am happy to say that I know the issue. We have removed the [pytest.ini](https://github.com/mendersoftware/meta-mender/blob/master/tests/acceptance/pytest.ini), which adds `-s` option required to run pytest and fabric together. See also this better documented [pytest.ini](https://github.com/mendersoftware/mender-dist-packages/blob/master/tests/pytest.ini). You can fix this by having such file in the directory from where the tests are run (this is important) or by just adding `-s` in the pytest call. I personally prefer the former, as it will work too on local runs of pytest.

Great, adding a `pytest.ini` file to the test directory. Thanks :tada:

lluiscampos (approved):

lluiscampos:
@oleorhagen We need to somehow port this to maintained release branches, or at least do a workaround there as the old branches will still try to wget the common files from `meta-mender/master`.

oleorhagen:
Yes, I thought you guys said you were porting it today?

oleorhagen:
Or should I do it?

kacf:
I was going to, but indeed I think I will not have time today after all. If you have time, please do!

oleorhagen:
Which branches are needed?

kacf:
For now, only dunfell. The branches have some definite differences in terms of running them successfully, so better to take one at a time I think.

lluiscampos:
@oleorhagen @kacf Sorry, I meant release branches of `mender-convert`!

@kacf will follow on meta-mender/dunfell and me on the other meta-mender ones, but the point stands valid for old mender-convert, for which we need fixes before (or during) the release.

@oleorhagen If you have the capacity please go ahead; otherwise I guess @kacf will have to do it as part of the upcoming bugfix releases.

oleorhagen:
> @oleorhagen @kacf Sorry, I meant release branches of `mender-convert`!
> 
Oh... Good, I guess. I'll drop the current migration (I was thinking meta-mender :see_no_evil: )
> @kacf will follow on meta-mender/dunfell and me on the other meta-mender ones, but the point stands valid for old mender-convert release branches, for which we need fixes before (or during) the release.
> 
:+1: 
> @oleorhagen If you have the capacity please go ahead; otherwise I guess @kacf will have to do it as part of the upcoming bugfix releases.

It looks like this will fall on @kacf for now I'm afraid

PR #263, opened 2020-11-27
MEN-4097: Install Mender Shell addon

lluiscampos:
* Changelog: Support installing mender-shell addon. Not installed by
default, it can be configured using MENDER_ADDON_SHELL_INSTALL and
MENDER_ADDON_SHELL_VERSION variables.
    
* Changelog: Set mender-shell version to master
    
* Changelog: Create demo configuration for Mender Shell addon in
bootstrap-rootfs-overlay-demo-server.sh script

### DIFF
--- a/mender-convert-modify
+++ b/mender-convert-modify
@@ -49,6 +49,7 @@ echo "Running $(basename $0): $@"
 source modules/bootstrap.sh
 source modules/disk.sh
 source modules/probe.sh
+source modules/deb.sh
 
 # The mender_convert_config is always used and provides all the defaults
 declare -a configs=("configs/mender_convert_config")
@@ -95,24 +96,27 @@ mkdir -p work/rootfs
 sudo mount ${boot_part} work/boot
 sudo mount ${root_part} work/rootfs
 
-mkdir -p work/mender-deb/files
+mkdir -p work/mender-client-deb/files
 
 log_info "Installing Mender client and related files"
 
 deb_arch=$(probe_debian_arch_name)
-DEBIAN_REVISION="-1"
-if [ "${MENDER_CLIENT_VERSION}" = "latest" ]; then
-  DEBIAN_REVISION=""
-fi
-deb_name="mender-client_${MENDER_CLIENT_VERSION}${DEBIAN_REVISION}_${deb_arch}.deb"
-run_and_log_cmd "wget -Nq ${MENDER_STORAGE_URL}/${MENDER_CLIENT_VERSION}/dist-packages/debian/${deb_arch}/${deb_name} -P work/mender-deb"
 
-cd work/mender-deb
+if [ "${MENDER_CLIENT_VERSION}" = "latest" ]; then
+  deb_name=$(deb_from_repo_get "work/mender-client-deb" ${MENDER_APT_REPO_URL} ${deb_arch} "stable" "mender-client" "latest")
+elif [ "${MENDER_CLIENT_VERSION}" = "master" ]; then
+  deb_name=$(deb_from_repo_get "work/mender-client-deb" ${MENDER_APT_REPO_URL} ${deb_arch} "experimental" "mender-client" "latest")
+else
+  DEBIAN_REVISION="-1"
+  deb_name=$(deb_from_repo_get "work/mender-client-deb" ${MENDER_APT_REPO_URL} ${deb_arch} "experimental" "mender-client" "${MENDER_CLIENT_VERSION}${DEBIAN_REVISION}")
+fi
+
+cd work/mender-client-deb
 run_and_log_cmd "ar -xv ${deb_name}"
 run_and_log_cmd "sudo tar xJf data.tar.xz -C files"
 cd - > /dev/null 2>&1
 
-run_and_log_cmd "sudo rsync --archive --keep-dirlinks --verbose work/mender-deb/files/ work/rootfs/"
+run_and_log_cmd "sudo rsync --archive --keep-dirlinks --verbose work/mender-client-deb/files/ work/rootfs/"
 
 if [ "${MENDER_ENABLE_SYSTEMD}" == "y" ]; then
   run_and_log_cmd "sudo ln -sf /lib/systemd/system/mender-client.service \
--- a/mender-convert-modify
+++ b/mender-convert-modify
@@ -96,27 +96,22 @@ mkdir -p work/rootfs
 sudo mount ${boot_part} work/boot
 sudo mount ${root_part} work/rootfs
 
-mkdir -p work/mender-client-deb/files
+mkdir -p work/deb-packages
 
 log_info "Installing Mender client and related files"
 
 deb_arch=$(probe_debian_arch_name)
 
 if [ "${MENDER_CLIENT_VERSION}" = "latest" ]; then
-  deb_name=$(deb_from_repo_get "work/mender-client-deb" ${MENDER_APT_REPO_URL} ${deb_arch} "stable" "mender-client" "latest")
+  deb_name=$(deb_from_repo_get "work/deb-packages" ${MENDER_APT_REPO_URL} ${deb_arch} "stable" "mender-client" "latest")
 elif [ "${MENDER_CLIENT_VERSION}" = "master" ]; then
-  deb_name=$(deb_from_repo_get "work/mender-client-deb" ${MENDER_APT_REPO_URL} ${deb_arch} "experimental" "mender-client" "latest")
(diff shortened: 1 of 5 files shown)


mender-convert-modify
```diff
@@ -95,30 +96,49 @@ mkdir -p work/rootfs
...
-DEBIAN_REVISION="-1"
+
 if [ "${MENDER_CLIENT_VERSION}" = "latest" ]; then
-  DEBIAN_REVISION=""
+  deb_name=$(deb_from_repo_get "work/deb-packages" ${MENDER_APT_REPO_URL} ${deb_arch} "stable" "mender-client" "latest")
+elif [ "${MENDER_CLIENT_VERSION}" = "master" ]; then
+  deb_name=$(deb_from_repo_get "work/deb-packages" ${MENDER_APT_REPO_URL} ${deb_arch} "experimental" "mender-client" "latest")
+else
+  DEBIAN_REVISION="-1"
+  deb_name=$(deb_from_repo_get "work/deb-packages" ${MENDER_APT_REPO_URL} ${deb_arch} "experimental" "mender-client" "${MENDER_CLIENT_VERSION}${DEBIAN_REVISION}")
```
kacf, line 111: Shouldn't this be `stable`? I assume the `else` block is when you specify an exact version.

mender-convert-modify
```diff
@@ -95,30 +96,49 @@ mkdir -p work/rootfs
...
+
+  log_info "Installing Mender addons"
+
+  if [ "${MENDER_ADDON_SHELL_VERSION}" = "latest" ]; then
+    deb_name=$(deb_from_repo_get "work/deb-packages" ${MENDER_APT_REPO_URL} ${deb_arch} "stable" "mender-shell" "latest")
+  elif [ "${MENDER_ADDON_SHELL_VERSION}" = "master" ]; then
+    deb_name=$(deb_from_repo_get "work/deb-packages" ${MENDER_APT_REPO_URL} ${deb_arch} "experimental" "mender-shell" "latest")
+  else
+    DEBIAN_REVISION="-1"
+    deb_name=$(deb_from_repo_get "work/deb-packages" ${MENDER_APT_REPO_URL} ${deb_arch} "experimental" "mender-shell" "${MENDER_ADDON_SHELL_VERSION}${DEBIAN_REVISION}")
```
kacf, line 131: Same here.

modules/deb.sh
```diff
@@ -0,0 +1,88 @@
...
+  fi
+  local -r deb_package="$(pwd)/${1}"
+  local -r dest_dir="$(pwd)/${2}"
+
+  local -r extract_dir=$(mktemp -d)
+  cd ${extract_dir}
+  run_and_log_cmd "ar -xv ${deb_package}"
+  mkdir -p files
+  run_and_log_cmd "sudo tar xJf data.tar.xz -C files"
+  cd - > /dev/null 2>&1
```
kacf, line 83: Why such silencing? Isn't this already silent? And if it fails, you would want to know, or?

mender-convert-modify
lluiscampos: Thanks for asking, this might indicate that I should split the helper functions it two completely ones.

If you [look carefully at the code](https://github.com/mendersoftware/mender-convert/pull/263/commits/6fd2060b907af78b5abf4b955b9896e14e90e7e1#diff-3ef5430cd3b9a9759c354bc7fea58f7f0f49179cf317f5f3f5576d78a70a680fR52-R56), you can see that for specific version the distribution (i.e. stable or experimental) is ignored and the function get the deb package directly from the pool (where all packages are together, independently on the distribution).

As this caused a question by your side, do you think we should split `deb_from_repo_get` in two functions? The interface is somewhat already confusing with two optional parameters and one possibly ignored one.

modules/deb.sh
lluiscampos: To be hones I don't know why. The code was there and I just refactor it. @mirzak do you have a good reason to keep the silencing?

mirzak: Actually it is not silent by default, e.g:

```
mirzak@mk2000 ~
$ cd /tmp/
mirzak@mk2000 /tmp
$ cd -
/home/mirzak
```

Mostly a cosmetic thing as above print would pop-up in mender-convert output, and look malformed.

kacf: Oh, you're right, `cd -` is not silent, only `cd actual-name` is. I guess I had not noticed before.

mender-convert-modify
kacf: Hmm, you're right, this is confusing. Since the check inside `deb_from_repo_get` is `"${version}" = "latest"`, doesn't that mean that `stable` is the only one actually used?

I think I agree then, it would be better to split the function.

lluiscampos: Thanks. I will rework...

(the check inside the function for latest is, for example to get latest from experimental (which is master))

lluiscampos: @kacf I changed it now into two different functions. See last commit.

mender-convert-modify
```diff
@@ -95,30 +96,49 @@ mkdir -p work/rootfs
...
+    DEBIAN_REVISION="-1"
+    deb_name=$(deb_from_repo_pool_get "work/deb-packages" ${MENDER_APT_REPO_URL} ${deb_arch} "mender-shell" "${MENDER_ADDON_SHELL_VERSION}${DEBIAN_REVISION}")
+  fi
+
+  deb_extract_package "work/deb-packages/${deb_name}" "work/rootfs/"
+
+  if [ "${MENDER_ENABLE_SYSTEMD}" == "y" ]; then
+    run_and_log_cmd "sudo ln -sf /lib/systemd/system/mender-shell.service \
+          work/rootfs/etc/systemd/system/multi-user.target.wants/mender-shell.service"
+  fi
```
lluiscampos, line 139: @merlin-northern This piece enables the service by default. Is this something that we don't want? Then we have two options:
* Have a separate variable `MENDER_ADDON_SHELL_ENABLE_SYSTEMD`
* Ignore it all together

I don't like any of the above, do we have other ideas?

lluiscampos:
Rebased on top of latest master to bring the framework changes

kacf (approved):
Much better! Thanks!

mender-convert-modify
merlin-northern: The service should *not* be enabled (as in: started) by default. At least that is what I understood from all the discussions and comments by Eystein (@estenberg). if you do not like a separate variable, we can safely let user to start it, and do not enable it by default.

lluiscampos:
Design rework following last discussions. Updated Changelog's accordingly.

@mirzak @merlin-northern please have a look at the [bootstrap-rootfs-overlay-demo-server.sh changes](https://github.com/mendersoftware/mender-convert/pull/263/commits/9e96418719e4c9f068c2cc5af9b2a01d394be15e#diff-65736fffc12a25bb909fed551b6cc68aa0d30e0334a7c5ad723fe6d9324e640bR69)

merlin-northern (approved):
lgtm!

estenberg:
I'm late here, but is this now enabled in demo mode as we agreed?

I know we might only have discussed that for the yocto demo layer, but it would at least be important for the RPi images we offer as part of Get started: they should support remote terminal out of the box I believe - as they're primarily for testing all the features.

lluiscampos:
@estenberg As it is now, it is not installed nor enabled in the pre-converted images. It is configured (but still not installed by default) with the bootstrap demo script with this PR and will be configured (but still not installed) in other bootstrap scripts in [MEN-4212](https://tracker.mender.io/browse/MEN-4212).

My original idea was to enable, but the general agreement (ping @merlin-northern ) was not to. If this needs to change again, we shall decide ASAP.

estenberg:
To me it seems very important to enable for the Get started images:
* https://docs.mender.io/get-started/preparation/prepare-a-raspberry-pi-device
* https://docs.mender.io/get-started/preparation/prepare-a-virtual-device

Otherwise it will not be possible to easily try and expose this feature to users (they would have to rebuild their own images, which is a tough thing to ask from a new user).

So as the minimum short-term solution we could just build those images with Remote Terminal enabled.

However, since we include it in the demo layer in Yocto I think it would be better if we enable this as a "demo mode" feature in general to make it consistent -- just like we shorten poll intervals consistently for eval / demo purposes.

Sorry if it wasn't clear that enabling RT is important for evaluation / testing, I believe I covered it but maybe not enough. Basically during evaluation / demo we want to enable everything easily and make it very responsive.

If there is still something to discuss / tradeoffs perhaps we could set aside some time on Wed meetings?

lluiscampos:
I added this point to today's meetings.

PR #265, opened 2020-12-01
Use separate chown and chgrp commands when creating rootfs overlay.

kacf:
This is needed for MacOS compatibility as the default chown version there does
not understand the user.group syntax.

Changelog: Title

Signed-off-by: Drew Moseley <drew.moseley@northern.tech>
(cherry picked from commit d8531dd1544bd34e0998855496881b6335cf684f)

PR #267, opened 2020-12-02
Freeze MENDER_ACCEPTANCE_URL 

oleorhagen:
mender-image-tests packaged cherry

### DIFF
--- a/scripts/test/test-utils.sh
+++ b/scripts/test/test-utils.sh
@@ -1,4 +1,4 @@
-MENDER_ACCEPTANCE_URL="https://raw.githubusercontent.com/mendersoftware/meta-mender/master/tests/acceptance"
+MENDER_ACCEPTANCE_URL="https://raw.githubusercontent.com/mendersoftware/meta-mender/c791015fb14f27b7b79143494ee50e42f70c4661/tests/acceptance"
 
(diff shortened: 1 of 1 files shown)


oleorhagen:
This currently blocks: https://github.com/mendersoftware/mender-convert/pull/265

scripts/test/test-utils.sh
```diff
@@ -134,10 +136,10 @@ run_tests() {
 }
 
 get_pytest_files() {
-  wget -N ${MENDER_ACCEPTANCE_URL}/pytest.ini -P $WORKSPACE/mender-image-tests
```
lluiscampos, line 137: I suggest we just modify `MENDER_ACCEPTANCE_URL` to something like `https://raw.githubusercontent.com/mendersoftware/meta-mender/c791015fb14f27b7b79143494ee50e42f70c4661/tests/acceptance` (last commit before the refactoring.

The reason is that 2.2.x is not prepared to execute the runtime tests (`test_update.py`) and it is not worth the effort.

lluiscampos (changes requested):
I suggest a completely different (and simple) workaround. See comment below.

lluiscampos (approved):
Thanks. Please adjust the PR tittle :)

PR #270, opened 2020-12-09
2.2.x: Better parameter checks for read-only options.

kacf:


lluiscampos:
@kacf I think this deserves a `Changelog` tag. At least here that it will result in a bugfix release (I didn't arrive on time for the `master`'s PR)

kacf:
Makes sense. I will submit a followup to master as well, after this is merged.

lluiscampos (approved):

kacf:
Pushed a corresponding change for master [here](https://github.com/mendersoftware/mender-convert/commit/6ff145401a6abb6039759b5939891fc335bfa60a).

PR #271, opened 2020-12-10
Use unique work directories when building with docker.

drewmoseley:
This allows multiple builds to be run in parallel.

Changelog: Title

### Description

Thank you!

### DIFF
--- a/docker-mender-convert
+++ b/docker-mender-convert
@@ -25,13 +25,23 @@ MENDER_CONVERT_DIR="$(pwd)"
 #Pass in version in-case we are added as a git submodule
 MENDER_CONVERT_VERSION=$(git_mender_convert_version)
 
+# Create a unique work directory so we can run multiple containers in parallel
+WORK_DIR=$(mktemp -d -p $(pwd) -t work.XXXX)
+WORK_SUFFIX=$(echo $WORK_DIR | awk -F. '{print $NF}')
+LOG_FILE="convert.log.${WORK_SUFFIX}"
+
+set +e
 docker run \
   --rm \
   -v $MENDER_CONVERT_DIR:/mender-convert \
+  -v $WORK_DIR:/mender-convert/work \
   --privileged=true \
   --cap-add=SYS_MODULE \
   -v /dev:/dev \
   -v /lib/modules:/lib/modules:ro \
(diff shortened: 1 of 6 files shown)


drewmoseley:
Looks like this requires "sudo" in the host OS.

docker-mender-convert
```diff
@@ -25,13 +25,21 @@ MENDER_CONVERT_DIR="$(pwd)"
...
   --privileged=true \
   --cap-add=SYS_MODULE \
   -v /dev:/dev \
   -v /lib/modules:/lib/modules:ro \
   --env MENDER_ARTIFACT_NAME=${MENDER_ARTIFACT_NAME} \
   --env MENDER_CONVERT_VERSION=${MENDER_CONVERT_VERSION} \
   $IMAGE_NAME "$@"
+
+mv ${WORK_DIR}/convert.log ./convert.log.${WORK_SUFFIX}
+sudo rm -rf ${WORK_DIR}
```
kacf, line 45: This step can be done inside docker instead to avoid the need for sudo.

lluiscampos (commented):
One disadvantage that I can see is that the log is lost (say, hidden) when the tool fails and aborts half way. At least we echo at the beginning which is the workdir being used, so that on failures it can be inspected manually.

kacf:
> One disadvantage that I can see is that the log is lost (say, hidden) when the tool fails and aborts half way. At least we echo at the beginning which is the workdir being used, so that on failures it can be inspected manually.

Then I suggest putting the log in a different directory, either the deploy directory, or a dedicated volume for logs. Then make sure it has the same owner as the directory containing it.

docker-mender-convert
drewmoseley: The _contents_ of WORK_DIR can be removed but not work.XXXX in the parent.

drewmoseley:
I reworked a bit to address the above comments and hopefully fix the CI failures.

drewmoseley:
@kacf @lluiscampos @mirzak  any thoughts on the latest version of this? It's going out to a customer soon and if it's merged it's one less step for me.

oleorhagen (approved):
LGTM

kacf (approved):

kacf:
Yes, sorry, this got a bit lost.

PR #272, opened 2020-12-11
[tests cfg] Rename feature variable to MENDER_FEATURES

lluiscampos:
Following recent changes in meta-mender and the test framework.

### DIFF
--- a/mender-convert-package
+++ b/mender-convert-package
@@ -340,15 +340,15 @@ log_info "Conversion has completed! \o/"
 if [ "${MENDER_GRUB_EFI_INTEGRATION}" == "y" ]; then
   boot_part_mountpoint="/boot/efi"
 
-  # This is the name of the DISTRO_FEATURES in Yocto
-  distro_feature="mender-grub"
+  # This is the name of the MENDER_FEATURES in Yocto
+  bootloader_feature="mender-grub"
 else
   boot_part_mountpoint="/uboot"
 
-  # This is the name of the DISTRO_FEATURES in Yocto
-  distro_feature="mender-uboot"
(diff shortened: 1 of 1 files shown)


mender-convert-package
```diff
@@ -341,14 +341,14 @@ if [ "${MENDER_GRUB_EFI_INTEGRATION}" == "y" ]; then
   boot_part_mountpoint="/boot/efi"
 
   # This is the name of the DISTRO_FEATURES in Yocto
```
oleorhagen, line 343: Should this be updated then also? `DISTRO_FEATURES` that is?

mender-convert-package
```diff
@@ -364,7 +364,7 @@ MENDER_PARTITION_ALIGNMENT="${MENDER_PARTITION_ALIGNMENT}"
 MENDER_STORAGE_TOTAL_SIZE_MB="${MENDER_STORAGE_TOTAL_SIZE_MB}"
 MENDER_UBOOT_ENV_STORAGE_DEVICE_OFFSET="12582912"
 MENDER_ARTIFACT_NAME="${artifact_name}"
-DISTRO_FEATURES="${distro_feature}"
+MENDER_FEATURES="${mender_feature}"
```
oleorhagen, line 367: Also, for symmetry, should the variable be named `mender_features` too?

oleorhagen (approved):

mender-convert-package
lluiscampos: @oleorhagen :+1: 

See the force-push, I decided to use `bootloader_feature` and `mender_features` accordingly. Syntactically, it should make more sense now.

kacf (approved):

PR #274, opened 2020-12-14
mender-convert-modify: remove superfluous 'rm' in empty dir cleanup

mirzak:
Should be back-ported to 2.2.x as well.


kacf (approved):

lluiscampos:
> Should be back-ported to 2.2.x as well.

See #275

PR #266, opened 2020-12-01
Use numeric uid and gid for better cross-compatibility.

drewmoseley:
Ubuntu calls group 0 "root" but MacOS calls it "wheel".  This change allows
the same code to function on both.

Signed-off-by: Drew Moseley <drew.moseley@northern.tech>

# External Contributor Checklist

🚨 Please review the [guidelines for contributing](https://github.com/mendersoftware/mender/blob/master/CONTRIBUTING.md) to this repository.

- [ ] Make sure that all commits have a [`Changelog`](https://github.com/mendersoftware/mender/blob/master/CONTRIBUTING.md#changelog-tags) tag. If nothing should be added to the Changelog, add a `Changelog: None` tag. If there is a change, add `Changelog: Commit`, or `Changelog: Title`, depending on what should be included in the changelog.

- [ ] Make sure that all commits are signed with [`git --signoff`](https://git-scm.com/book/en/v2/Git-Tools-Signing-Your-Work). Also note that the signoff author must match the author of the commit.

### Description

Please describe your pull request.

Thank you!

PR #283, opened 2021-01-07
Use numeric uid and gid for better cross-compatibility.

lluiscampos:
Ubuntu calls group 0 "root" but MacOS calls it "wheel".  This change allows
the same code to function on both.


lluiscampos:
Already reviewed at #266

PR #287, opened 2021-01-14
[pipeline] Override MENDER_ADDON_CONNECT_VERSION on build

lluiscampos:
Extend the mechanism to override mender (client) version to also include
the new addon.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -6,6 +6,8 @@ variables:
   # search for and change them too.
   MENDER_ARTIFACT_VERSION: master
   MENDER_CLIENT_VERSION: latest
+  # MENDER_ADDON_CONNECT_VERSION: latest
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -6,6 +6,8 @@ variables:
   # search for and change them too.
   MENDER_ARTIFACT_VERSION: master
   MENDER_CLIENT_VERSION: latest
+  # MENDER_ADDON_CONNECT_VERSION: latest
```
oleorhagen, line 9: ?

oleorhagen (approved):

.gitlab-ci.yml
lluiscampos: It is not yet published in stable. See https://github.com/mendersoftware/mender-convert/blob/master/configs/mender_convert_config#L131 and https://tracker.mender.io/browse/MEN-4200

PR #295, opened 2021-01-19
2.3.x: Revert "Update to new Raspios image of 2021-02-11."

lluiscampos:
This reverts commit 82f9f85fdb9f876450f1dbb8603ad032a84d7864.

Seems like we are having issues with this image on RPi 4. Revert for now
and follow-up independently of the release.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -10,8 +10,8 @@ variables:
   MENDER_ADDON_CONNECT_VERSION: 1.0.0-build2
   # Make sure to update the link in mender-docs to the new one when changing
   # this.
-  RASPBIAN_URL: https://downloads.raspberrypi.org/raspios_lite_armhf/images/raspios_lite_armhf-2021-01-12/2021-01-11-raspios-buster-armhf-lite.zip
-  RASPBIAN_NAME: 2021-01-11-raspios-buster-armhf-lite
+  RASPBIAN_URL: http://downloads.raspberrypi.org/raspios_lite_armhf/images/raspios_lite_armhf-2020-05-28/2020-05-27-raspios-buster-lite-armhf.zip
+  RASPBIAN_NAME: 2020-05-27-raspios-buster-lite-armhf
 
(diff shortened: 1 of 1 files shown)


kacf:
Not sure if I agree with this. This is the most popular image for the most popular board. Shouldn't we fix this instead?

lluiscampos:
> Not sure if I agree with this. This is the most popular image for the most popular board. Shouldn't we fix this instead?

As discussed a bit privately,

This is an image published few days ago (January 2021) so I don't think it should block a release happening at the same time.

We have to follow-up on this ASAP and independently fix. But IMO the process shall be: integrate with master -> works -> port to release branches -> publish.

I'll create a JIRA issue for it and make noise before next planning.

.gitlab-ci.yml
```diff
@@ -10,8 +10,8 @@ variables:
   MENDER_ADDON_CONNECT_VERSION: 1.0.0-build2
   # Make sure to update the link in mender-docs to the new one when changing
   # this.
-  RASPBIAN_URL: https://downloads.raspberrypi.org/raspios_lite_armhf/images/raspios_lite_armhf-2021-01-12/2021-01-11-raspios-buster-armhf-lite.zip
-  RASPBIAN_NAME: 2021-01-11-raspios-buster-armhf-lite
+  RASPBIAN_URL: http://downloads.raspberrypi.org/raspios_lite_armhf/images/raspios_lite_armhf-2020-05-28/2020-05-27-raspios-buster-lite-armhf.zip
```
alfrunes, line 13: Is this a typo?
```suggestion
  RASPBIAN_URL: https://downloads.raspberrypi.org/raspios_lite_armhf/images/raspios_lite_armhf-2020-05-28/2020-05-27-raspios-buster-lite-armhf.zip
```

lluiscampos: Good catch.

The original comment upgraded http->https, and here I just reverted the full commit without paying attention to that detail. We'll fix this in master and then cherry-pick back

PR #299, opened 2021-01-21
Add bash formatter

lluiscampos:
This is more of a RFC than a strong imposition... what do you think?

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -52,6 +52,25 @@ build:
     paths:
       - image.tar
 
+test:format:
+  stage: test
+  needs: []
+  image: python:3
+  before_script:
+    - pip install beautysh
+  script:
+    # Core script
+    - beautysh --indent-size 2 mender-convert*
+    # Other files
+    - find modules configs scripts -type f
+        -not -name "*.md"
+        -not -wholename "scripts/test/*" |
+        xargs beautysh --indent-size 2
+    # Print diff
+    - git diff HEAD
+    # Actual test: exits with non zero status if diff
+    - git diff-index --quiet HEAD
+
 .template_convert_raspbian: &convert_raspbian
   stage: convert
   needs:
--- a/configs/generic_x86-64_hdd_config
+++ b/configs/generic_x86-64_hdd_config
@@ -40,12 +40,12 @@ MENDER_STORAGE_TOTAL_SIZE_MB=16000
 MENDER_COPY_BOOT_GAP="n"
 
 function platform_modify() {
-    #
-    # Make sure /lib64 exists since the Mender binary requires it.
-    # Some systems put everything under /lib (ie Yocto) and a simple
-    # symlink is enough to find everything Mender needs.
-    #
-    if [ ! -e work/rootfs/lib64 ]; then
-        run_and_log_cmd "ln -s /lib work/rootfs/lib64"
-    fi
+  #
+  # Make sure /lib64 exists since the Mender binary requires it.
+  # Some systems put everything under /lib (ie Yocto) and a simple
+  # symlink is enough to find everything Mender needs.
+  #
+  if [ ! -e work/rootfs/lib64 ]; then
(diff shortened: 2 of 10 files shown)


oleorhagen (approved):
Yes, Yes, and Yes from me (but I really like these things).
Just looking at the amount of things it found should be enough I think.

I've been looking for something similar, but haven't stumbled upon this one! Good find!

.gitlab-ci.yml
```diff
@@ -52,6 +52,25 @@ build:
...
 
+test:format:
+  stage: test
+  needs: []
+  image: python:3
+  before_script:
+    - pip install beautysh
+  script:
+    # Core script
+    - beautysh --indent-size 2 mender-convert*
```
kacf, line 63: I guess this becomes more of a vote, but could we make it `4` instead of `2` to match Python? Otherwise this becomes the third indentation style (Golang-Tabs, Python-4spc) in our projects. It is also especially relevant in Yocto context, where shell and Python are mixed in the same file (not that I expect it to handle this, but I'm talking about our policy).

I haven't checked explicitly, but I suspect we have more 4-space scripts than 2-space scripts in general. Mender-convert seems to have a bit of both.

configs/generic_x86-64_hdd_config
```diff
@@ -18,9 +18,9 @@
 #
 # Converted with the following command:
 #
-#	MENDER_ARTIFACT_NAME=release-1 ./docker-mender-convert \
-#		--disk-image input/centos.img \
-#		--config configs/centos_x86-64_hdd_config \
+#	MENDER_ARTIFACT_NAME=release-1 ./docker-mender-convert
+#		--disk-image input/centos.img
+#		--config configs/centos_x86-64_hdd_config
```
kacf, line 23: This is wrong, the comment is incorrect now. Not a huge deal, I know, but can we make it ignore comments completely?

kacf (commented):
Nice, this area has been quite inconsistent!

.gitlab-ci.yml
lluiscampos: I am in favor for 4 spaces. Not only for the reasons you mentioned above, but because it is the default in `beautysh` and there is no way to configure it via config file -> Meaning, every time you run it manually you need to remember this cli parameter.

This will however modify virtually all the code of the core scripts in this repository. Are we sure about it?

configs/generic_x86-64_hdd_config
lluiscampos: Yeah, I know. I'll investigate a bit more.

lluiscampos: I found a way to disabling formatting on this part of the file. See upcoming new commit.

.gitlab-ci.yml
oleorhagen: Just to add here: When we started out, `mender-convert` was supposed to follow the guidelines, as laid out by google's [shell-guide](https://google.github.io/styleguide/shellguide.html#s5.1-indentation). Which is why `2` was chosen. 

I would personally choose to stay with the `2` spaces which is already here. Both for consistency with the style already here, and because I _personally_ think `4` is a little verbose for shell scripts. The argument to beatush isn't really an issue imo

kacf: I think so. The opposite would modify most files in some other repository, so I think it's a catch-22. I'm happy to bring it up to a larger vote though.

lluiscampos: Okey. One voice against a bit change is enough to block it.

Good points, by the way. Thanks!

oleorhagen: Also, I see [this](https://github.com/mvdan/sh) is a pretty popular shell-formatter, as compared to `beautysh`. Should we consider it?

NilsOlavKJohansen (approved):
Good idea! It looks better, it also corrected my comment error. That's great!

lluiscampos:
As discussed in slack, we will go for https://github.com/mvdan/sh. New PR will be coming up soon :)

PR #301, opened 2021-01-25
Add a formatter, and format all code files

oleorhagen:
This is a follow up of the old Format PR (Note to self, link to it tomorrow).

This adds a CI job, which verifies the formatting

And: formats all the source code files with according to the new formatting rules.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -52,6 +52,53 @@ build:
     paths:
       - image.tar
 
+test:format:
+  stage: test
+  needs: []
+  image: alpine:3.13
+  before_script:
+    - apk update
+    - apk upgrade
+    - apk add --update bash git shfmt
+    - shfmt -version
+  script:
+    - /bin/bash
+    - SCRIPTS=$(find modules configs scripts -type f
+            -not -name "*.md"
+            -not -wholename "scripts/test/*")
+    # shfmt
+    ## Language=Bash
+    #     -ln bash
+    ## Indent=4 spaces
+    #     -i 4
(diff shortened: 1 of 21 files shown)


oleorhagen:
Follow up of: https://github.com/mendersoftware/mender-convert/pull/299

.gitlab-ci.yml
```diff
@@ -52,6 +52,53 @@ build:
...
+  before_script:
+    - apk update
+    - apk upgrade
+    - apk add --update bash git shfmt
+    - shfmt -version
+  script:
+    - /bin/bash
+    - SCRIPTS=$(find modules configs scripts -type f
+            -not -name "*.md"
+            -not -wholename "scripts/test/*")
```
lluiscampos, line 68: I haven't seen this syntax before. Does this work as expected? Opens `bash` and then runs the rest of the script?

lluiscampos (approved):
Nice! Thanks for following-up, :medal_sports: for you

.gitlab-ci.yml
oleorhagen: 1. Works as expected
2. Surprised me too
3. We could stick to `sh`, but I like `Bash` more.

lluiscampos: No, this is very nice to know. I don't remember where, but we have used some weird hacks to run things in bash.

Actually, it kind of makes sense. It just runs a `bash` process, and then gitlab keeps piping to/from stdin/stdout.

oleorhagen: haha, yup. It actually makes kinda sense :)

PR #310, opened 2021-02-22
Add a bot for automatically updating the test images

oleorhagen:
This is done through parsing the HTML of the given sites, and reacting to a new
sort of the indices on the given sites.

The links are then updated, and a PR with the new links created.

The bot is run once a week on a schedule

Changelog: None


oleorhagen:
Example run: https://github.com/oleorhagen/mender-convert/pull/2/files

scripts/monitor-bb-sd.js
```diff
@@ -0,0 +1,61 @@
+const jsdom = require("jsdom");
+const { JSDOM } = jsdom;
+
```
lluiscampos, line 3: I think this should be moved out of `scripts` to its own dir, to be separated from user scripts. An alternative could be a different subdir, like `scripts/images-updater` or something like that.

.github/workflows/autobot.yml
```diff
@@ -0,0 +1,33 @@
+name: automatically-update-test-links
```
lluiscampos, line 1: I would name the file to something more specific, like `workflows/update-test-images.yml`

lluiscampos (commented):
Really impressive work, Ole.

Unfortunately, I have no possibility of reviewing the javascript code. The workflow part "looks" good :)

I suggest we merge this (after Manuel's review) as a way to evaluate GitHub Actions. If we like it and are cheap to run, we can utilize it more in other repos. :+1:

scripts/monitor-bb-sd.js
oleorhagen: Yes, I agree. I've been playing with the same idea myself.

oleorhagen:
> I suggest we merge this (after Manuel's review) as a way to evaluate GitHub Actions. If we like it and are cheap to run, we can utilize it more in other repos. +1

Actions were surprisingly smooth, to be honest!

oleorhagen:
@lluiscampos there is one last missing piece, which is kinda annoying, as it sort of defeats the ease of use of all this, and that is our own:
```
UBUNTU_IMAGE_URL="https://downloads.mender.io/mender-convert/images/Ubuntu-Bionic-x86-64.img.gz"
```
Since we don't have an index I can parse :/ 

Any ideas on how we could monitor this? I'm sure `s3` has some sort of API, or... :thinking:

lluiscampos:
> @lluiscampos there is one last missing piece, which is kinda annoying, as it sort of defeats the ease of use of all this, and that is our own:
> 
> ```
> UBUNTU_IMAGE_URL="https://downloads.mender.io/mender-convert/images/Ubuntu-Bionic-x86-64.img.gz"
> ```
> 
> Since we don't have an index I can parse :/
> 
> Any ideas on how we could monitor this? I'm sure `s3` has some sort of API, or... thinking

We cannot monitor it... because we build it manually ;)

If we could integrate an x86_64 board we could use its vendor image instead. But until then we need this to have some coverage on this architecture.

scripts/monitor-tinker.js
```diff
@@ -0,0 +1,67 @@
...
+    var document = dom.window.document;
+    var section = document.getElementById("tinker-board-Download");
+    var boards = section.getElementsByTagName("b")
+    for (var i=0; i<boards.length; i++) {
+        if (boards[i].textContent == "TinkerOS-Debian") {
+            var newRef = boards[i].parentElement.href
+            var regNewMatch = newRef.match(reg)
+            var regOldMatch = imageName.match(reg)
+            let vn = parseFloat(regNewMatch.groups.version)
+            let vo = parseFloat(regOldMatch.groups.version)
```
mzedel, line 34: just since it is so clear here and also for consistency: try to choose a variable type generation and stick to it (so `var` or `let / const`)... my recommendation is to default to `const` and only if you think you will change the variable content at some point fall back to `let`... IMHO that helps to think about side-effects during writing + also allows catching some errors if you write somewhere you shouldn't write to (oh and just ~var~ 😉)...
side-effect freedom is also a great reason to use `map/ reduce` as the data flow gets more explicit when assigning the result of a reducer to a variable... (which then again helps in decomposing parts into smaller functions)

scripts/monitor-bb-sd.js
```diff
@@ -0,0 +1,61 @@
...
+        fs.writeFile('test/run-tests.sh', data, (err, data) => {
+            if (err) {
+                console.error(err)
+            }
+        })
+    } catch (err) {
+        console.error(err)
+        process.exit(1)
+    }
+}
```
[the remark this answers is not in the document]
mzedel: besides this you _could_ use `const fs = require('fs').promises;` and then get to
```
const fs = require('fs').promises
const updateURLLink = async (newLine, target) => {
  const fileContent = await fs.readFile('test/run-tests.sh', 'utf8');
  const data = fileContent.replace(RegExp(`## Auto-update\n${target}=.*`), `## Auto-update\n${newLine}`);
  const result = await fs.writeFile('test/run-tests.sh', data);
  if (result) {
    console.error(result);
  }
};
```
or 
```
const fs = require('fs').promises
const updateURLLink = (newLine, target) =>
  fs.readFile('test/run-tests.sh', 'utf8').then(fileContent => {
    const data = fileContent.replace(RegExp(`## Auto-update\n${target}=.*`), `## Auto-update\n${newLine}`);
    return fs.writeFile('test/run-tests.sh', data);
  })
  .catch(err => {
    console.log(err);
    process.exit(1);
  });
```
but unless you want to depend on the result somewhere that's purely FYI

scripts/monitor-bb.js
```diff
@@ -0,0 +1,76 @@
...
+        return acc
+        }, [])
+        .sort((a,b) => {
+            return Date.parse(b) - Date.parse(a)
+        })
+    if (matches[0] !== latestDate) {
+        console.error("We've got a new release! \\o/");
+        console.error(`${url}/${matches[0]}`);
+        console.error("Old:");
+        console.error(`${url}/${latestDate}/`);
```
mzedel, line 54: just FYI: the template strings also support multilines
```suggestion
        console.error(`
        We've got a new release! \\o/
        ${url}/${matches[0]}
        Old:
        ${url}/${latestDate}/`);
```

scripts/monitor-tinker.js
```diff
@@ -0,0 +1,67 @@
...
+            if (vn > vo) {
+                console.error("A new version has been released \\o/")
+                var newImage = regNewMatch[0]
+                updateURLLink(`TINKER_IMAGE_URL=\"${newImage}\"`)
+                break;
+            } else if ( vn == vo ) {
+                let dn = parseInt(regNewMatch.groups.date)
+                let _do = parseInt(regOldMatch.groups.date)
```
mzedel, line 42: any particular reason to use `_do` instead of `do` or (if your text editor supports autocompletion) `dateOld` to be extra explicit 😛

mzedel (commented):
I left some comments to consider - or not!
it is quite repetitive and I can't think of a really super nice alternative - one idea I would suggest would be to put the `JSDOM` data retrieval into a shared function (something like `getImageMatch(url, targetId, targetTagName, reducer)` - query the document depending on targetId being set by elementId or tagname or both) and then just return the result of the reducer... you could get away with a 
```
const boardVendorMap = {
  BBB_DEBIAN_SDCARD_IMAGE_URL: {
    url: 'https://happy.place',
    targetId: ...,
    targetTagName: ...,
    reducer: (accu, item) => { ...; return accu; }
  },
  ...
};
```
and then call 
```
const { url, targetId, targetTagName, reducer } = boardVendorMap.BBB_DEBIAN_SDCARD_IMAGE_URL;
const results = getImageMatch(url, targetId, targetTagName, reducer);
...continue with results
```
but - that is still not so nice I think...

scripts/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "dependencies": {
+    "bent": "^7.3.12",
```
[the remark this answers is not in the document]
oleorhagen: Sure is!

scripts/monitor-tinker.js
```diff
@@ -0,0 +1,67 @@
...
+} catch (err) {
+    console.error(err)
+    process.exit(1)
+}
+
+JSDOM.fromURL(url, {}).then(dom => {
+    var document = dom.window.document;
+    var section = document.getElementById("tinker-board-Download");
+    var boards = section.getElementsByTagName("b")
+    for (var i=0; i<boards.length; i++) {
```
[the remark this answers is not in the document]
oleorhagen: haha, holy shit, I'm gonna have to read this sentence thrice!

scripts/monitor-tinker.js
oleorhagen: I was afraid it was a keyword :see_no_evil:

mzedel: oops it really is... damn.. ah well - I just go for the longform... `while this is for do not key of` as a kebab or something...

scripts/monitor-bb-sd.js
oleorhagen: Moved to `/scripts/linkbot`

oleorhagen:
@mzedel I follow the `common.js` tip for `updateURLLinks`, and it was real nice. The rest I can't be bothered unifying anymore than this. I also changed `monitor-tinker` to use their `API` which I found in the source. 

U happy?

scripts/monitor-ub-server.js
```diff
@@ -0,0 +1,46 @@
+const jsdom = require("jsdom");
```
mzedel, line 1: that seems to be a leftover duplicate of `scripts/linkbot/monitor-ub-server.js`

scripts/.gitignore
```diff
@@ -0,0 +1,2 @@
+node_modules
+Dockerfile
```
mzedel, line 2: any particular reason to exclude this here? (I can imagine one, but based on that I doubt it should be excluded)

scripts/linkbot/monitor-tinker.js
```diff
@@ -0,0 +1,52 @@
+const jsdom = require("jsdom");
+const { JSDOM } = jsdom;
+const bent = require('bent')
+const getJSON = bent('json')
```
mzedel, line 4: it all makes sense now with the `package.json` entry ❤️ 😄

scripts/linkbot/monitor-tinker.js
```diff
@@ -0,0 +1,52 @@
...
+    result.Result.Obj[0].Files.push({
+        Title: "TinkerOS_Debian",
+        Version: `V${major}.${minor}.${patch}`,
+    })
+    let matches = result.Result.Obj[0].Files.filter(obj => obj.Title.match("TinkerOS_Debian"))
+        .sort((a,b) => {
+            let matchA = a.Version.match(versionRegexp)
+            let matchB = b.Version.match(versionRegexp)
+            if (matchA && matchB) {
+                return parseInt(matchB.major) - parseInt(matchA.major) ||
```
mzedel, line 39: another note on javascript - no need to `parseInt` a number string - you can just subtract straight away and the language will take care of figuring out if it works... and if it doesn't work => everything will just fail with the same `NaN` 🎉 😄
# but don't you dare adding string-numbers!

mzedel (changes requested):
if you remove the duplicate file or give me a good reason to keep it I'll be happy... I promise!

scripts/linkbot/monitor-tinker.js
oleorhagen: hahah, I had to try this obviously, and it seems:
`1 - 1 = 0`, and `1 + 1=11`, just as expected xD

oleorhagen: https://images.app.goo.gl/9963oU2BdPyuZT6x8

mzedel (approved):
👍 sweet line-count-batman!

oleorhagen:
na-na-na-na, na-na-na-na-na-na:
:bat: :man: 
:tada:

PR #314, opened 2021-03-03
Enable manually triggering a bot-run

oleorhagen:
Changelog: None

### DIFF
--- a/.github/workflows/automatically-update-test-links.yml
+++ b/.github/workflows/automatically-update-test-links.yml
@@ -2,6 +2,8 @@ name: automatically-update-test-links
 on:
   schedule:
     - cron: "0 03 * * 1"
+  workflow_dispatch:
+
 jobs:
   createPullRequest:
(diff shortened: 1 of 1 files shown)


.github/workflows/automatically-update-test-links.yml
```diff
@@ -2,6 +2,8 @@ name: automatically-update-test-links
 on:
   schedule:
     - cron: "0 03 * * 1"
+  workflow_dispatch:
```
lluiscampos, line 5: Is this syntax correct? Shouldn't it be:
```suggestion
  workflow_dispatch
```

I am not sure at all, but I think the `:` is only added when there are arguments. See https://docs.github.com/en/actions/reference/events-that-trigger-workflows#manual-events

oleorhagen: haha, I thought so too, but apparently it is wrong (I checked).

lluiscampos (approved):

PR #318, opened 2021-03-04
MEN-4256: Remove /etc/mender/artifact_info from RPi pre-converted image

lluiscampos:
Update tests accordingly.

### DIFF
--- a/mender-convert-package
+++ b/mender-convert-package
@@ -144,7 +144,6 @@ rootfs_part_sectors=$(((${disk_image_total_sectors} - ${data_part_sectors} - \
 rootfs_part_sectors=$(disk_align_sectors ${rootfs_part_sectors} ${MENDER_PARTITION_ALIGNMENT})
 
 device_type=$(cat work/rootfs/data/mender/device_type | sed 's/[^=]*=//')
-artifact_name=$(cat work/rootfs/etc/mender/artifact_info | sed 's/[^=]*=//')
 
 # Get the name from the input disk_image
 temp_img_name=$(basename "$disk_image")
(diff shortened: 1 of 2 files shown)


lluiscampos:
Tests PR at https://github.com/mendersoftware/mender-image-tests/pull/42

lluiscampos:
Note that I decided not to add a changelog tag. My reasoning is that the software behavior is not changed with this PR, only the configuration for the distributed image (for which we don't generate a changelog). Let me know if you disagree.

mender-convert-package
```diff
@@ -212,7 +211,7 @@ disk_create_file_system_from_folder "work/rootfs/" "work/rootfs.img" \
 log_info "Copying root filesystem image to deploy directory"
 run_and_log_cmd "cp --sparse=always work/rootfs.img deploy/${image_name}.${image_fs_type}"
 
-mender_create_artifact "${device_type}" "${artifact_name}" "${image_name}"
+mender_create_artifact "${device_type}" "${MENDER_ARTIFACT_NAME}" "${image_name}"
```
oleorhagen, line 214: Question:
Wrong previously?

oleorhagen (approved):

oleorhagen:
> Note that I decided not to add a changelog tag. My reasoning is that the software behavior is not changed with this PR, only the configuration for the distributed image (for which we don't generate a changelog). Let me know if you disagree.

Hmm, yeah, difficult one. I think I'd favour in adding it though, as if I was using this to convert `rpi` images, my output would indeed have changed, as a result of this.

kacf (approved):
Agree with @oleorhagen, I think I'd be in favor of a changelog entry. The profile is usable by anyone, even though we are the primary users.

lluiscampos:
> (...) I think I'd favour in adding it though, as if I was using this to convert `rpi` images, my output would indeed have changed, as a result of this.

I think you are wrong. If you are converting rpi images, you are using `config/raspberrypi*_config` files but not `config/images/raspberrypi_raspbian_config`, which is the one we are modifying in this PR.

@oleorhagen @kacf you can still convince me that we need a Changelog tag, but please note that normal use of the tool has not been altered.

lluiscampos:
Added:

```
Changelog: raspberrypi_raspbian_config: Remove /etc/mender/artifact_info
from image. For pre-converted images, the file is expected to be
populated at setup time.
```

mender-convert-package
lluiscampos: Not really wrong. It was somewhat safer, as `mender-convert-package` did not rely on what `mender-convert-modify` and just read from fs, but not we had to introduce this dependency as the file might have been removed by user hook in `mender-convert-modify`

kacf:
Looks good!

PR #319, opened 2021-03-04
MEN-4117: Support headless configuration on Raspberry Pi

lluiscampos:
By introducing `MENDER_BOOT_PART_MOUNT_LOCATION` config parameter and
setting it to `/boot` for Raspberry Pi

Changelog: New config parameter `MENDER_BOOT_PART_MOUNT_LOCATION` to
specify the mount point for the boot partition. Empty (default) will
result in `/boot/efi` when UEFI boot is supported, `/uboot` otherwise.

Changelog: Set `MENDER_BOOT_PART_MOUNT_LOCATION=/boot` on Raspberry Pi
configuration to support headless configuration.

### DIFF
--- a/configs/mender_convert_config
+++ b/configs/mender_convert_config
@@ -230,6 +230,10 @@ IMAGE_ROOTFS_EXTRA_SPACE="0"
 # why it is missing a "MENDER_" prefix.
 IMAGE_OVERHEAD_FACTOR="1.5"
 
+# User specified mount point for the boot partition. If empty (default), is mounted
+# at '/boot/efi' for UEFI integration, or at '/uboot' for other cases (U-boot)
+MENDER_BOOT_PART_MOUNT_LOCATION=""
+
 source configs/mender_grub_config
 
 # Function to create Mender Artifact
--- a/configs/raspberrypi_config
+++ b/configs/raspberrypi_config
@@ -4,6 +4,9 @@
 # Raspberry Pi does not support GRUB bootloader integration, fallback to U-boot.
 MENDER_GRUB_EFI_INTEGRATION=n
 
+# Use same boot point that upstream to support Headless configuration
+MENDER_BOOT_PART_MOUNT_LOCATION="/boot"
+
 # Nothing to copy
 MENDER_COPY_BOOT_GAP=n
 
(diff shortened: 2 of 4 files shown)


lluiscampos:
I'll test it tomorrow on real hardware, but the code can be reviewed already.

configs/mender_convert_config
```diff
@@ -230,6 +230,10 @@ IMAGE_ROOTFS_EXTRA_SPACE="0"
 # why it is missing a "MENDER_" prefix.
 IMAGE_OVERHEAD_FACTOR="1.5"
 
+# User specified mount point for the boot partition. If empty (default), is mounted
+# at '/boot/efi' for UEFI integration, or at '/uboot' for other cases (U-boot)
```
oleorhagen, line 234: Question:
Is U-boot the only other case? It sounds like it from the description, but `cases` is plural

configs/raspberrypi_config
```diff
@@ -4,6 +4,9 @@
 # Raspberry Pi does not support GRUB bootloader integration, fallback to U-boot.
 MENDER_GRUB_EFI_INTEGRATION=n
 
+# Use same boot point that upstream to support Headless configuration
```
oleorhagen, line 7:
```suggestion
# Use same boot point as upstream to support Headless configuration
```

oleorhagen (approved):
To the point

configs/raspberrypi_config
```diff
@@ -4,6 +4,9 @@
...
+MENDER_BOOT_PART_MOUNT_LOCATION="/boot"
```
kacf, line 8: Hmm, will this work? The reason that we don't mount the boot partition directly on `/boot` is that the kernel is hosted in `/boot`, but it is still just a directory in the rootfs. If we mount the boot partition here, then you cannot update your kernel with a rootfs update.

lluiscampos: It boots and works, but I think you are right with regards to an update with contents in /boot (i.e. Kernel update). TBH I implemented what we said in the JIRA task, but I'll rethink a bit more now.

configs/mender_convert_config
lluiscampos: Yeah, I didn't know how to phrase it well. Is technically speaking "all other cases", as the logic is if EFI this, else /uboot. So in a way we assume that if it is not EFI is uboot, but I guess it could be other case being whatever boot-loader and we still mount it in /uboot dir

lluiscampos:
Discussed with @kacf privately; this is the wrong approach, we don't see any other fix but to modify upstream systemd service files. I'll launch a new PR.

PR #328, opened 2021-03-25
[raspberrypi_config] MEN-4567: Enable UART in U-Boot config.txt

lluiscampos:
This is required for U-Boot to operate correctly, see
https://source.denx.de/u-boot/u-boot/-/blob/master/arch/arm/mach-bcm283x/Kconfig

Changelog: Title

### DIFF
--- a/configs/raspberrypi_config
+++ b/configs/raspberrypi_config
@@ -106,4 +106,13 @@ function platform_modify() {
         run_and_log_cmd "sudo unlink work/rootfs/etc/rc3.d/S01resize2fs_once"
     fi
     run_and_log_cmd "sudo rm -f work/rootfs/etc/init.d/resize2fs_once"
+
+    # Enable UART. This is required for U-Boot to operate correctly, see
+    # https://source.denx.de/u-boot/u-boot/-/blob/master/arch/arm/mach-bcm283x/Kconfig
+    log_info "Enabling UART in U-Boot configuration"
+    cat <<- EOF >> work/boot/config.txt
+
+# Enable UART
+enable_uart=1
+EOF
 }


oleorhagen:
@lluiscampos `RPi3` has the same description:

```
config TARGET_RPI_3_32B
	bool "Raspberry Pi 3 32-bit build"
	help
	  Support for all BCM2837-based Raspberry Pi variants, such as
	  the RPi 3 model B, in AArch32 (32-bit) mode.

	  This option assumes the VideoCore firmware is configured to use the
	  mini UART (rather than PL011) for the serial console. This is the
	  default on the RPi 3. To enable the UART console, the following non-
	  default option must be present in config.txt: enable_uart=1. This is
	  required for U-Boot to operate correctly, even if you only care
	  about the HDMI/usbkbd console.

	  This option creates a build targeting the ARMv7/AArch32 ISA.
	select BCM2837_32B

config TARGET_RPI_3
	bool "Raspberry Pi 3 64-bit build"
	help
	  Support for all BCM2837-based Raspberry Pi variants, such as
	  the RPi 3 model B, in AArch64 (64-bit) mode.

	  This option assumes the VideoCore firmware is configured to use the
	  mini UART (rather than PL011) for the serial console. This is the
	  default on the RPi 3. To enable the UART console, the following non-
	  default option must be present in config.txt: enable_uart=1. This is
	  required for U-Boot to operate correctly, even if you only care
	  about the HDMI/usbkbd console.

	  At the time of writing, the VC FW requires a non-default option in
	  config.txt to request the ARM CPU boot in 64-bit mode:
	  arm_control=0x200

	  The VC FW typically provides ARM "stub" code to set up the CPU and
	  quiesce secondary SMP CPUs. This is not currently true in 64-bit
	  mode. In order to boot U-Boot before the VC FW is enhanced, please
	  see the commit description for the commit which added RPi3 support
	  for a workaround. Since the instructions are temporary, they are not
	  duplicated here. The VC FW enhancement is tracked in
	  https://github.com/raspberrypi/firmware/issues/579.

	  This option creates a build targeting the ARMv8/AArch64 ISA.
	select BCM2837_64B
```

lluiscampos:
> @lluiscampos `RPi3` has the same description:
> 

Ouch. Thanks! I'll rework

configs/raspberrypi3_config
```diff
@@ -4,3 +4,14 @@ MENDER_KERNEL_IMAGETYPE="zImage"
 MENDER_DEVICE_TYPE="raspberrypi3"
 
 source configs/raspberrypi_config
+
+#
+# Enable UART. This is a requirement from U-boot v2021-01, see
+# https://source.denx.de/u-boot/u-boot/-/commit/c6bcf05fcd61fce1092b7fca945982a65d0c723b
```
oleorhagen, line 10: Nit:
This commit does not contain the RPi3 changes

configs/raspberrypi3_config
```diff
@@ -4,3 +4,14 @@ MENDER_KERNEL_IMAGETYPE="zImage"
...
+#
+function enable_uart() {
+    log_info "Enabling UART in U-Boot configuration"
+    run_and_log_cmd "echo -e '\n# Enable UART\nenable_uart=1' >> work/boot/config.txt"
```
oleorhagen, line 14: Feel free to ignore:
```
cat <<EOF>> work/boot/config.txt
# Enable UART
enable_uart=1
EOF
```
I just think it reads better

oleorhagen (approved):

lluiscampos:
Reworked to enable UART for all Raspberry Pi configurations

oleorhagen (approved):

PR #333, opened 2021-03-30
Add hook to create the Mender Configure state folder in the data partition

tranchitella:


kacf:
:+1: Was one minute too late to approve, but it is of course!

PR #343, opened 2021-04-26
MEN-4658: Support Raspberry Pi apps expecting files at /boot

lluiscampos:
By symbolic linking to all existing files in /uboot. This fix deprecates
the previous hand-picked symbolic links.

Changelog: Raspberry Pi: create symbolic links from /boot to /uboot for
all existing files in boot partition. Some Raspberry Pi apps, like Pi
Camera, expect firmware files at /boot.

### DIFF
--- a/configs/raspberrypi_config
+++ b/configs/raspberrypi_config
@@ -86,10 +86,11 @@ function platform_modify() {
     run_and_log_cmd "sudo cp work/rpi/binaries/uboot-git-log.txt work/boot"
 
     # Raspberry Pi applications expect to find this on the device and in some
-    # cases parse the options to determine the functionality.
-    run_and_log_cmd "sudo ln -fs /uboot/config.txt work/rootfs/boot/config.txt"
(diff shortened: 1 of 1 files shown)


lluiscampos:
See [MEN-4658](https://tracker.mender.io/browse/MEN-4658). I'll validate the change later this week on my weekly visit to the office.

oleorhagen (approved):
Nuclear option :boom: 

Be interesting to see how this works!

oleorhagen:
> @oleorhagen I tested it on the pie and realized that this fix does _not_ deprecate the modifying service file need. The reason is that the headless configuration asks user to touch or create files in the FAT partition (aka `/uboot`) and this is done after the conversion. Is kind of obvious but I got confused earlier this week.
> 
> Anyway, if you could have a second look I'd appreciate it pray

Ohh, I see. Makes sense to me :)

oleorhagen (approved):

PR #347, opened 2021-05-03
MEN-4256 Do not install the client by default on Prebuilt RPi images

oleorhagen:
This adds the:

* `MENDER_CLIENT_INSTALL` option, to allow an image to come without the
mender-client installed by default.

The option is on by default, but turned off for the custom image configuration
we use for building the images we provide as a part of our documentation onboarding.

Changelog: Add an option: 'MENDER_CLIENT_INSTALL={y,n}', to allow a converted
image to come without the Mender client installed by default.

- [x] Make sure that all commits have a [`Changelog`](https://github.com/mendersoftware/mender/blob/master/CONTRIBUTING.md#changelog-tags) tag. If nothing should be added to the Changelog, add a `Changelog: None` tag. If there is a change, add `Changelog: Commit`, or `Changelog: Title`, depending on what should be included in the changelog.

### Description

Thank you!

### DIFF
--- a/mender-convert-package
+++ b/mender-convert-package
@@ -349,6 +349,8 @@ log_info "Conversion has completed! \o/"
 
 ##################### Create configuration file for tests ######################
 
+mender_features="mender-convert"
+
 if [ "${MENDER_GRUB_EFI_INTEGRATION}" == "y" ]; then
     boot_part_mountpoint="/boot/efi"
 
@@ -360,7 +362,11 @@ else
     # This is the name of the MENDER_FEATURES in Yocto
     bootloader_feature="mender-uboot"
 fi
-mender_features="${bootloader_feature} mender-convert"
+mender_features="${mender_features} ${bootloader_feature}"
+
+if [ "${MENDER_CLIENT_INSTALL}" = "y" ]; then
+    mender_features="${mender_features} mender-client-install"
+fi
 
 testscfg_add "MENDER_BOOT_PART" "${boot_part_device}"
 testscfg_add "MENDER_ROOTFS_PART_A" "${root_part_a_device}"
--- a/mender-convert-package
+++ b/mender-convert-package
@@ -148,7 +148,7 @@ device_type=$(cat work/rootfs/data/mender/device_type | sed 's/[^=]*=//')
 # Get the name from the input disk_image
 temp_img_name=$(basename "$disk_image")
 
-# Check if user has chooses custom filename
+# Check if user has chosen custom filename
 if [ -z "${DEPLOY_IMAGE_NAME}" ]; then
     # Strip the image suffix from temp_img_name and set the original image name as
     # the output image name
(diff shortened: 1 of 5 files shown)


lluiscampos (approved):
Looks good!

We need to find a tweak for the tests. If you don't have any good idea, we can have a short call for this tomorrow.

Also, for the changelog tag:
```
Changelog: Add an option: `MENDER_CLIENT_INSTALL={y,n}`, to configure install or not the Mender client into the converted image. Defaults to "yes".
```

It is a bit confusing, the option is not to "install the Mender client by default", but just to install it or not. And defaults to yes. Trying to rephrase:

```
Changelog: Add an option: 'MENDER_CLIENT_INSTALL={y,n}', to allow a converted
image to come without the Mender client installed by default.

```

oleorhagen:
> Changelog: Add an option: `MENDER_CLIENT_INSTALL={y,n}`, to install or not the
> Mender client into the converted image. Defaults to "yes".
> ```

How about:

```Added an option `MENDER_CLIENT_INSTALL=y/n`, in order to configure the installation of the Mender client into the converted image. Defaults to `yes`.
```

mender-convert-package
```diff
@@ -212,7 +215,9 @@ disk_create_file_system_from_folder "work/rootfs/" "work/rootfs.img" \
 log_info "Copying root filesystem image to deploy directory"
 run_and_log_cmd "cp --sparse=always work/rootfs.img deploy/${image_name}.${image_fs_type}"
 
-mender_create_artifact "${device_type}" "${artifact_name}" "${image_name}"
+if [ "${MENDER_CLIENT_INSTALL}" = "y" ]; then
+    mender_create_artifact "${device_type}" "${artifact_name}" "${image_name}"
+fi
```
lluiscampos, line 220: I understand why we do this: an Artifact without mender in it is going to fail.

It somehow changes the interface of the whole utility, I suggest we add here an INFO being explicit: `Skipping image Artifact generation, mender Client not installed` (or something like that...)

oleorhagen: I just wanted to get a successful run...

The reason being:

```
    artifact_name=$(cat work/rootfs/etc/mender/artifact_info | sed 's/[^=]*=//')
fi
```
When we don't install the client, the `artifact_info` file does not exist, so we either have to put it back in, or give the artifact name a standard name. 

Also, what is the artifact needed for when it does not contain  a client?

lluiscampos:
@oleorhagen I reworked a bit your PR. Namely:
* Reverting the Artifact creation - as discussed we let it be for the sake of consistency. Added a warning.
* Adding "mender-client-install" as a yocto feature, so that tests can be skip.

See also https://github.com/mendersoftware/mender-image-tests/pull/48

lluiscampos:
The tests look good now. We need the fix from https://github.com/mendersoftware/mender-convert/pull/349 and an small style fix.

Can you @oleorhagen review my new commits? (I cannot ping you on a PR that you are the author).

oleorhagen:
@lluiscampos looks good!

Note to thy and myself, I have not fixed up the Changelog in the first commit still. This is required before merging!

lluiscampos:
> Note to thy and myself, I have not fixed up the Changelog in the first commit still. This is required before merging!

Fixed.

Also rebased on top of master and updated https://github.com/mendersoftware/mender-image-tests/pull/48 (rebased on top of its own master).

Everything should go green now :green_heart:

PR #351, opened 2021-05-10
[dependabot] Track dependencies of linkbot npm package

lluiscampos:


lluiscampos:
I don't think this is very important, really. But we are getting annoying security alerts for these, so better to just keep them up to date... right?

oleorhagen (approved):

mzedel (approved):
Just fyi: we will still get the security notes - as they are issued more frequently than the dependabot check will be run... (as is the case for the GUI repo e.g.)

lluiscampos:
> Just fyi: we will still get the security notes - as they are issued more frequently than the dependabot check will be run... (as is the case for the GUI repo e.g.)

Yeah, I know. But at least on a week time they should disappear...

It is different in GUI repo where javascript code is actually in production, so some of the alerts might be very important to catch immediately.

lluiscampos:
Stopping the build to save resources.

PR #358, opened 2021-06-18
MEN-4462: Add a config for demo configuration during conversion

oleorhagen:
This moves the device configuration to a config script in configs/, from the
previous setup in the bootstrap scripts.

It does so by creating a new resources/ folder in the top-level directory, which
is then populated with the configuration file, and the server parameters only.

When the conversion is run with the demo configuration
configs/mender_convert_demo_config the conversion will install all the add-on's
available, and modify the configuration file in resources to align with our
regular Mender demo configuration.

### DIFF
--- a/configs/mender_convert_demo_config
+++ b/configs/mender_convert_demo_config
@@ -1,0 +1,35 @@
+# Install all the add-on's by default
+MENDER_ADDON_CONNECT_INSTALL="y"
+MENDER_ADDON_CONFIGURE_INSTALL="y"
+
+enable_demo_configuration_in_file() {
+
+    log_info "Modifying mender.conf to follow a demo setup"
+
+    run_and_log_cmd "mkdir -p work/rootfs/etc/mender"
+
+    if [ ! -f work/rootfs/etc/mender/mender.conf ]; then
+        log_error "No Mender configuration file found in the work folder."
+        exit 1
+    fi
+
+    if [ ! -f work/rootfs/etc/mender/mender-connect.conf ]; then
+        log_info "Installing the mender-connect.conf file"
+        cat <<- EOF > work/rootfs/etc/mender/mender-connect.conf
+            {
+                "ShellCommand": "/bin/sh",
+                "User": "root"
+            }
+EOF
+        run_and_log_cmd "sudo chmod 0600 work/rootfs/etc/mender/mender-connect.conf"
+    fi
+
+    log_info "Adding the demo configuration to the Mender client"
+    tmpfile=work/mender.conf.bak
+    run_and_log_cmd "cat work/rootfs/etc/mender/mender.conf |  jq '.InventoryPollIntervalSeconds=5 | .RetryPollIntervalSeconds=30 |    .UpdatePollIntervalSeconds=5' > ${tmpfile}"
+    run_and_log_cmd "cp ${tmpfile} work/rootfs/etc/mender/mender.conf"
+    run_and_log_cmd "sudo chmod 0600 work/rootfs/etc/mender/mender.conf"
+    log_warn "New mender.conf contents: $(cat  work/rootfs/etc/mender/mender.conf)"
+
+}
+PLATFORM_MODIFY_HOOKS+=(enable_demo_configuration_in_file)
--- a/mender-convert-modify
+++ b/mender-convert-modify
@@ -105,7 +105,7 @@ deb_arch=$(probe_debian_arch_name)
 
 if [ "${MENDER_CLIENT_INSTALL}" = "y" ]; then
 
-    log_info "Installing Mender Client version ${MENDER_CLIENT_VERSION}"
+    log_info "Installing Mender client version ${MENDER_CLIENT_VERSION}"
 
     if [ "${MENDER_CLIENT_VERSION}" = "latest" ]; then
         deb_name=$(deb_from_repo_dist_get "work/deb-packages" ${MENDER_APT_REPO_URL} ${deb_arch} "stable" "mender-client")
--- a/mender-convert-modify
+++ b/mender-convert-modify
@@ -290,6 +290,15 @@ EOF
 run_and_log_cmd "sudo cp work/mender.conf.data work/rootfs/data/mender/mender.conf"
 run_and_log_cmd "sudo chmod 600 work/rootfs/data/mender/mender.conf"
 
+if [ -f resources/mender.conf ]; then
+    log_info "Installing the local mender.conf file"
+    run_and_log_cmd "mkdir -p work/rootfs/etc/mender"
+    run_and_log_cmd "cp resources/mender.conf work/rootfs/etc/mender"
+    run_and_log_cmd "sudo chmod 600 work/rootfs/etc/mender/mender.conf"
+else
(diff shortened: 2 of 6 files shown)


configs/mender_convert_demo_config
```diff
@@ -0,0 +1,37 @@
+
```
lluiscampos, line 1: Missing copyright and license

configs/mender_convert_demo_config
```diff
@@ -0,0 +1,37 @@
...
+MENDER_ADDON_CONNECT_INSTALL="y"
+MENDER_ADDON_CONFIGURE_INSTALL="y"
+
+enable_demo_configuration_in_file() {
+
+    log_info "Modifying mender.conf to follow a demo setup"
+
+    run_and_log_cmd "mkdir -p work/rootfs/etc/mender"
+
+    if [ ! -e work/rootfs/etc/mender/mender.conf ]; then
```
lluiscampos, line 12: Shouldn't these be regular files? Then I would use `-f`

configs/mender_convert_demo_config
```diff
@@ -0,0 +1,37 @@
...
+                "User": "root"
+            }
+EOF
+        run_and_log_cmd "sudo chmod 0600 work/rootfs/etc/mender/mender-connect.conf"
+    fi
+
+
+    log_info "Adding the demo configuration to the Mender client"
+    tmpfile=work/mender.conf.bak
+    run_and_log_cmd "cat work/rootfs/etc/mender/mender.conf |  jq '.InventoryPollIntervalSeconds=5 | .RetryPollIntervalSeconds=30 |    .UpdatePollIntervalSeconds=5' > ${tmpfile}"
```
lluiscampos, line 31: Nitpick: extra spaces
```suggestion
    run_and_log_cmd "cat work/rootfs/etc/mender/mender.conf |  jq '.InventoryPollIntervalSeconds=5 | .RetryPollIntervalSeconds=30 | .UpdatePollIntervalSeconds=5' > ${tmpfile}"
```

configs/mender_convert_demo_config
```diff
@@ -0,0 +1,37 @@
...
+    run_and_log_cmd "cp ${tmpfile} work/rootfs/etc/mender/mender.conf"
+    run_and_log_cmd "sudo chmod 0600 work/rootfs/etc/mender/mender.conf"
+    log_warn "New mender.conf contents: $(cat  work/rootfs/etc/mender/mender.conf)"
```
lluiscampos, line 34: I like the idea of having a warning informing, but I think it would be more informative with a diff of what actually has been changed. Or maybe the diff output gets too noisy?

mender-convert-modify
```diff
@@ -105,7 +105,7 @@ deb_arch=$(probe_debian_arch_name)
 
 if [ "${MENDER_CLIENT_INSTALL}" = "y" ]; then
 
-    log_info "Installing Mender Client version ${MENDER_CLIENT_VERSION}"
+    log_info "Installing Mender client version ${MENDER_CLIENT_VERSION}"
```
lluiscampos, line 108: Thanks :+1:

mender-convert-modify
```diff
@@ -290,6 +290,15 @@ EOF
 run_and_log_cmd "sudo cp work/mender.conf.data work/rootfs/data/mender/mender.conf"
 run_and_log_cmd "sudo chmod 600 work/rootfs/data/mender/mender.conf"
 
+log_info "Installing the local mender.conf file"
+if [ ! -e resources/mender.conf ]; then
+    log_fatal "No mender.conf file found in the resources folder.\
+ Did you remember to run the bootstrap script?"
```
lluiscampos, line 296: Mmm. What if a user provides a custom overlay with the config file? Is it not mandatory to run one of the bootstrap scripts?

scripts/bootstrap-rootfs-overlay-hosted-server.sh
```diff
@@ -57,20 +57,15 @@ if [ -e ${output_dir} ]; then
...
 }
 EOF
 
-chmod 600 ${output_dir}/etc/mender/mender.conf
-
 sudo chown -R 0 ${output_dir}
 sudo chgrp -R 0 ${output_dir}
 
-echo "Configuration file for using Hosted Mender written to: ${output_dir}/etc/mender"
+echo "Configuration file for using Hosted Mender written to: ${root_dir}/resources/mender.conf"
```
lluiscampos, line 71: Missing to modify this line in `scripts/bootstrap-rootfs-overlay-demo-server.sh`

lluiscampos (changes requested):
The deign looks good! We achieve exactly what we wanted: separate "client demo settings" from "backend settings". It is only the question (see comment below) on how a user could explicitly write his/her own mender.conf file :man_shrugging:

configs/mender_convert_demo_config
oleorhagen: We don't have a copyright header in any of the configuration files as of now, but maybe we should? :thinking:

mender-convert-modify
oleorhagen: As it is now, it will pass if a bootstrap has been run previously.

If an overlay exists for `/etc/mender/mender.conf` then the demo settings in `mender_convert_demo_conf` will have no effect, as the overlay takes presedence.

But perhaps we should have an extra check that the bootstrap has been run? :thinking: 
I'm not sure to be honest, I mean, if the `resources/mender.conf` file exists, then bootstrap has been run at least once previously.

Difficult topic

oleorhagen:
> The deign looks good! 

Great, then I'm moving on to amending the tests.

We achieve exactly what we wanted: separate "client demo settings" from "backend settings". It is only the question (see comment below) on how a user could explicitly write his/her own mender.conf file man_shrugging

An overlay still takes precedence,  I feel like this is a sane default. You disagree? It is also possible to change the file in the `resources` folder, after bootstrap (perhaps we should notify about this :thinking: ), then only upon adding the demo config the demo intervals are overridden, the rest of the file is untouched.

configs/mender_convert_demo_config
oleorhagen: I like this idea! Giving it a spin

lluiscampos:
> An overlay still takes precedence, I feel like this is a sane default. You disagree? 

I don't disagree, the creating an overlay is now a manual process, so it makes sense that it takes precedence.

What I am concern is that running one of the bootstrap scripts is now a requirement, even when a user is providing a custom mender.conf. I don't know which kind of flows the users are having with mender-convert neither, so it is very hard to tell what can we break with this.

Maybe we can merge it, as is, and have a quick demo tomorrow to gather input from Alan and Eystein.

> It is also possible to change the file in the `resources` folder, after bootstrap (perhaps we should notify about this ), then only upon adding the demo config the demo intervals are overridden, the rest of the file is untouched.

That could also be a valid flow, but we don't need to do anything else to support it.

lluiscampos (approved):

configs/mender_convert_demo_config
oleorhagen: Didn't really add that much I think

PR #364, opened 2021-07-21
Switch to publish Raspberry Pi images using mender-convert version.

kacf:
This is done for several reasons:

* The images do not contain the client anymore.
* We want to have the mender-convert version in the name, so that we
  don't overwrite old, working images.

We also add the build variable `MENDER_CONVERT_PUBLISH_VERSION`, to
allow publishing of tagged mender-convert versions with small fixes on
top.

And finally, we switch the pipeline to auto-publish from tags and from
branch names, like most pipelines do.

Changelog: None

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -13,6 +13,8 @@ variables:
   RASPBIAN_URL: https://downloads.raspberrypi.org/raspios_lite_armhf/images/raspios_lite_armhf-2021-01-12/2021-01-11-raspios-buster-armhf-lite.zip
   RASPBIAN_NAME: 2021-01-11-raspios-buster-armhf-lite
 
+  TEST_MENDER_CONVERT: "true"
+
   # Which version of mender-convert to use in published image name. Normally
   # empty, in which case the branch or tag name will be used, but this variable
   # is available so that it's possible to make small build fixes on an already
@@ -171,6 +173,8 @@ convert_raspbian_raspberrypi4:
       - report_*.html
     reports:
       junit: results_*.xml
+  rules:
+    - if: '$TEST_MENDER_CONVERT == "true"'
 
 .template_test_acceptance_prebuilt_raspberrypi: &test_acceptance_prebuilt_raspberrypi
   <<: *test_acceptance
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -13,6 +13,16 @@ variables:
   RASPBIAN_URL: https://downloads.raspberrypi.org/raspios_lite_armhf/images/raspios_lite_armhf-2021-01-12/2021-01-11-raspios-buster-armhf-lite.zip
   RASPBIAN_NAME: 2021-01-11-raspios-buster-armhf-lite
 
+  # Which version of mender-convert to use in published image name. Normally
+  # empty, in which case the branch or tag name will be used, but this variable
+  # is available so that it's possible to make small build fixes on an already
+  # tagged version.
+  MENDER_CONVERT_PUBLISH_VERSION: ""
+
+  # Whether to publish images. Note that images from branches are always
+  # published, even if this is false.
+  PUBLISH_MENDER_CONVERT_IMAGES: "false"
+
   DEBIAN_FRONTEND: noninteractive
 
   # Docker dind configuration.
@@ -214,7 +224,6 @@ test_acceptance_ubuntu:
     - ./scripts/test/run-tests.sh --only ubuntu
 
 publish:s3:
-  when: manual
   stage: publish
   image: debian:buster
   before_script:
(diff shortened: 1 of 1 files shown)


kacf:
@lluiscampos: I did some smoke testing on this using a `echo Hello world` pipeline, but it's hard to test the real thing here.

.gitlab-ci.yml
```diff
@@ -13,6 +13,12 @@ variables:
   RASPBIAN_URL: https://downloads.raspberrypi.org/raspios_lite_armhf/images/raspios_lite_armhf-2021-01-12/2021-01-11-raspios-buster-armhf-lite.zip
   RASPBIAN_NAME: 2021-01-11-raspios-buster-armhf-lite
 
+  # Which version of mender-convert to use in published image name. Normally
+  # empty, in which case the branch or tag name will be used, but this variable
+  # is available so that it's possible to make small build fixes on an already
+  # tagged version.
+  MENDER_CONVERT_PUBLISH_VERSION: ""
```
lluiscampos, line 20: I don't understand in which cases we are going to use this. Both for master and release branches the image will be continuously updated now, but I guess we will be using only tagged versions in the docs. Is it the idea that after a given release we can do some "hot fix" without tagging? Setting this to `2.5.0-1` or something like that?

If that is the case, I think is better to have this variable to hold only the `-1` and we concatenate it in the script.

.gitlab-ci.yml
```diff
@@ -233,21 +238,26 @@ publish:s3:
...
           s3://$S3_BUCKET_NAME/${RASPBIAN_NAME}/arm/${PUBLISH_NAME}
     -   aws s3api put-object-acl --acl public-read --bucket $S3_BUCKET_NAME
           --key ${RASPBIAN_NAME}/arm/${PUBLISH_NAME}
     - done
 
-  only:
-    - /^(master|[0-9]+\.[0-9]+\.x)$/
+  rules:
+    - if: '$CI_COMMIT_REF_NAME =~ /^(master|[0-9]+\.[0-9]+\.(x|[0-9]+))$/ && $MENDER_CONVERT_PUBLISH_VERSION == ""'
+      when: on_success
```
lluiscampos, line 261: `on_success` is the default action. I like it being explicit, but we haven't use it in other pipelines so for the sake of consistency I'd vote for removing it

.gitlab-ci.yml
```diff
@@ -233,21 +238,26 @@ publish:s3:
...
+    - if: '$CI_COMMIT_REF_NAME =~ /^(master|[0-9]+\.[0-9]+\.(x|[0-9]+))$/ && $MENDER_CONVERT_PUBLISH_VERSION != ""'
+      when: manual
```
lluiscampos, line 263: The rules look okey. An alternative (simplified?) way could be:
```suggestion
  rules:
    - if: '$MENDER_CONVERT_PUBLISH_VERSION != ""'
      when: manual
    - if: '$CI_COMMIT_REF_NAME =~ /^(master|[0-9]+\.[0-9]+\.(x|[0-9]+))$/'
      when: on_success
```

It allows to manually publish from any branch. But maybe could come in handy to publish some candidate to even ask a user to try it out or something. The point is that once we are overriding the version string to publish, I don't see why we need to restrict it to master and release branches.

lluiscampos (commented):
I would like to discuss this a bit more, maybe on a call.

All the "standard" repositories always publish from master and release branches themselves, but from tags only from mender-qa. I think we should keep that principle here too, so that the tag publishing happens when triggered from mender-qa (which is not working ATM, but we have plans to fix).

.gitlab-ci.yml
kacf: True, this is an unnecessary restriction. I'll remove it.

.gitlab-ci.yml
kacf: Yeah, I was internally debating that one for a bit. Will remove it then!

.gitlab-ci.yml
kacf: The main motivation is to be able to publish `<IMAGE>-mender-convert-2.5.0.img.xz`, IOW the one we are doing right now. We can't retag 2.5.0, but we know the only significant change is this `.gitlab-ci.yml` change, therefore it makes sense to publish it with a little cheat: It's not actually the pure 2.5.0 tag, but close enough, and manually verified by us.

The thinking is a little bit like how we publish images from dunfell. Even though we usually publish from the commit that becomes the tag, this is not enforced, and has certainly been sidestepped from time to time.

But I suppose it's possible that this is so rare that it doesn't make sense to submit this variable. We can just do it this once from a PR, and then close it without merging it afterwards.

Do you think we should remove it?

kacf:
> All the "standard" repositories always publish from master and release branches themselves, but from tags only from mender-qa. I think we should keep that principle here too, so that the tag publishing happens when triggered from mender-qa (which is not working ATM, but we have plans to fix).

No need to discuss, I think you're right, and I was not thinking about the whole picture here. I'll change it into a `PUBLISH_` variable instead, which we can set from the mender-qa pipeline.

.gitlab-ci.yml
lluiscampos: You convinced my by aligning this idea to what we _can_ do with `meta-mender`. So let's keep it!

With great power comes great responsibility. We shall do these publishes with great care.

.gitlab-ci.yml
kacf: I removed it, but replaced it with an expression forbidding `pr_*` references. This has no practical effect, since you can't publish from those anyway (missing keys), but is for convenience: If you forget about it, you may spend two hours waiting for tests to pass, only to find out you can't publish afterwards. With the that expression it will be immediately obvious from the pipeline graph.

.gitlab-ci.yml
```diff
@@ -233,21 +242,25 @@ publish:s3:
...
     -   aws s3api put-object-acl --acl public-read --bucket $S3_BUCKET_NAME
           --key ${RASPBIAN_NAME}/arm/${PUBLISH_NAME}
     - done
 
-  only:
-    - /^(master|[0-9]+\.[0-9]+\.x)$/
+  rules:
+    - if: '$CI_COMMIT_REF_NAME =~ /^[0-9]+\.[0-9]+\.[0-9]+(-build[0-9]+)?$/ && $PUBLISH_MENDER_CONVERT == "true"'
+    - if: '$CI_COMMIT_REF_NAME =~ /^(master|[0-9]+\.[0-9]+\.x)$/'
+    - if: '$CI_COMMIT_REF_NAME !~ /^pr_[0-9]+$/ && $MENDER_CONVERT_PUBLISH_VERSION != "" && $PUBLISH_MENDER_CONVERT == "true"'
```
kacf, line 266: How about this? I'm a little bit unsure about the second line. It is what we want in terms of behavior, but it's also kind of unintuitive considering `PUBLISH_MENDER_CONVERT` is false.

.gitlab-ci.yml
```diff
@@ -13,6 +13,16 @@ variables:
...
 
+  # Which version of mender-convert to use in published image name. Normally
+  # empty, in which case the branch or tag name will be used, but this variable
+  # is available so that it's possible to make small build fixes on an already
+  # tagged version.
+  MENDER_CONVERT_PUBLISH_VERSION: ""
+
+  # Whether to publish images. Note that images from branches are always
+  # published, even if this is false.
+  PUBLISH_MENDER_CONVERT: "false"
```
kacf, line 24: Btw, I realize this should be `PUBLISH_MENDER_CONVERT_IMAGES`. I'll change it, no need to comment on it.

lluiscampos (approved):

kacf:
@lluiscampos: I also added https://github.com/mendersoftware/mender-convert/pull/364/commits/a3bcfd2a086d00e7fca8b56d2dad6fece2e0a3dc, since my changes don't affect anything but publishing, and this is occasionally useful when long running tests are not necessary.

PR #366, opened 2021-07-23
Add checksumming to artifacts from temporary S3 bucket.

kacf:
The temporary S3 bucket is not perfectly secure, because credentials
can be obtained by outsiders by submitting a malicious pull request,
and then later they can be used to manipulate objects while a
privileged pipeline is running (during a release). Fix this by
submitting a checksum file using the standard Gitlab artifact
mechanism, and check that what we uploaded in one job, is what we get
in the next one.

Changelog: None

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -107,10 +107,18 @@ test:check-shell-formatting:
       -c configs/images/raspberrypi_raspbian_config
       -c versions_override_config
 
-    # Upload to temporary S3 bucket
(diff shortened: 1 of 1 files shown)


kacf:
@lluiscampos: This one has been bugging me for a while. Finally this part of the pipeline can be secure!

lluiscampos (approved):
Excellent :+1:

kacf:
Cherry-picked to 2.4.x and 2.5.x as well.

PR #367, opened 2021-07-23
Add checksumming to artifacts from temporary S3 bucket.

kacf:
The temporary S3 bucket is not perfectly secure, because credentials
can be obtained by outsiders by submitting a malicious pull request,
and then later they can be used to manipulate objects while a
privileged pipeline is running (during a release). Fix this by
submitting a checksum file using the standard Gitlab artifact
mechanism, and check that what we uploaded in one job, is what we get
in the next one.

Changelog: None

(cherry picked from commit 9d18a8717d8812dba135c3e2231e786afeb1ac6c)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -134,10 +134,18 @@ test:format:
       -c configs/images/raspberrypi_raspbian_config
       -c versions_override_config
 
-    # Upload to temporary S3 bucket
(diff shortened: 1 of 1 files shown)


kacf:
The failure is unrelated. As it involves a not quite trivial submodule update, I will not fix it now.

PR #374, opened 2021-08-16
Linkbot now only handles Ubuntu LTS images

oleorhagen:


scripts/linkbot/monitor-ub-server.js
```diff
@@ -26,12 +26,14 @@ JSDOM.fromURL(url, {}).then(dom => {
     var refs = document.getElementsByTagName("a");
     var matches = Array.from(refs)
         .filter(ref => ref.textContent.match(reg))
+        .filter(ref => parseInt(ref.textContent.match(reg).groups.release) % 2 == 0) // Only LTS
+        .filter(ref => !ref.textContent.match(reg).groups.minor) // Ignore the minor number
```
kacf, line 30: Is this enough? You also need to check for "04", no? They have a "10" release which is not LTS.

scripts/test/run-tests.sh
```diff
@@ -29,7 +29,7 @@ RASPBIAN_IMAGE_URL="http://downloads.raspberrypi.org/raspbian_lite/images/raspbi
 UBUNTU_IMAGE_URL="https://downloads.mender.io/mender-convert/images/Ubuntu-Bionic-x86-64.img.gz"
 
 ## Auto-update
-UBUNTU_SERVER_RPI_IMAGE_URL="http://cdimage.ubuntu.com/ubuntu/releases/20.04.2/release/ubuntu-20.04.2-preinstalled-server-armhf+raspi.img.xz"
+UBUNTU_SERVER_RPI_IMAGE_URL="http://cdimage.ubuntu.com/ubuntu/releases/20.04/release/ubuntu-20.04.2-preinstalled-server-armhf+raspi.img.xz"
```
kacf, line 32: I probably don't understand the bot, but won't this be overwritten by the next update? Or does it use the existing link structure as input?

scripts/linkbot/monitor-ub-server.js
oleorhagen: It's done here:
https://github.com/mendersoftware/mender-convert/blob/f075b8f137c39697389468c28dfcf5dcdc210cb5/scripts/linkbot/monitor-ub-server.js#L8

scripts/test/run-tests.sh
oleorhagen: Not wholly sure what you mean by next update, but if you mean the next run of the bot, then yes, in reality it overwrites it every-time, but this is not really shown, since in this case, the link is the same.
I guess we could add a check for this, but it's not really necessary.

scripts/linkbot/monitor-ub-server.js
kacf: Aha, thanks! Strictly speaking you could then match this in the same regex with `(?<release>[0-9][02468])`, but I leave it to you to decide whether that's tidier or not.

scripts/test/run-tests.sh
kacf: Ah, so this link is actually generated?

kacf (approved):

scripts/test/run-tests.sh
oleorhagen: Yes

scripts/linkbot/monitor-ub-server.js
oleorhagen: Good point! Giving it a spin :)

oleorhagen: Working fine ;)

PR #375, opened 2021-08-19
Upgrade to grub-mender-grubenv tools with new names. 

kacf:
Changelog: None

### DIFF
--- a/configs/mender_grub_config
+++ b/configs/mender_grub_config
@@ -17,7 +17,7 @@ GRUB_VERSION=2.04
 MENDER_GRUB_KERNEL_BOOT_ARGS=""
 
 # grub-mender-grubenv is the Mender integration for the GRUB bootloader
-MENDER_GRUBENV_VERSION="aa7e8f8c76c6aca6dca1820aaa42dc2cbf9762a1"
+MENDER_GRUBENV_VERSION="1a7db967495bbe8be53b7a69dcb42822f39d9a74"
 MENDER_GRUBENV_URL="${MENDER_GITHUB_ORG}/grub-mender-grubenv/archive/${MENDER_GRUBENV_VERSION}.tar.gz"
 
 # Name of the storage device containing root filesystem partitions in GRUB


scripts/test/run-tests.sh
```diff
@@ -44,26 +44,55 @@ get_pytest_files
...
+      shift
+      ;;
+    --config)
+      usage_if_empty "$2"
+      EXTRA_CONFIG="$EXTRA_CONFIG --config $2"
+      shift
+      ;;
+  esac
+  shift
+done
```
lluiscampos, line 78: Much better. Thanks! :pray:

lluiscampos (approved):

modules/deb.sh
```diff
@@ -47,7 +47,8 @@ function deb_from_repo_dist_get()  {
     fi
 
     local -r filename=$(basename $deb_package_path)
-    run_and_log_cmd "wget -Nq ${repo_url}/${deb_package_path} -P ${download_dir}"
+    local -r deb_package_url=$(echo ${repo_url}/${deb_package_path} | sed 's/+/%2B/g')
+    run_and_log_cmd "wget -Nq ${deb_package_url} -P ${download_dir}"
```
lluiscampos, line 51: Fantastic! Nice catch.

This bug has been hidden for a while because our pre-converted image did not install any-more the client (hence the `versions_override_config` is ignored, this by the way could be cleaned up) and the rest of the conversions were using default (latest).

This commit requires a changelog, though (it fixes installing "master" packages) and a cherry-pick to release branches.

lluiscampos (changes requested):
Requesting changes for the changelog request, see below.

modules/deb.sh
kacf: I think you are right. But thinking about it, there is something weird going on. Why did this succeed for Raspberry Pi (production), but fail for the others (only tests)?

kacf: Ah, wait, it's because the package is not installed, of course. Then never mind, I'm fixing.

lluiscampos (approved):

PR #380, opened 2021-09-13
MEN-5057: configs/images/raspberrypi_raspbian_config: Allow APT Suite change

lluiscampos:
As per today, our pre-converted image ships with Debian buster, while
Raspberry Pi OS APT repository has moved already to Debian bullseye,
moving buster from "stable" to "oldstable".

This configuration allows apt to update the information without
interaction from the user. This is the default behaviour in newer
versions of apt, see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=931566

### DIFF
--- a/configs/images/raspberrypi_raspbian_config
+++ b/configs/images/raspberrypi_raspbian_config
@@ -49,3 +49,18 @@ EOF
 }
 
 PLATFORM_MODIFY_HOOKS+=(grow_data_partition)
+
+#
+# Allow APT Suite change in the release info
+#
+# Upstream repository http://raspbian.raspberrypi.org/raspbian has changed buster
+# release suite from "stable" to "oldstable". The following configuration option
+# allows APT to update without user interaction.
+# This is the default un newer versions of apt, see:
+#   https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=931566
+#
+function apt_allow_release_change_suite() {
+    run_and_log_cmd "echo 'Acquire::AllowReleaseInfoChange::Suite \"true\";' \
+                     | sudo tee work/rootfs/etc/apt/apt.conf.d/99release-change-suite"
+}
+PLATFORM_MODIFY_HOOKS+=(apt_allow_release_change_suite)


0lmi (approved):
LGTM. The only thing to consider additionally is [docs](https://docs.mender.io/get-started/preparation/prepare-a-raspberry-pi-device#step-1-flash-the-raspberry-pi) update with this info.

lluiscampos:
> LGTM. The only thing to consider additionally is [docs](https://docs.mender.io/get-started/preparation/prepare-a-raspberry-pi-device#step-1-flash-the-raspberry-pi) update with this info.

This is not the only change we do, see the [main config file](https://github.com/mendersoftware/mender-convert/blob/master/configs/raspberrypi_config) for example. I think that for really interested people they would go to the Release notes, right? That will be recorded as we have a Changelog tag in the commit.

configs/images/raspberrypi_raspbian_config
```diff
@@ -49,3 +49,18 @@ EOF
...
+# Allow APT Suite change in the release info
+#
+# Upstream repository http://raspbian.raspberrypi.org/raspbian has changed buster
+# release suite from "stable" to "oldstable". The following configuration option
+# allows APT to update without user interaction.
+# This is the default un newer versions of apt, see:
+#   https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=931566
+#
+function apt_allow_release_change_suite() {
+    run_and_log_cmd "echo 'Acquire::AllowReleaseInfoChange::Suite \"true\"' \
```
0lmi, line 63:
```suggestion
    run_and_log_cmd "echo 'Acquire::AllowReleaseInfoChange::Suite \"true\";' \
```

0lmi (changes requested):

configs/images/raspberrypi_raspbian_config
lluiscampos: Thanks. fixed

0lmi (approved):

oleorhagen (approved):

PR #381, opened 2021-09-14
QA-322: Align pipeline with Mender release process     

lluiscampos:
The idea is two have two publish jobs: one manual and one automatic
    controlled via env variable.
    
    The manual one should cover all special use cases that we were
    considering before.

Additionally:
* pre-converted images: remove client version from Artifact name
The client is not being installed anymore in the pre-converted images,
so this version string baked in into the Artifact name is misleading.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -2,18 +2,18 @@
 variables:
   DOCKER_REPOSITORY: mendersoftware/mender-convert
   S3_BUCKET_NAME: mender-convert-images
+
   # These variables are present elsewhere in the repository too. Make sure to
   # search for and change them too.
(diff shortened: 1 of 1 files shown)


kacf (approved):

oleorhagen (approved):

lluiscampos:
Rebasing it on top of #382 for a release build attempt :crossed_fingers:

lluiscampos:
Reworked to skip all tests conditionally. [Here](https://gitlab.com/Northern.tech/Mender/mender-convert/-/pipelines/372628828) the normal build (failed due to the pipeline not using mender master) and [here](https://gitlab.com/Northern.tech/Mender/mender-convert/-/pipelines/372630966) the one disabling the tests.

PR #390, opened 2021-09-29
pipeline: Use mendertesting common function to save/load tmp artifacts

lluiscampos:


lluiscampos:
It doesn't work, see https://github.com/mendersoftware/mendertesting/pull/158#issuecomment-930124444

lluiscampos:
Now it works :)

kacf (approved):

PR #393, opened 2021-10-06
Modify `ln` calls for data store links to support successive runs

lluiscampos:
Fixes MEN-5078

According to `ln` man page, the correct way to create a link on a
directory is with `ln [OPTION]... TARGET... DIRECTORY`.

Create first `DIRECTORY`, and then use the above form for the link. This
way a successive run of `ln ...` will not end up creating a
`.../mender/mender` kind of link.

Changelog: Title

### DIFF
--- a/mender-convert-modify
+++ b/mender-convert-modify
@@ -167,10 +167,11 @@ fi
 # the Debian package is installed later, this folder has to preexist. It is part
 # of the rootfs setup, independently of the software.
 log_info "Creating state folder in the data partition for Mender add-ons"
+run_and_log_cmd "sudo mkdir -p work/rootfs/var/lib"
 run_and_log_cmd "sudo mkdir -p work/rootfs/data/mender-configure"
-run_and_log_cmd "sudo ln -sf /data/mender-configure work/rootfs/var/lib/mender-configure"
+run_and_log_cmd "sudo ln -sf /data/mender-configure work/rootfs/var/lib"
 run_and_log_cmd "sudo mkdir -p work/rootfs/data/mender-monitor"
-run_and_log_cmd "sudo ln -sf /data/mender-monitor work/rootfs/var/lib/mender-monitor"
+run_and_log_cmd "sudo ln -sf /data/mender-monitor work/rootfs/var/lib"
 
 if [ "${MENDER_GRUB_EFI_INTEGRATION}" == "y" ]; then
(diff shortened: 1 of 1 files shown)


mender-convert-modify
```diff
@@ -167,10 +167,11 @@ fi
 # the Debian package is installed later, this folder has to preexist. It is part
 # of the rootfs setup, independently of the software.
 log_info "Creating state folder in the data partition for Mender add-ons"
+run_and_log_cmd "sudo mkdir -p work/rootfs/var/lib"
 run_and_log_cmd "sudo mkdir -p work/rootfs/data/mender-configure"
-run_and_log_cmd "sudo ln -sf /data/mender-configure work/rootfs/var/lib/mender-configure"
+run_and_log_cmd "sudo ln -sf /data/mender-configure work/rootfs/var/lib"
 run_and_log_cmd "sudo mkdir -p work/rootfs/data/mender-monitor"
-run_and_log_cmd "sudo ln -sf /data/mender-monitor work/rootfs/var/lib/mender-monitor"
+run_and_log_cmd "sudo ln -sf /data/mender-monitor work/rootfs/var/lib"
```
merlin-northern, line 174: wait so /var/lib is a link  to /data/mender-monitor?

merlin-northern: ok it it will work, provided var/lib exists

oleorhagen (approved):

mender-convert-modify
```diff
@@ -167,10 +167,11 @@ fi
 # the Debian package is installed later, this folder has to preexist. It is part
 # of the rootfs setup, independently of the software.
 log_info "Creating state folder in the data partition for Mender add-ons"
+run_and_log_cmd "sudo mkdir -p work/rootfs/var/lib"
 run_and_log_cmd "sudo mkdir -p work/rootfs/data/mender-configure"
-run_and_log_cmd "sudo ln -sf /data/mender-configure work/rootfs/var/lib/mender-configure"
+run_and_log_cmd "sudo ln -sf /data/mender-configure work/rootfs/var/lib"
```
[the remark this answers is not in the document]
lluiscampos: We are absolutely sure because we are creating `/var/lib` directory few lines above, and because the tests are verifying that the links exist in the resulting images.

I personally like more the syntax with `.../var/lib`.

PR #417, opened 2022-01-03
Add beaglebone emmc support

mkonnov-mobica:
The 'flasher' is done in style of BBB scripts, with one significant
difference - it's not running as a different init callee, but as an additinoal systemd
unit nice'd up to 20. This may be not very correct, but at the time I
have actively fiddled with it, I found it the only working way.
However, this method showed itself as a reliable enough and I never
experienced a problem during the image flashing. Again, I could be
wrong in my solution.

- configs/beaglebone_black_debian_emmc_flasher_config:
  added the procedure to copy the custom grub.cfg to support the load
  from eMMC. It was not possible to implement as an overlay rootfs.

- rootfs_overlay_beaglebone_emmc/:
  An overlay that should be applied to the standalone eMMC image
  (artifact)

- rootfs_overlay_beaglebone_emmc_flasher/:
  An overlay that should be applied to the eMMC flasher SD card image

This way, the generation of standalone eMMC image would look like:
MENDER_ARTIFACT_NAME=test_artifact_emmc ./docker-mender-convert --disk-image input/bone-debian-10.9-console-armhf-2021-05-28-1gb.img --config configs/beaglebone_black_debian_emmc_config --overlay rootfs_overlay_mender --overlay rootfs_overlay_beaglebone_emmc

And the same for flasher SD card image:
MENDER_ARTIFACT_NAME=test_artifact_flasher ./docker-mender-convert --disk-image input/bone-debian-10.9-console-armhf-2021-05-28-1gb.img --config configs/beaglebone_black_debian_emmc_flasher_config --overlay rootfs_overlay_mender --overlay rootfs_overlay_beaglebone_emmc_flasher

Signed-off-by: Mark Konnov <mark.konnov@mobica.com>

### DIFF
--- a/configs/beaglebone_black_debian_emmc_flasher_config
+++ b/configs/beaglebone_black_debian_emmc_flasher_config
@@ -1,0 +1,8 @@
+source configs/beaglebone_black_base_config
+MENDER_STORAGE_TOTAL_SIZE_MB="3648"
+MENDER_DEVICE_TYPE="beaglebone-emmc"
+MENDER_STORAGE_DEVICE_BASE="/dev/mmcblk0p"
+
+function platform_modify() {
+  run_and_log_cmd "cp files_beaglebone/beaglebone_grub_emmc_flasher.cfg work/boot/EFI/BOOT/grub.cfg"  
+}
--- a/configs/raspberrypi4_ubuntu_config
+++ b/configs/raspberrypi4_ubuntu_config
@@ -1,7 +1,4 @@
-# Binaries generated with the following script:
-#
-#    https://github.com/drewmoseley/mender-convert-integration-scripts/blob/master/build-uboot-rpi.shraspberrypi-integration-scripts.tar.gz
-RASPBERRYPI_BINARIES="raspberrypi4-integration-2019.01.tar.gz"
+RASPBERRYPI_CONFIG="raspberrypi4"
 RASPBERRYPI_KERNEL_IMAGE="vmlinuz"
 MENDER_KERNEL_IMAGETYPE="zImage"
 MENDER_DEVICE_TYPE="raspberrypi4"
--- a/files_beaglebone/beaglebone_grub_emmc.cfg
+++ b/files_beaglebone/beaglebone_grub_emmc.cfg
@@ -1,0 +1,187 @@
+# Start of ---------- 00_mender_grubenv_defines_grub.cfg ----------
+mender_rootfsa_part=2
+mender_rootfsb_part=3
+mender_grub_storage_device=hd0
+kernel_imagetype=kernel
+initrd_imagetype=initrd
+mender_kernel_root_base=/dev/mmcblk1p
+# End of ---------- 00_mender_grubenv_defines_grub.cfg ----------
+# Start of ---------- 01_console_bootargs_grub.cfg ----------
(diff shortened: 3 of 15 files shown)


kacf:
Thanks for the contribution, @mkonnov-mobica!

I'm wondering if this isn't a bit too specific to your use case though. Why don't you keep the files in a separate repository and create a topic under the [Debian family](https://hub.mender.io/c/board-integrations/debian-family/11) on Mender Hub, which describes why and how to use it?

mkonnov-mobica:
Hi @kacf 
> I'm wondering if this isn't a bit too specific to your use case though

I don't think it's too specific. I'll describe it below. And we already dicussed the problem caused by this [here](https://hub.mender.io/t/unable-to-perform-the-ota-update-on-beaglebone-green-emmc/3329/29).

> Why don't you keep the files in a separate repository and create a topic under the Debian family on Mender Hub, which describes why and how to use it?

Good point. When I started to work on this problem, decoupling my custom files from mender-convert was my primary idea.
But after having some troubles with the config files located externally, I was needed to refuse from my idea and fork the upstream repository. It's already hard to remember what specific troubles I faced, most probably it was the problems with the build context with dockerized build.

I can try to move these files out of mender-convert directory and try the same with the upstream mender-convert, not forked.

However, I have something to say in a protection of this PR.

As Mender provides the configs for the some number of mainstream platforms including BBB eMMC, I've had a problem with the OTA process. This way, OTA updated _should_ work for the platform that is stated as 'supported'. But there was nothing but `beaglebone_black_debian_emmc_config` which just defines MENDER_STORAGE_TOTAL_SIZE_MB, MENDER_DEVICE_TYPE and MENDER_STORAGE_DEVICE_BASE variables. But it's not enough to convert the vanilla image and make the further OTA process successful.

kacf:
> Good point. When I started to work on this problem, decoupling my custom files from mender-convert was my primary idea. But after having some troubles with the config files located externally, I was needed to refuse from my idea and fork the upstream repository. It's already hard to remember what specific troubles I faced, most probably it was the problems with the build context with dockerized build.
> 
> I can try to move these files out of mender-convert directory and try the same with the upstream mender-convert, not forked.

Let's say you call it "my-bbb-repo", then I suspect that you need to clone my-bbb-repo inside of mender-convert for it to work. I reckon this can be part of the usage instructions for the repository.

> However, I have something to say in a protection of this PR.
> 
> As Mender provides the configs for the some number of mainstream platforms including BBB eMMC, I've had a problem with the OTA process. This way, OTA updates _should_ work for the platform that is stated as 'supported'. But there was nothing but `beaglebone_black_debian_emmc_config` which just defines MENDER_STORAGE_TOTAL_SIZE_MB, MENDER_DEVICE_TYPE and MENDER_STORAGE_DEVICE_BASE variables. But it's not enough to convert the vanilla image and make the further OTA process successful.

Beaglebone serves more as an example than a fully supported board in mender-convert (our documentation does not mention it at all). I understand this can be a bit frustrating, but we have limited capacity to maintain advanced user scripts for this platform. At least there is an upside to keeping the files separate, it means you don't have to go through the Mender review process to add changes to them.

Looks like there has been [some interest](https://hub.mender.io/t/emmc-flasher/4020/2) in this before, so you may get some engagement from the community.

PR #422, opened 2022-01-31
MEN-5410: Modify Debian packages installation for new repositories

lluiscampos:
Changelog: Download and install Debian packages taking into account the
    target OS. Now downloads.mender.io serves four distributions: the two
    latests releases for Debian and Ubuntu. Probe /etc/os-release to figure
    out the correct package to install, and fallback to Debian Buster
    packages which was the previous default.

### DIFF
--- a/mender-convert-modify
+++ b/mender-convert-modify
@@ -102,17 +102,23 @@ mkdir -p work/deb-packages
 log_info "Installing Mender client and related files"
 
 deb_arch=$(probe_debian_arch_name)
+deb_distro=$(probe_debian_distro_name)
+deb_codename=$(probe_debian_distro_codename)
+if ! [[ "$MENDER_APT_REPO_DISTS" == *"${deb_distro}/${deb_codename}"* ]]; then
+    deb_distro="debian"
+    deb_codename="buster"
+fi
 
 if [ "${MENDER_CLIENT_INSTALL}" = "y" ]; then
 
     log_info "Installing Mender client version ${MENDER_CLIENT_VERSION}"
 
     if [ "${MENDER_CLIENT_VERSION}" = "latest" ]; then
-        deb_name=$(deb_from_repo_dist_get "work/deb-packages" ${MENDER_APT_REPO_URL} ${deb_arch} "stable" "mender-client")
+        deb_name=$(deb_from_repo_dist_get "work/deb-packages" ${MENDER_APT_REPO_URL} ${deb_arch} "${deb_distro}/${deb_codename}/stable" "mender-client")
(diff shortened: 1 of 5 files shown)


mender-convert-modify
```diff
@@ -102,17 +102,23 @@ mkdir -p work/deb-packages
 log_info "Installing Mender client and related files"
 
 deb_arch=$(probe_debian_arch_name)
+deb_distro=$(probe_debian_distro_name)
```
oleorhagen, line 105: Suggestion:
Remember this will not only be run on debian images, necessarily, even though it is the main intention.

lluiscampos:
Best reviewed by commit

mender-convert-modify
```diff
@@ -109,20 +109,24 @@ if [ -z "$deb_distro" -o -z "$deb_codename" ]; then
...
 
-if [ "${MENDER_CLIENT_INSTALL}" = "y" ]; then
-
-    log_info "Installing Mender client version ${MENDER_CLIENT_VERSION}"
-
-    if [ "${MENDER_CLIENT_VERSION}" = "latest" ]; then
-        deb_name=$(deb_from_repo_dist_get "work/deb-packages" ${MENDER_APT_REPO_URL} ${deb_arch} "${deb_distro}/${deb_codename}/stable" "mender-client")
-    elif [ "${MENDER_CLIENT_VERSION}" = "master" ]; then
-        deb_name=$(deb_from_repo_dist_get "work/deb-packages" ${MENDER_APT_REPO_URL} ${deb_arch} "${deb_distro}/${deb_codename}/experimental" "mender-client")
+function install_deb_pacakge() {
```
oleorhagen, line 112: Nice!
How about moving it to a module also?

oleorhagen (approved):
Easy peasy!

mender-convert-modify
lluiscampos: ~Yes, that was the case before also. That is why we manually unpack the contents of the packages instead of using any `dpkg` -like tools.~

You made me realize that we need to check if distro is in supported distros (`debian`, `ubuntu`), otherwise default to Debian. Thanks!

mender-convert-modify
lluiscampos: Okey. Good.

lluiscampos:
@oleorhagen The two new commits are two extra fixes I found while manually testing all combinations. It works as expected now :+1:

oleorhagen:
> @oleorhagen The two new commits are two extra fixes I found while manually testing all combinations. It works as expected now +1

Looks good :thumbsup:

lluiscampos:
@oleorhagen Yet another fix. See last commit.

PR #423, opened 2022-01-31
modify fstab instead of replacing in order to keep modifications made

sensslen:
With this PR modifications to fstab are kept and the required changes for mender to work are made.

Changelog: Keep changes made to fstab before executing mender-convert
Signed-off-by:Simon Ensslen <simon.ensslen@griesser.ch>

### DIFF
--- a/mender-convert-modify
+++ b/mender-convert-modify
@@ -356,14 +356,17 @@ fi
 log_info "Using boot partition device in fstab: $boot_part_device"
 log_info "Using data partition device in fstab: $data_part_device"
 
-sudo bash -c "cat <<- EOF > work/rootfs/etc/fstab
-# stock fstab - you probably want to override this with a machine specific one
+fstab=$(sed '/\/boot/d' work/rootfs/etc/fstab)
+fstab=$(sed '/.*\s\/\s.*/d' <<< "$fstab")
+cat <<- EOF > work/rootfs/etc/fstab
+${boot_part_device}   ${boot_part_mountpoint}          auto       ${MENDER_BOOT_PART_FSTAB_OPTS}    0  0
 /dev/root            /                    auto       ${MENDER_ROOT_PART_FSTAB_OPTS} 1  ${MENDER_ROOT_PART_FS_PASSNO}
-proc                 /proc                proc       defaults              0  0
 
-${boot_part_device}   ${boot_part_mountpoint}          auto       defaults,sync    0  0
 ${data_part_device}   /data          auto       ${MENDER_DATA_PART_FSTAB_OPTS}      0  0
-EOF"
+
+# entries kept from original fstab
+$fstab
+EOF
 
 #
 # Make sure to re-label rootfs when selinux is in enforcing mode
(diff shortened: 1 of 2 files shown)


mender-convert-modify
```diff
@@ -318,14 +318,10 @@ fi
...
 
-sudo bash -c "cat <<- EOF > work/rootfs/etc/fstab
-# stock fstab - you probably want to override this with a machine specific one
-/dev/root            /                    auto       ${MENDER_ROOT_PART_FSTAB_OPTS} 1  ${MENDER_ROOT_PART_FS_PASSNO}
-proc                 /proc                proc       defaults              0  0
-
-${boot_part_device}   ${boot_part_mountpoint}          auto       defaults,sync    0  0
-${data_part_device}   /data          auto       ${MENDER_DATA_PART_FSTAB_OPTS}      0  0
-EOF"
+sed -i "s:.*/boot.*:${boot_part_device}   ${boot_part_mountpoint}          auto       ${MENDER_ROOT_PART_FSTAB_OPTS}    0  0:" work/rootfs/etc/fstab
```
oleorhagen, line 321: Should be `MENDER_BOOT_PART_FSTAB_OPTS` I guess?

oleorhagen (commented):
I like it.

A few ideas.

Maybe it's just as easy to remove all the lines we're replacing in one swoop, and then do the regular:

```
cat <<-EOF >> /work/rootfs/etc/fstab
all our appendings...
```
Instead of line by line like you have done here? :thinking:

mender-convert-modify
sensslen: you are absolutely right - I corrected this

sensslen:
> I like it.
> 
> A few ideas.
> 
> Maybe it's just as easy to remove all the lines we're replacing in one swoop, and then do the regular:
> 
> ```
> cat <<-EOF >> /work/rootfs/etc/fstab
> all our appendings...
> ```
> 
> Instead of line by line like you have done here? 🤔

I don't really see why removing lines and then batch adding is better.... To be honerst, I am not too experienced with bash so I don't feel confident enough to refacter the code wo modify the file in exactly one batch.

oleorhagen (approved):

sensslen:
Changes squashed into one commit - what do you mean by "Changelog would be nice"?

oleorhagen:
> Changes squashed into one commit - what do you mean by "Changelog would be nice"?

We generate a changelog for every release.

What this means in practice, is that changes, such as this one, which is user-facing, should have a Changelog entry.

So, please add:

```
Changelog: The fstab file from the image being converted is now preserved across convertions, with the Mender specific additions merged with the existing fstab file, as opposed to replacing it completely. Which was the previous approach.
```

oleorhagen:
> OK, I rephrased the changelog (which by the way was already available in the commit).

Oh my :see_no_evil: - I apologise.

Anyway, looking good now :)

oleorhagen (approved):

oleorhagen:
Thank you @sensslen for your contribution :)

PR #426, opened 2022-02-02
MEN-5257 Debian 11 tests support.

merlin-northern:
ChangeLog:title
Signed-off-by: Peter Grzybowski <peter@northern.tech>

### DIFF
--- a/scripts/test/run-tests.sh
+++ b/scripts/test/run-tests.sh
@@ -28,6 +28,8 @@ RASPBIAN_IMAGE_URL="http://downloads.raspberrypi.org/raspbian_lite/images/raspbi
 
 UBUNTU_IMAGE_URL="https://downloads.mender.io/mender-convert/images/Ubuntu-Focal-x86-64.img.gz"
 
+DEBIAN_IMAGE_URL="https://downloads.mender.io/mender-convert/images/Debian-11-x86-64.img.gz"
+
 ## Auto-update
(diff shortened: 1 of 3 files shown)


scripts/test/run-tests.sh
```diff
@@ -160,5 +162,29 @@ else
...
+    echo >&2 "Running the uncompressed test"
+    echo >&2 "----------------------------------------"
+    rm -rf deploy
+    gunzip --force "input/Debian-11-x86-64.img.gz"
+    run_convert "release-2" \
+                "input/Debian-11-x86-64.img" \
+                "--config configs/debian-qemux86-64_config $EXTRA_CONFIG" || test_result=$?
+    ret=0
+    test -f deploy/Debian-11-x86-64-qemux86_64-mender.img || ret=$?
+    assert "${ret}" "0" "Expected uncompressed file deploy/Debian-11-x86-64-qemux86_64-mender.img"
```
lluiscampos, line 184: This we don't need - it is enough with one uncompressed test for one image

merlin-northern: removed.

lluiscampos:
As commented privately, I am not confident that using `mkosi` is the right approach, although judging by the JIRA task description and estimate it seems to be the planned thing to do...

We can clear it out tomorrow with @kacf :+1:

merlin-northern:
> As commented privately, I am not confident that using `mkosi` is the right approach, although judging by the JIRA task description and estimate it seems to be the planned thing to do...
> 
> We can clear it out tomorrow with @kacf 👍

right! thank you!

kacf (approved):
Looks great!

PR #429, opened 2022-02-10
docker-mender-convert: fix mktemp invocation for macOS

TheYoctoJester:
docker-mender-convert: fix mktemp invocation for macOS

As suggested per https://hub.mender.io/t/bug-convert-a-raw-disk-image-in-macos-fails-mktemp-illegal-option-p/4428/3,
the '-p' switch is not supported on macOS and needs to be worked around.

Signed-off-by: Josef Holzmayr <josef.holzmayr@northern.tech>

### DIFF
--- a/docker-mender-convert
+++ b/docker-mender-convert
@@ -26,7 +26,7 @@ MENDER_CONVERT_DIR="$(pwd)"
 MENDER_CONVERT_VERSION=$(git_mender_convert_version)
 
 # Create a unique work directory so we can run multiple containers in parallel
-WORK_DIR=$(mktemp -d -p $(pwd) -t work.XXXX)
+WORK_DIR=$(mktemp -d 2>/dev/null || mktemp -d -t 'mytmpdir')
 WORK_SUFFIX=$(echo $WORK_DIR | awk -F. '{print $NF}')
 LOG_FILE="convert.log.${WORK_SUFFIX}"
 


docker-mender-convert
```diff
@@ -26,7 +26,7 @@ MENDER_CONVERT_DIR="$(pwd)"
 MENDER_CONVERT_VERSION=$(git_mender_convert_version)
 
 # Create a unique work directory so we can run multiple containers in parallel
-WORK_DIR=$(mktemp -d -p $(pwd) -t work.XXXX)
+WORK_DIR=$(mktemp -d 2>/dev/null || mktemp -d -t 'mytmpdir')
```
kacf, line 29: I think the directory has to be created as a subfolder of this directory, otherwise it will not be visible inside Docker. I have a couple of suggested workarounds. Either:

```
WORK_DIR=$(mktemp -d $(pwd)/work.XXXX)
```

or

```
WORK_DIR=$(env TMPDIR=$(pwd) mktemp -d work.XXXX)
```

kacf (changes requested):
First change! :partying_face: 

All our changes require changelogs. Everyone should familiarize themselves with [this section](https://github.com/mendersoftware/mender/blob/master/CONTRIBUTING.md#changelog-tags).

TheYoctoJester:
Will reorganize my branches and then re-submit.

PR #432, opened 2022-02-14
MEN-5221: Handle secure boot Grub

oleorhagen:


mender-convert-package
```diff
@@ -257,8 +257,8 @@ if [ "${partition_scheme}" == "gpt" ]; then
     run_and_log_cmd "${PARTED} -s ${img_path} mklabel gpt"
     run_and_log_cmd "${PARTED} -s ${img_path} unit s mkpart ESP fat32 ${boot_part_start} ${boot_part_end}"
     # Keep the uuid of the original gpt table for the boot partition
-    log_info "Updating GPT partition uuid of the boot partition in image: ${img_path} to ${boot_partuuid}"
     boot_partuuid=$(disk_get_part_value "${disk_image}" 1 UUID)
+    log_info "Updating GPT partition uuid of the boot partition in image: ${img_path} to ${boot_partuuid}"
```
kacf, line 261: Nitpick: Should this be in the first commit?

modules/probe.sh
```diff
@@ -344,3 +344,12 @@ is_efi_compatible_kernel() {
...
     return 0
 }
+
+# has_secureboot_shim
+#
+# $1 - the boot partition to search for a secureboot shim
+#
+# Checks the EFI/* filesystem for the presence of a signed boot shim
+has_secureboot_shim() {
+    find "${1}" -type f -name 'shim*' -print0 | grep -qz shim
```
kacf, line 354: I'd use `shim*.efi`.

modules/grub.sh
```diff
@@ -155,24 +110,23 @@ EOF
     sudo rm -rf work/boot/EFI/systemd
     sudo rm -rf work/boot/NvVars
     for empty_dir in $(
-                     cd work/boot find . -maxdepth 1 -type d -empty
+                     cd work/boot && find . -maxdepth 1 -type d -empty
```
kacf, line 113: I would consider squashing this whole commit (the refactor) with the previous one. It's a bit confusing to review the first one, since it is in fact wrong, and due to this change, it won't even work. When viewed with `-w`, the two commits together look pretty good!

kacf (approved):
After looking at the combined commits, it looks good! I'm pre-approving, but there are a couple of nitpicks below.

oleorhagen:
All fixed - merging!

PR #433, opened 2022-02-15
Make docker container independent from mender-convert repository

sensslen:
Changelog: The mender-convert docker image does now contain the mender-convert application and does no longer need companion checkouts. With this change publishing the docker container to docker hub is possible and desirable

Signed-off-by:Simon Ensslen simon.ensslen@griesser.ch

### Description
This PR makes the mender-convert docker container a first class citizen (as it now contains the application) and should be ready to be published to docker-hub. This allows for mender-convert versions to be distinguished by the docker container as opposed to git (as is the current state).

For anybody using mender-convert in a ci environment this lends checking out mender-convert git repository unnecessary as one can then just install the correct version of the docker container.

### DIFF
--- a/docker-entrypoint.sh
+++ b/docker-entrypoint.sh
@@ -20,9 +20,9 @@ set -e
 
 cd /mender-convert
 
-echo "Running mender-convert "$@""
+echo "Running mender-convert $*"
 
 ./mender-convert "$@"
 
-# Set owner and group to same as launch directory.
-[ -d deploy ] && chown -R --reference=. deploy
+# Set owner and group to same as input directory.
+[ -d deploy ] && chown -R --reference=input deploy
--- a/docker-mender-convert
+++ b/docker-mender-convert
@@ -16,38 +16,35 @@
 
 set -e
 
-. modules/git.sh
+IMAGE_NAME=${IMAGE_NAME:-"mendersoftware/mender-convert"}
+DEPLOY_DIRECTORY=${DEPLOY_DIRECTORY:-"$(pwd)/deploy"}
+INPUT_DIRECTORY=${INPUT_DIRECTORY:-"$(pwd)/input"}
+LOGS_DIRECTORY=${LOGS_DIRECTORY:-"$(pwd)/logs"}
 
-IMAGE_NAME=${IMAGE_NAME:-mender-convert}
+mkdir -p "$DEPLOY_DIRECTORY"
+mkdir -p "$LOGS_DIRECTORY"
 
-MENDER_CONVERT_DIR="$(pwd)"
+# Create a unique log file so we can run multiple containers in parallel
+LOG_FILE="convert.log.$(date +%s)-$$"
 
-#Pass in version in-case we are added as a git submodule
-MENDER_CONVERT_VERSION=$(git_mender_convert_version)
-
-# Create a unique work directory so we can run multiple containers in parallel
-WORK_DIR=$(mktemp -d $(pwd)/work.XXXX)
-WORK_SUFFIX=$(echo $WORK_DIR | awk -F. '{print $NF}')
-LOG_FILE="convert.log.${WORK_SUFFIX}"
+echo "using log file at: ${LOGS_DIRECTORY}/${LOG_FILE}"
 
 set +e
 docker run \
   --rm \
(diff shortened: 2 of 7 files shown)


sensslen:
Well reading your comment it sounds like there should maybe exist two docker containers. One that is used by a dev to have a stable environment to test the app and another one that is targeted towards the users of the app. I - as a user of mender-convert - do find it quite cumbersome to always checkout mender-convert repository to be able to build my app.

docker-mender-convert
```diff
@@ -16,38 +16,34 @@
...
 
 set +e
 docker run \
   --rm \
-  -v $MENDER_CONVERT_DIR:/mender-convert \
-  -v $WORK_DIR:/mender-convert/work \
+  -v "$INPUT_DIRECTORY":/mender-convert/input \
+  -v "$LOGS_DIRECTORY":/mender-convert/logs \
+  -v "$DEPLOY_DIRECTORY":/mender-convert/deploy \
+  --mount type=tmpfs,destination=/mender-convert/work \
```
kacf, line 36: `tmpfs` is not a good idea. Images can easily be bigger than all available memory, which would not work with `tmpfs`.

Is a volume even required for the work dir?

docker-entrypoint.sh
```diff
@@ -20,9 +20,9 @@ set -e
...
 
-echo "Running mender-convert "$@""
+echo "Running mender-convert $*"
 
 ./mender-convert "$@"
 
-# Set owner and group to same as launch directory.
-[ -d deploy ] && chown -R --reference=. deploy
+# Set owner and group to same as input directory.
+find deploy -mindepth 1 -maxdepth 1 -exec chown -R --reference=input {} \;
```
kacf, line 28: Why doesn't the original work?

kacf (commented):
Even though this makes it more inconvenient for developers, I think @sensslen has a point. What makes it convenient for developers, makes it *inconvenient* for users. A docker container is a more handy distribution mechanism.

But it needs more work than just this. I can think of at least these things:

* The gitlab-ci file needs to publish the image.
* The documentation [here](https://docs.mender.io/system-updates-debian-family/convert-a-mender-debian-image), and probably several other places in mender-docs, needs to be updated.
* Mender Hub posts needs to be updated as well, at least the posts that belong to Northern.tech (probably you don't have access, but we can apply a diff for you).
* It's failing for many platforms still, for reasons I have not looked into.

I'm in favor of this personally, but I don't consider it a priority for us, so it'll be up to @sensslen to pull most of the weight.

docker-mender-convert
sensslen: The reason I used a volume was because they are temporary files - from my understanding of docker you should never change the internal file system during docker run - I might be wrong. Good point though about the size of the images!

sensslen: Reading it up you're right - no need for a volume here....

sensslen:
Things do now look OK (the tests pass) and I also updated the documentation. For now I don't see how I could potentially generate a diff for the "posts" you mention. I also think that the publishing of the container should be done on your side - I don't have access to any mender accounts and stuff.....

kacf:
Following [this discussion ](https://github.com/mendersoftware/mender-docs/pull/1691#discussion_r814689988), it looks like most of the documentation changes won't be necessary. However, there are still a few things remaining:

* The name of the container needs to be changed to `mendersoftware/mender-convert` so it fits under our namespace.
* There needs to be an upload job for the container. You can take a look at [this template](https://github.com/mendersoftware/mendertesting/blob/master/.gitlab-ci-check-docker-build.yml#L92-L117) for inspiration. I can help you with testing the final version, since you won't have access to it, but would be nice to have the prototype ready first.
* Since this is a breaking change because of the input folder, there should be a changelog entry which says that all input files must now be placed in the `input` folder and not anywhere else (just add a second changelog entry in addition to the existing one).

sensslen:
I just added a prototype for publishing the docker image - as said: testing is not really possible for me - At least there seem to be no syntax issues in the gitlab ci file. I also added the other changes requested (also on the docs PR)

kacf:
Thanks @sensslen, I have fixed some merge conflicts and am re-testing it [here](https://github.com/mendersoftware/mender-convert/pull/440).

PR #434, opened 2022-02-25
README: clarify requisites and compatibility

TheYoctoJester:

### Description

Thank you!

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,11 @@ For more detailed information about `mender-convert` please visit the
 
 ### Prepare image and configuration
 
+*Only 32bit-Images are supported at the moment!*
+
+The following steps give a quick start using Raspbian. For a more detailed guide,
+especially concerning version compatibilities, please visit [the the corresponding thread](https://hub.mender.io/t/raspberry-pi-3-model-b-b-raspbian/140) on the Mender Hub.
+
 Download the raw Raspberry Pi disk image into a subdirectory input:
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -21,6 +21,19 @@ on your device.
 
 ![Mender logo](https://github.com/mendersoftware/mender/raw/master/mender_logo.png)
 
+## Requisites
+
+`mender-convert` is supported on the following development platform(s):
+
+	- Ubuntu 18.04 x86, 64bit
+
+Other platforms may work, but are not under active testing. Patches to add additional
+platforms or fix compatibility issues are welcome.
+
+The storage used during the process, specifically the work directories (usually below the clone of
+`mender-convert`) and the output directory should be located on local, Unix-style filesystem.
+Using network storage or emulated filesystems might cause permission or owner issues.
+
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -34,6 +47,11 @@ For more detailed information about `mender-convert` please visit the
 
 ### Prepare image and configuration
 
+*Only 32bit-Images are supported at the moment!*
+
+The following steps give a quick start using Raspbian. For a more detailed guide,
+especially concerning version compatibilities, please visit [the the corresponding thread](https://hub.mender.io/t/raspberry-pi-3-model-b-b-raspbian/140) on the Mender Hub.
```
kacf, line 53: Suggestion: This section is a bit confusing (it already was before you edited it). It makes the assumption that you are using Raspberry Pi, but nothing before this section indicates this. Should we perhaps put this under its own Raspberry Pi header?

README.md
```diff
@@ -21,6 +21,19 @@ on your device.
...
+`mender-convert` is supported on the following development platform(s):
+
+	- Ubuntu 18.04 x86, 64bit
+
+Other platforms may work, but are not under active testing. Patches to add additional
+platforms or fix compatibility issues are welcome.
+
+The storage used during the process, specifically the work directories (usually below the clone of
+`mender-convert`) and the output directory should be located on local, Unix-style filesystem.
+Using network storage or emulated filesystems might cause permission or owner issues.
```
kacf, line 35: :+1:

README.md
TheYoctoJester: I agree that it is suboptimal. My main concern is that we're seeing a split of source of truth here. The current README suggests that the various threads on the Hub are valid information on supported boards, which I find problematic. I wanted to get something out quickly to avoid further confusion in the mean time - but my personal favortie would be to list the actively supported boards (via included configurations) in the README, and refer to the Hub only as "additional" information exchange.

README.md
```diff
@@ -31,9 +44,41 @@ For more detailed information about `mender-convert` please visit the
...
+## Included configurations
+
+### Core configurations
+
+These configurations are officially supported.
+
+| Configuration | Supported OS / hardware |
+| :------------ | :---------------------- |
+| beaglebone_black_debian_emmc_config | BeagleBone Black 4GB, Debian 32bit on internal eMMC storage |
+| beaglebone_black_debian_sdcard_config | BeagleBone Black, Debian 32bit on external SD card storage |
```
kacf, line 56: I would put Beaglebone under contributed. It's true that we added it, and we are testing it during releases, but it's only because of self-interest: We want to verify that the UEFI code path for ARM works on at least one physical board. We have not successfully converted Raspberry Pi to UEFI yet, but it's looking increasingly likely we will do this soon, and then we will likely kill all efforts for Beaglebone. Hence better to avoid the term supported.

kacf (changes requested):

kacf (approved):
Perfect, thanks!

PR #438, opened 2022-03-14
MEN-5219: Grub.d integration

kacf:


scripts/test/generate-image.sh
```diff
@@ -0,0 +1,169 @@
...
+            -display vnc=:23 \
+            -m 512 \
+            || ret=$?
+        [ $ret -eq 0 ] && break
+    done
+
+    return $ret
+}
+
+eval $GENERATE_VARIANT
```
oleorhagen, line 169: Q:
Is eval even needed here?

modules/grub.sh
```diff
@@ -55,10 +60,83 @@ EOF
...
+        cd work/grub-mender-grubenv-${MENDER_GRUBENV_VERSION}
+        run_and_log_cmd "make 2>&1"
+        run_and_log_cmd "sudo make DESTDIR=$PWD/../ BOOT_DIR=boot install-boot-env"
+        run_and_log_cmd "sudo make DESTDIR=$PWD/../rootfs install-grub.d-boot-scripts"
+        run_and_log_cmd "sudo make DESTDIR=$PWD/../rootfs install-tools"
+        # We need this for running the scripts once.
+        run_and_log_cmd "sudo make DESTDIR=$PWD/../rootfs install-offline-files"
     )
 
+    # Mender-convert usually runs in a container. It's difficult to launch
```
oleorhagen, line 97: War story addition

modules/grub.sh
```diff
@@ -173,7 +173,7 @@ function grub_install_mender_grub() {
     sudo rm -rf work/boot/EFI/systemd
     sudo rm -rf work/boot/NvVars
     for empty_dir in $(
-                     cd work/boot && find . -maxdepth 1 -type d -empty
+                     cd work/boot && find . -maxdepth 1 -type d -empty -not -name .
```
oleorhagen, line 176: Q:
! empty ?

scripts/test/run-tests.sh
```diff
@@ -31,7 +31,7 @@ UBUNTU_IMAGE_URL="https://downloads.mender.io/mender-convert/images/Ubuntu-Focal
 DEBIAN_IMAGE_URL="https://downloads.mender.io/mender-convert/images/Debian-11-x86-64.img.gz"
 
 ## Auto-update
-UBUNTU_SERVER_RPI_IMAGE_URL="http://cdimage.ubuntu.com/ubuntu/releases/20.04/release/ubuntu-20.04.3-preinstalled-server-armhf+raspi.img.xz"
+UBUNTU_SERVER_RPI_IMAGE_URL="http://cdimage.ubuntu.com/ubuntu/releases/20.04/release/ubuntu-20.04.4-preinstalled-server-armhf+raspi.img.xz"
```
oleorhagen, line 34: This the bot should have handled :(

Have you looked at it?

oleorhagen (approved):
Real nice!

Hard to take it in in this timeline, but from the commit messages, comments, and structure, I do get the general gist of it.

Some serious :ninja: hacks with `chroot` for the grub-file generation...

Mostly nitpicks and a few questions..

Nits most often encountered:
`local -r` (`-r` missing)
`echo >&2`

For all intents and purposes production ready I think :tada: :rocket: :1st_place_medal:

scripts/test/run-tests.sh
kacf: I think it was released today. The bot will not look at it until next week, or?

I did look, there was nothing from the bot today.

kacf:
> Some serious ninja hacks with `chroot` for the grub-file generation...

Yeah, to be honest I don't really like it. GRUB should be better prepared to handle the offline case. You can mock quite a lot of the stuff with `grub-probe` like I did, which helps a lot, but you still have to run under `chroot`, and it's not streamlined at all.

scripts/test/generate-image.sh
```diff
@@ -0,0 +1,169 @@
...
+    done
+    losetup -d $LO_DEVICE
+    rmdir tmp-p1
+    rmdir tmp-p2
+}
+
+generate_debian() {
+    local image="Debian-11-x86-64.img"
+
+    mkosi --root-size=2G --distribution=debian --release=bullseye --format=gpt_ext4 --bootable --checksum \
```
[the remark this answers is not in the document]
kacf: At the moment I don't really see us supporting more than one, which is already more than enough work. If we start supporting more than one it would make sense to make this configurable.

scripts/test/generate-image.sh
```diff
@@ -0,0 +1,169 @@
...
+# possible when offline. This is something which is easier with systemd-boot, so
+# longterm GRUB will probably follow, or systemd-boot will take over. Anyway,
+# let's do it by using a systemd service to perform the job, and then shut down.
+create_grub_regeneration_service() {
+    local boot="$1"
+    local root="$2"
+
+    cat > "$root/etc/systemd/system/mender-regenerate-grub-and-shutdown.service" <<EOF
+[Unit]
+Description=Regenerate grub scripts, disable itself and then shut down.
```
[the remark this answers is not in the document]
kacf: Yeah. In retrospect, it may have been better to use the `chroot` approach here as well, but I did this long before I got to the `mender-convert` part of it. It's quite a lot of trial and error to change it now, and it's only for internal use, so I think I'll leave this for now.

scripts/test/generate-image.sh
```diff
@@ -0,0 +1,169 @@
...
+        echo "Generating GRUB boot files live..."
+        qemu-system-x86_64 \
+            $maybe_kvm \
+            -drive file="$image",if=ide,format=raw \
+            -drive file=/usr/share/OVMF/OVMF_CODE.fd,if=pflash,format=raw,unit=0,readonly=on \
+            -drive file="$nvvars",if=pflash,format=raw,unit=1 \
+            -display vnc=:23 \
+            -m 512 \
+            || ret=$?
+        [ $ret -eq 0 ] && break
```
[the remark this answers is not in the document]
kacf: No reason. But does it make it clearer? I was thinking that splitting it up was more readable... But you are the reader! What do you say? 🙂

modules/grub.sh
kacf: Not sure what the question is... You mean use `\! -empty` operator? That's not correct, because we do want empty directories, just not `.` if there's nothing at all.

scripts/test/generate-image.sh
```diff
@@ -0,0 +1,169 @@
...
+generate_debian() {
+    local image="Debian-11-x86-64.img"
+
+    mkosi --root-size=2G --distribution=debian --release=bullseye --format=gpt_ext4 --bootable --checksum \
+          --password password --package=openssh-server,dhcpcd5 --package grub-efi-amd64-signed \
+          --package shim-signed --package lsb-release --output="$image" build
+
+    post_process_image "$image"
+
+    echo "Image successfully generated!"
```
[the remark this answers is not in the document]
kacf: Fixed, and all the others too.

modules/grub.sh
oleorhagen: Discussed locally

scripts/test/run-tests.sh
oleorhagen: Hmm... Let's keep this on our minds. I"ll have a quick look now

scripts/test/generate-image.sh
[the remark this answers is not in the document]
oleorhagen: Either or is fine :)

Just curious

scripts/test/generate-image.sh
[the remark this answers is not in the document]
oleorhagen: Yeh, it is fine imo

scripts/test/generate-image.sh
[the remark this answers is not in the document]
oleorhagen: I was thinking more along the lines, at what point will we remember to bump this?

PR #441, opened 2022-03-18
Prevent rsync from changing permissions of already existing files and folders

sensslen:
Changelog: Use the permissions of the target folder when applying an overlay
Signed-off-by: Simon Ensslen <simon.ensslen@griesser.ch>

### Description
When applying an overlay previously the ownership and the permissions of the overlay were applied to the target. This could lead to changing the ownership of the root folder. With this PR the permissions of the target folder are used.

### DIFF
--- a/mender-convert-modify
+++ b/mender-convert-modify
@@ -44,7 +44,14 @@ function trap_term() {
 trap trap_term INT TERM
 trap trap_exit EXIT
 
-echo "Running $(basename $0): $@"
+function join_array_by {
+  local d=${1-} f=${2-}
+  if shift 2; then
+    printf %s "$f" "${@/#/$d}"
+  fi
+}
+
+echo "Running $(basename "$0"): $*"
 
 source modules/bootstrap.sh
 source modules/disk.sh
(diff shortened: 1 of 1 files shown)


mender-convert-modify
```diff
@@ -280,7 +280,7 @@ done
 
 for overlay in "${overlays[@]}"; do
     log_info "Applying rootfs overlay: ${overlay}"
-    run_and_log_cmd "sudo rsync --archive --keep-dirlinks --verbose ${overlay}/ work/rootfs/"
+    run_and_log_cmd "sudo rsync --archive --keep-dirlinks --no-super --verbose ${overlay}/ work/rootfs/"
```
kacf, line 283: While I completely agree that the current method where one has to be root is quite annoying, I'm wondering if this will work correctly. Won't this prevent the `work/rootfs` folder from containing any root-owned files? The resulting filesystem will then have the wrong file owners.

sensslen: You're right - the documentation sais something like that. My tests led to all files copied inheriting the rights of the containing folder.

kacf: I just tested this, and it won't work correctly. Say you have an overlay which defines a home directory with directories owned by non-root users. Then it should appear like this:

```
drwxr-xr-x 3    0    0 4096 Mar 24 09:13 /home/
drwxr-xr-x 2 1000 1000 4096 Mar 24 09:13 /home/user
```

But with your change, it will appear like this:

```
drwxr-xr-x 3    0    0 4096 Mar 24 09:13 /home/
drwxr-xr-x 2    0    0 4096 Mar 24 09:13 /home/user
```

sensslen: I agree that that is not correct - neither is the way it works today. Do you have any ideas?

kacf: Unfortunately I don't. I've been thinking that maybe we can do something with fakeroot, but I don't think it works together with Docker.

An alternative that the user can do is to set up the owners correctly using a hook instead of an overlay (or in addition to an overlay). That would execute inside Docker, where you are already root and can do whatever you want. But it is not automatic, of course.

sensslen: Just thinking: how about mounting overlays with overlayfs?

kacf: Doesn't overlayfs have the same problem? I don't think it's possible to manipulate owners without being root, even on overlayfs.

sensslen:
As noted above, this might be an issue but there is no real solution, as one might well change permissions in an overlay.

PR #445, opened 2022-04-20
Handle xz images in monitor-raspbian

oleorhagen:
They've changed their image extensions again

Changelog: None

### DIFF
--- a/scripts/linkbot/monitor-raspbian-os.js
+++ b/scripts/linkbot/monitor-raspbian-os.js
@@ -35,13 +35,13 @@ function getNewRaspbian(target, url, separator) {
         var document = dom.window.document;
         var refs = document.getElementsByTagName("a");
         var test = Array.from(refs)
-            .filter((ref) => ref.textContent.match("raspios-.*-lite.*.zip$"))
+            .filter((ref) => ref.textContent.match("raspios-.*-lite.*\.img\.(zip|xz)$"))
             .reduce((acc, element) => {
-                acc.push(element.textContent.match("raspios-.*-lite.*.zip$").input);
+                acc.push(element.textContent.match("raspios-.*-lite.*\.img\.(zip|xz)$").input);
                 return acc;
             }, [])[0];
         console.log(`Test var: ${test}`);
-        const imageName = test.substring(0, test.length - 4);
+        const imageName = test.substring(0, test.lastIndexOf('.img'));
         if (separator) {
             return {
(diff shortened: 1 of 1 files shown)


scripts/linkbot/monitor-raspbian-os.js
```diff
@@ -35,13 +35,13 @@ function getNewRaspbian(target, url, separator) {
...
-            .filter((ref) => ref.textContent.match("raspios-.*-lite.*.zip$"))
+            .filter((ref) => ref.textContent.match("raspios-.*-lite.*.(zip|xz)$"))
             .reduce((acc, element) => {
-                acc.push(element.textContent.match("raspios-.*-lite.*.zip$").input);
+                acc.push(element.textContent.match("raspios-.*-lite.*.(zip|xz)$").input);
                 return acc;
             }, [])[0];
         console.log(`Test var: ${test}`);
-        const imageName = test.substring(0, test.length - 4);
+        const imageName = test.replace(/\.img\..*/, "");
```
mzedel, line 44:
```suggestion
        const imageName = test..substring(0, test.lastIndexOf('.img'));
```
since the regex might fail if the next time the image is provided as a plain `.img`

scripts/linkbot/monitor-raspbian-os.js
```diff
@@ -35,13 +35,13 @@ function getNewRaspbian(target, url, separator) {
         var document = dom.window.document;
         var refs = document.getElementsByTagName("a");
         var test = Array.from(refs)
-            .filter((ref) => ref.textContent.match("raspios-.*-lite.*.zip$"))
+            .filter((ref) => ref.textContent.match("raspios-.*-lite.*.(zip|xz)$"))
             .reduce((acc, element) => {
-                acc.push(element.textContent.match("raspios-.*-lite.*.zip$").input);
+                acc.push(element.textContent.match("raspios-.*-lite.*.(zip|xz)$").input);
```
mzedel, line 40: would it be possible to check for the `img` extension too and maybe exclude the files we're sure not to want (maybe `.md5` or something...)?

mzedel (approved):
this should work, but I wouldn't hold my breath on how long the regex holds...

scripts/linkbot/monitor-raspbian-os.js
oleorhagen: Uhh, smrt!

PR #454, opened 2022-06-08
fix: Upgrade to latest version of grub-mender-grubenv.

kacf:
This fixes a boot problem which was introduced in the standalone grub
boot scripts, during the grub.d integration work.

Changelog: None
Ticket: None


kacf:
Superseded by #455.

PR #455, opened 2022-06-10
Fix for mender-convert regressions.

kacf:


scripts/test/generate-image.sh
```diff
@@ -136,7 +119,7 @@ Description=Regenerate grub scripts, disable itself and then shut down.
 
 [Service]
 Type=oneshot
-ExecStart=sh -c "grub-install && update-grub && systemctl disable mender-regenerate-grub-and-shutdown.service && poweroff"
+ExecStart=sh -c "grub-install && grub-install --removable && update-grub && systemctl disable mender-regenerate-grub-and-shutdown.service && poweroff"
```
oleorhagen, line 122: Hah, never thought of running systemctl from a service file before

oleorhagen (approved):

scripts/test/generate-image.sh
kacf: I'm sure Lennart would be proud! (or not...)

PR #464, opened 2022-07-01
fix: added support for kernel8.img on newer raspberry pi oses

oldgiova:
Signed-off-by: Roberto Giovanardi <rob.giovanardi@gmail.com>

the form below:

```
fix: added support for kernel8.img on newer raspberry pi oses

Newer Raspberry Pi OSes uses the new kernel8.img instead the kernel8l.img, so `mender-convert` can't build up the converted image.

Changelog: added support for kernel8.img on newer Raspberry Pi OSes
Ticket: [MEN-5708](https://tracker.mender.io/browse/MEN-5708)
```

### Description

Newer Raspberry Pi OSes uses the new kernel8.img instead the kernel8l.img, so `mender-convert` can't build up the converted image. This modification would match both either old and new Raspberry Pi OS builds.


kacf:
Looks like this didn't work in practice (it detects more than one file, and thus fails the copy). Either the kernel installation section will have to be improved, or alternatively, the raspberrypi4 config file could be split into a "legacy" file and a new file. For usability I would vote for the former approach, but depends on how easy it is...

oldgiova:
> Looks like this didn't work in practice (it detects more than one file, and thus fails the copy). Either the kernel installation section will have to be improved, or alternatively, the raspberrypi4 config file could be split into a "legacy" file and a new file. For usability I would vote for the former approach, but depends on how easy it is...

Right, let me drop this PR for a better and more robust solution

PR #468, opened 2022-08-18
feat: Install a bootstrap Artifact along with the client

oleorhagen:
Changelog: Title
Ticket: MEN-5594

### DIFF
--- a/mender-convert-modify
+++ b/mender-convert-modify
@@ -204,11 +204,6 @@ fi
 run_and_log_cmd "echo 'device_type=${device_type}' > work/device_type"
 run_and_log_cmd "sudo install -m 0444 work/device_type work/rootfs/data/mender/"
 
-if [ "${MENDER_CLIENT_INSTALL}" = "y" ]; then
-    run_and_log_cmd "echo 'artifact_name=${MENDER_ARTIFACT_NAME}' \
-                     | sudo tee work/rootfs/etc/mender/artifact_info"
-fi
-
 log_info "Creating state scripts version file."
 case "${MENDER_CLIENT_VERSION}" in
     1*)  VERSION_STRING='2' ;;
--- a/mender-convert-package
+++ b/mender-convert-package
@@ -39,6 +39,7 @@ echo "Running $(basename $0): $@"
 source modules/bootstrap.sh
 source modules/disk.sh
 source modules/testscfg.sh
+source modules/versioncmp.sh
 
 # The mender_convert_config is always used and provides all the defaults
 declare -a configs=("configs/mender_convert_config")
@@ -203,6 +204,15 @@ fi
 # when writing the bootstrap Artifact we have to move the datadir out of the
 # way, and write the rootfs image first.
 run_and_log_cmd "mv --force work/rootfs/data work/data"
+
+if [[ "${MENDER_CLIENT_INSTALL}" == y ]]; then
+    # If the client version is less than 3.4, then log a warning
+    if ! miminum_required_version "3.4.0" "${MENDER_CLIENT_VERSION}"; then
+        log_warn "You are now installing a Mender client which does not support the " \
+                "bootstrap Artifact present in the image. This means that updates will not " \
+                "work, and your image is impossible to update"
+    fi
+fi
 
 disk_create_file_system_from_folder "work/rootfs/" "work/rootfs.img" \
     "${rootfs_image_sectors}" "${image_fs_type}"
--- a/mender-convert-package
+++ b/mender-convert-package
@@ -199,17 +199,32 @@ else
     log_fatal "Could not determine root file-system type. Aborting..."
 fi
 
-disk_create_file_system_from_folder "work/rootfs/data/" "work/data.img" \
-    "${data_part_sectors}" "${image_fs_type}"
-
-# Clear this area as it will be contained in the data.img
-sudo rm -rf work/rootfs/data/*
+# NOTE: Since the data partition has a dependency on the rootfs.img checksum
+# when writing the bootstrap Artifact we have to move the datadir out of the
+# way, and write the rootfs image first.
+run_and_log_cmd "mv --force work/rootfs/data work/data"
 
 disk_create_file_system_from_folder "work/rootfs/" "work/rootfs.img" \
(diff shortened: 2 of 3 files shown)


lluiscampos (commented):
This does exactly what the acceptance criteria of the task says, but @kacf why we don't _provide_ rootfs checksum in the artifact for `mender-convert` images?

mender-convert-modify
```diff
@@ -205,8 +205,11 @@ run_and_log_cmd "echo 'device_type=${device_type}' > work/device_type"
 run_and_log_cmd "sudo install -m 0444 work/device_type work/rootfs/data/mender/"
 
 if [ "${MENDER_CLIENT_INSTALL}" = "y" ]; then
-    run_and_log_cmd "echo 'artifact_name=${MENDER_ARTIFACT_NAME}' \
-                     | sudo tee work/rootfs/etc/mender/artifact_info"
+    log_info "Installing the bootstrap Artifact"
+    run_and_log_cmd "mender-artifact write bootstrap-artifact \
+                        --artifact-name "${MENDER_ARTIFACT_NAME}" \
+                        --device-type  "${device_type}" \
+                        --output-path work/rootfs/data/mender/bootstrap.mender"
```
kacf, line 212: AFAIK, the bootstrap artifact doesn't provide any data of its own. I believe this was done so that users who want to deploy special bootstrap artifacts can start with a clean slate. So you will need to add this in the form of `-p rootfs-image.version:NAME` and `-p rootfs-image.checksum:abc...`. Check what a normal rootfs-image artifact has, just to be sure.

kacf (changes requested):

mender-convert-modify
lluiscampos: That answers my question too.

oleorhagen:
Feel free to have another lookover now.

One thing to note since last time, is that the bootstrap Artifact is installed unconditionally now, regardless of whether or not we install the client.

kacf (approved):
One question: Should we mention client version requirements in the changelog? I suspect that mender client 3.3 won't like that the `artifact_info` file is gone. Maybe something like:

"It is recommended to run Mender client 3.4 or later with this version of mender-convert."

mender-convert-package
```diff
@@ -199,8 +199,10 @@ else
...
 
-disk_create_file_system_from_folder "work/rootfs/data/" "work/data.img" \
-    "${data_part_sectors}" "${image_fs_type}"
+# NOTE: Since the data partition has a dependency on the rootfs.img checksum
+# when writing the bootstrap Artifact we have to move the datadir out of the
+# way, and write the rootfs image first.
+run_and_log_cmd "mv --force work/rootfs/data work/data"
 
 # Clear this area as it will be contained in the data.img
 sudo rm -rf work/rootfs/data/*
```
lluiscampos, line 208: Then we can remove these lines, right? I think we should to avoid confusion later on.

lluiscampos (approved):

lluiscampos:
> One question: Should we mention client version requirements in the changelog? I suspect that mender client 3.3 won't like that the `artifact_info` file is gone. Maybe something like:
> 
> "It is recommended to run Mender client 3.4 or later with this version of mender-convert."

Yes, I think we should. But it is not a recommendation, rather a hard requirement: older clients will not work at all without the file (they won't be able to check for updates).

kacf:
True, but the artifact will work, just not the full image. Maybe that should be specified explicitly.

oleorhagen:
> > One question: Should we mention client version requirements in the changelog? I suspect that mender client 3.3 won't like that the `artifact_info` file is gone. Maybe something like:
> > "It is recommended to run Mender client 3.4 or later with this version of mender-convert."
> 
> Yes, I think we should. But it is not a recommendation, rather a hard requirement: older clients will not work at all without the file (they won't be able to check for updates).

@lluiscampos I updated the Changelog.

Not also the commit https://github.com/mendersoftware/mender-convert/pull/468/commits/645884ca94ffaa5948e058296be823a66411438b which adds a check for the client version installed, and logs a warning (up for debate on what to do here) when the client is installed, and at an unusable version.

If we are to do like a `log_fatal`, then maybe we have to think about what to do with our versioning. Kristian thinks a major bump is not needed in any case.

mender-convert-package
```diff
@@ -204,8 +205,14 @@ fi
 # way, and write the rootfs image first.
 run_and_log_cmd "mv --force work/rootfs/data work/data"
 
-# Clear this area as it will be contained in the data.img
-sudo rm -rf work/rootfs/data/*
+if [[ "${MENDER_CLIENT_INSTALL}" == y ]]; then
+    # If the client version is less than 3.4, then log a warning
+    if ! miminum_required_version "3.4.0" "${MENDER_CLIENT_VERSION}"; then
```
lluiscampos, line 210: This will also evaluate to false with `master` or `latest`, right? In both cases the warning should not appear

(for `latest` as per today it should appear but by the time we release this code then it should not - so to keep it simple I think we can ignore already)

mender-convert-package
lluiscampos: nit pick: the removal of these lines should have been in the first commit

lluiscampos:
> If we are to do like a `log_fatal`, then maybe we have to think about what to do with our versioning. Kristian thinks a major bump is not needed in any case.

I think it is okey.

mender-convert-package
oleorhagen: haha, I knew you were gonna comment on this :p 

See https://github.com/mendersoftware/mender-convert/pull/468/files#diff-76ffda63fe1086e8e1289ee58c126e145c92c0bd8a5165f138eb4deb65055674R20 ;)

oleorhagen: The latest part is a good point. But don't you think it's fine like this, until we release?

lluiscampos: > haha, I knew you were gonna comment on this :p
> 

Sorry that I didn't read the comment :see_no_evil: 



> The latest part is a good point. But don't you think it's fine like this, until we release?

Yes, that is what I meant - it is okey

lluiscampos (approved):

PR #472, opened 2022-09-06
ci: Install mender-artifact tool from deb package

lluiscampos:


.gitlab-ci.yml
```diff
@@ -137,23 +137,36 @@ convert_raspbian_raspberrypi4:
   variables:
     RASPBERRYPI_PLATFORM: raspberrypi4
 
+.bash_helpers: &bash_helpers |
```
oleorhagen, line 140: Nice!

.gitlab-ci.yml
```diff
@@ -163,9 +176,19 @@ convert_raspbian_raspberrypi4:
...
-    - mv mender-artifact /usr/bin/
+    - if is_final_tag "${MENDER_ARTIFACT_VERSION}" || is_build_tag "${MENDER_ARTIFACT_VERSION}"; then
+      # Direct download from the pool
+    -   wget "https://downloads.mender.io/repos/debian/pool/main/m/mender-artifact/mender-artifact_${MENDER_ARTIFACT_VERSION}-1%2Bubuntu%2Bjammy_amd64.deb"
+    -   dpkg --install mender-artifact_${MENDER_ARTIFACT_VERSION}-1+ubuntu+jammy_amd64.deb
+    -   rm mender-artifact_${MENDER_ARTIFACT_VERSION}-1+ubuntu+jammy_amd64.deb
+    - else
+      # Assume master and install from experimental repo
+    -   curl -fLsS https://downloads.mender.io/repos/debian/gpg | apt-key add -
+    -   echo "deb [arch=$(dpkg --print-architecture)] https://downloads.mender.io/repos/debian ubuntu/jammy/experimental main" >
```
oleorhagen, line 187: I guess we could get `jammy` from the `release file`? But then again...

DAYP

oleorhagen (approved):
:zap:

.gitlab-ci.yml
```diff
@@ -137,23 +137,36 @@ convert_raspbian_raspberrypi4:
...
+    [[ "${version}" =~ ^[0-9]+\.[0-9]+\.[0-9]+-build[0-9]+$ ]] && return 0 || return $?
+  }
+
 .template_test_acceptance: &test_acceptance
   stage: test_acceptance
   needs:
     - job: build
       artifacts: true
-  image: ubuntu:22.04
+  image: ubuntu:jammy
```
kacf, line 157: * I would appreciate a comment that `jammy == 22.04`. I don't know all those names by heart! :older_man:

* Since `jammy` is repeated quite a lot below, could we make it a variable instead?

kacf (approved):
Pre-approving, but see below.

.gitlab-ci.yml
lluiscampos: I think making it a variable (and adding a comment) is a good idea, I'll follow up :+1:

The reason why I changed to the codename is that it used below, so that whenever updating it here you would see the references below. However having a variable makes it even easier!

.gitlab-ci.yml
lluiscampos: Which `release file` are you talking about? I used the instructions from docs.mender.io :thinking:

oleorhagen: It's nothing really. I was just thinking we could dynamically get the ubuntu version, instead of hard-coding it. But it's probs not worth it. nevermind

lluiscampos: That is something I also thought about for the mender-docs, where now we have 5 tabs to choose distro. If any, let's fix it there and eventually propagate to our CI :)

scripts/test/run-tests.sh
```diff
@@ -43,7 +43,7 @@ UBUNTU_IMAGE_URL="https://downloads.mender.io/mender-convert/images/Ubuntu-Focal
 DEBIAN_IMAGE_URL="https://downloads.mender.io/mender-convert/images/Debian-11-x86-64.img.gz"
 
 ## Auto-update
-UBUNTU_SERVER_RPI_IMAGE_URL="http://cdimage.ubuntu.com/ubuntu/releases/20.04/release/ubuntu-20.04.4-preinstalled-server-armhf+raspi.img.xz"
+UBUNTU_SERVER_RPI_IMAGE_URL="http://cdimage.ubuntu.com/ubuntu/releases/20.04/release/ubuntu-20.04.5-preinstalled-server-armhf+raspi.img.xz"
```
oleorhagen, line 46: Should we upgrade to `https` ? :thinking:

oleorhagen (approved):

PR #473, opened 2022-09-07
Fix/image updater

mzedel:
fix + also slight refactoring...

### DIFF
--- a/scripts/linkbot/index.js
+++ b/scripts/linkbot/index.js
@@ -1,0 +1,31 @@
+import { updateURLLink } from './common.js';
+import * as bbMonitor from './monitor-bb.js';
+import * as bbSdMonitor from './monitor-bb-sd.js';
+import * as rpiMonitor from './monitor-rpi.js';
+import * as ubuntuServerMonitor from './monitor-ub-server.js';
+import * as raspbianOsMonitor from './monitor-raspbian-os.js';
+
+const monitors = {
+  bbMonitor,
+  bbSdMonitor,
+  rpiMonitor,
+  ubuntuServerMonitor,
+  raspbianOsMonitor
+};
+
+Object.entries(monitors).map(async ([monitorName, { checkForUpdates, determineCurrentState, target, updateReference }]) => {
+  console.log(`${monitorName}: starting`);
+  // Read the input file, and parse the variable input
+  const state = await determineCurrentState();
+  const result = await checkForUpdates(state);
+  if (!result) {
+    console.log(`${monitorName}: no updates found`);
+    return;
+  }
+  console.log(`${monitorName}: We've got a new release! \\o/`);
+  if (updateReference) {
+    await updateReference(result);
+    return;
+  }
+  await updateURLLink({ ...result, target });
+});
--- a/scripts/linkbot/monitor-bb-sd.js
+++ b/scripts/linkbot/monitor-bb-sd.js
@@ -1,49 +1,23 @@
-const jsdom = require("jsdom");
-const { JSDOM } = jsdom;
-const fs = require('fs')
-const { updateURLLink } = require('./common');
+import { getCurrentTestData, getLinksByMatch } from './common.js';
 
-const target = "BBB_DEBIAN_SDCARD_IMAGE_URL"
+const reg = 'bone-debian-(?<version>[0-9]+.[0-9]+)-iot-armhf-(?<date>[0-9]{4}-[0-9]{2}-[0-9]{1,2})-4gb.img.xz$';
+export const target = 'BBB_DEBIAN_SDCARD_IMAGE_URL';
 
-var reg = "bone-debian-(?<version>[0-9]+\.[0-9]+)-iot-armhf-(?<date>[0-9]{4}-[0-9]{2}-[0-9]{1,2})-4gb.img.xz$"
-
-// Read the input file, and parse the variable input
-try {
-    const data = fs.readFileSync('../test/run-tests.sh', 'utf8')
-          .split('\n')
-          .filter(line => line.match(`${target}=.*`))
-    var line = data[0]
-    var m = line.match(".*=\"(?<url>[a-zA-Z-://\._]*)(?<imageName>bone-debian-(?<version>[0-9]+\.[0-9]+)-iot-armhf-(?<date>[0-9]{4}-[0-9]{2}-[0-9]{1,2})-4gb.img.xz)")
-    var url = m.groups.url
-    var currentImageName = m.groups.imageName
-} catch (err) {
-    console.error(err)
-    process.exit(1)
-}
-
-JSDOM.fromURL(url, {}).then(dom => {
-    var document = dom.window.document;
-    var table = document.getElementById("list");
-    var rows = table.rows;
-    var matches = Array.from(rows)
-        .filter(row => row.firstChild.textContent.match(reg))
-        .reduce((acc, element) => {
-            var regMatch = element.firstChild.textContent.match(reg)
-            acc.push({
-                text: element.firstChild.textContent,
-                version: regMatch.groups.version,
-                date: regMatch.groups.date,
-            })
-            return acc
-        }, [])
-        .sort((a,b) => {
-            // The bone-debian setup has two parts which needs comparing:
-            // * The release-version: i.e., 10.3
-            // * The date: i.e., 2020-04-06
-            return parseFloat(b.version) - parseFloat(a.version) || Date.parse(b.date) - Date.parse(a.date)
-        })
-    if (matches[0].text !== currentImageName) {
-        console.error("We've got a new release! \\o/");
-        updateURLLink(`${target}=\"${url}/${matches[0].text}\"`)
+export const checkForUpdates = ({ url, imageName }) =>
+  getLinksByMatch(url, reg).then(async (links) => {
+    // The bone-debian setup has two parts which needs comparing:
+    // * The release-version: i.e., 10.3
+    // * The date: i.e., 2020-04-06
+    const matches = links.sort((a, b) => parseFloat(b.match.groups.version) - parseFloat(a.match.groups.version) || Date.parse(b.date) - Date.parse(a.date));
+    const { element, link } = matches[0];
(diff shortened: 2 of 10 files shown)


lluiscampos (commented):
I cannot review this :see_no_evil:

oleorhagen (changes requested):
Trash

oleorhagen:
I'm disappointed

scripts/linkbot/monitor-bb-sd.js
```diff
@@ -1,49 +1,29 @@
-const jsdom = require("jsdom");
-const { JSDOM } = jsdom;
-const fs = require('fs')
-const { updateURLLink } = require('./common');
+const { getCurrentTestData, getLinksByMatch, updateURLLink } = require('./common');
 
-const target = "BBB_DEBIAN_SDCARD_IMAGE_URL"
+const target = 'BBB_DEBIAN_SDCARD_IMAGE_URL';
+const reg = 'bone-debian-(?<version>[0-9]+.[0-9]+)-iot-armhf-(?<date>[0-9]{4}-[0-9]{2}-[0-9]{1,2})-4gb.img.xz$';
```
oleorhagen, line 4: Q: Why change to single `'`?

scripts/linkbot/monitor-bb-sd.js
```diff
@@ -1,49 +1,29 @@
...
 
-var reg = "bone-debian-(?<version>[0-9]+\.[0-9]+)-iot-armhf-(?<date>[0-9]{4}-[0-9]{2}-[0-9]{1,2})-4gb.img.xz$"
+const checkForUpdates = async (url, imageName) =>
+  getLinksByMatch(url, reg).then(async (links) => {
+    // The bone-debian setup has two parts which needs comparing:
```
oleorhagen, line 8: Thank you from future us :smiley_cat:

scripts/linkbot/monitor-bb-sd.js
```diff
@@ -1,49 +1,29 @@
...
-    var currentImageName = m.groups.imageName
+  const { imageName, url } = getCurrentTestData(
+    target,
+    '.*="(?<url>[a-zA-Z-://._]*)(?<imageName>bone-debian-(?<version>[0-9]+.[0-9]+)-iot-armhf-(?<date>[0-9]{4}-[0-9]{2}-[0-9]{1,2})-4gb.img.xz)'
+  );
+  checkForUpdates(url, imageName);
 } catch (err) {
-    console.error(err)
-    process.exit(1)
+  console.error(err);
```
oleorhagen, line 27: Oof (shame on me)

scripts/linkbot/monitor-bb-sd.js
```diff
@@ -1,49 +1,29 @@
...
-const fs = require('fs')
-const { updateURLLink } = require('./common');
+const { getCurrentTestData, getLinksByMatch, updateURLLink } = require('./common');
 
-const target = "BBB_DEBIAN_SDCARD_IMAGE_URL"
+const target = 'BBB_DEBIAN_SDCARD_IMAGE_URL';
+const reg = 'bone-debian-(?<version>[0-9]+.[0-9]+)-iot-armhf-(?<date>[0-9]{4}-[0-9]{2}-[0-9]{1,2})-4gb.img.xz$';
 
-var reg = "bone-debian-(?<version>[0-9]+\.[0-9]+)-iot-armhf-(?<date>[0-9]{4}-[0-9]{2}-[0-9]{1,2})-4gb.img.xz$"
+const checkForUpdates = async (url, imageName) =>
```
oleorhagen, line 6: Nit: Technically, it now does more than `check` for updates, right?

scripts/linkbot/monitor-bb.js
```diff
@@ -1,58 +1,25 @@
...
-                acc.push(element.textContent.match("bone-debian.*\.img\.xz$").input)
-                return acc
-            }, [])[0]
-        return `${target}=\"${url}/${test}\"`;
-    });
+  console.error(err);
+  process.exit(1);
 }
-
-JSDOM.fromURL(url, {}).then(async dom => {
```
oleorhagen, line 37: Oh my....

scripts/linkbot/monitor-raspbian-os.js
```diff
@@ -1,67 +1,18 @@
...
-    });
-}
-
-
-function updateGitlabCILink(newLine, newURL, newName) {
+const updateGitlabCILink = ({ name: targetName, url: targetUrl }, newName, newUrl) => {
   try {
     const data = fs
-      .readFileSync("../../.gitlab-ci.yml", "utf8")
+      .readFileSync('../../.gitlab-ci.yml', 'utf8')
```
oleorhagen, line 10: I guess, with a little bit of sweat, we could unify this with the regular `run.sh` updater?

scripts/linkbot/monitor-raspbian-os.js
```diff
@@ -70,46 +21,40 @@ function updateGitlabCILink(newLine, newURL, newName) {
...
-    let bl = Date.parse(b.groups.date);
-    if (al == bl) {
-      let ad = Date.parse(a.groups.updated);
-      let bd = Date.parse(b.groups.updated);
-      return bd - ad;
+};
+
+const checkForUpdates = async (url, currentImageName) =>
+  getLinksByMatch(url, reg).then(async (links) => {
+    const { link, match } = links.sort((a, b) => Date.parse(b.match.groups.date) - Date.parse(a.match.groups.date))[0];
```
oleorhagen, line 28: tbh, I was hoping you were gonna show me some magic here.

I remember when I wrote this, that it felt like some abstraction was escaping me here, and there'd be som elegant way of forging this, considering it's pretty much similar in a lot of the updaters.

scripts/linkbot/monitor-raspbian-os.js
```diff
@@ -70,46 +21,40 @@ function updateGitlabCILink(newLine, newURL, newName) {
...
-      "RASPBIAN_URL",
-      `${url}${matches[0].input.split(" ")[0].split("/").slice(0, -1)}`,
-      (separator = ":")
-    ).then(({ replacementLine, imageName }) => {
-      console.log(`New release: ${replacementLine}`);
-      updateGitlabCILink(replacementLine, "RASPBIAN_URL", imageName);
-    });
-  }
-});
+// Read the input file, and parse the variable input
```
oleorhagen, line 46: Q: If the above is `updateCILink`. Should this be `GetCILink`?

oleorhagen (approved):
Real nice, mine freund!

oleorhagen:
> Trash

Alright, serious mode on:

Think it looks real nice! I knew my last stuff _was actually_ trash. But I'd still forgotten exactly how much :spaghetti: it was! _Oh.. my.. lord..._ Thanks for doing this @mzedel 

I also have a few extra ideas:

When I've been pondering rewriting this, I was imagining something along the lines of:

Having the monitors export each their own

 * fetcher
 * matcher
 * updater

And then simply do the same mechanics on all of them in the common file. i.e., fetch the links -> match -> filter, then update.

What do you think?

Finally

Thanks for doing this!

scripts/linkbot/monitor-bb-sd.js
mzedel: that's just my local `prettier` installation, makes all my JS code _prettier_ without me worrying about a thing 😄

scripts/linkbot/monitor-bb-sd.js
mzedel: absolutely right! I'll clean this up...

scripts/linkbot/monitor-bb-sd.js
oleorhagen: G.E.N.I.U.S

scripts/linkbot/monitor-raspbian-os.js
mzedel: hmm I would imagine a useable approach would be to collect all the variables in a `.env` file and use this from both locations when needed... OTOH it might be easier to just make the read/write files part more reusable...

scripts/linkbot/monitor-raspbian-os.js
mzedel: the date based comparisons are similar in places, but still dependent on the regex groups or where the date is gotten from, so still somewhat unique each monitor... IMO if the versioning scheme was the same for all the images that would be a more useful abstraction, but here we compare dates in one place and ubuntu version strings in another and then again something more semver like... that is to say: I don't see a better abstraction either unfortunately..

scripts/linkbot/monitor-raspbian-os.js
mzedel: getting this in line too!

scripts/linkbot/monitor-raspbian-os.js
oleorhagen: :'(

scripts/linkbot/monitor-raspbian-os.js
oleorhagen: Hmm, could be an idea

mzedel:
as an additional note: while there are 3 image updates pending, I left the related files untouched to check if the workflow does what it is expected to do...

oleorhagen:
> as an additional note: while there are 3 image updates pending, I left the related files untouched to check if the workflow does what it is expected to do...

Agreed :smile:

scripts/linkbot/index.js
```diff
@@ -0,0 +1,31 @@
+import { updateURLLink } from './common.js';
+import * as bbMonitor from './monitor-bb.js';
+import * as bbSdMonitor from './monitor-bb-sd.js';
+import * as rpiMonitor from './monitor-rpi.js';
+import * as ubuntuServerMonitor from './monitor-ub-server.js';
+import * as raspbianOsMonitor from './monitor-raspbian-os.js';
+
+const monitors = {
```
oleorhagen, line 8: Yaaz :raised_hands:

oleorhagen (approved):
very, very, very nicely done, sir!

I'm _impressed_!

PR #496, opened 2022-10-19
3.0.x: MEN-5944, MEN-5954 and MEN-5955 + misc cherry-picks

lluiscampos:


kacf:
Given the quite intrusive nature of this fix, and the fact that the new RPi image is technically speaking a "new platform", I would consider not cherry-picking this.

lluiscampos:
> Given the quite intrusive nature of this fix, and the fact that the new RPi image is technically speaking a "new platform", I would consider not cherry-picking this.

@kacf

I don't agree that it is a "new platform". It fixes the integration on newer images for a supported platform. It is a bugfix because latest released mender-artifact (and all released versions for all that matters) do not correctly integrate with newer images of Raspbian OS.

The level of "intrusiveness" can also be argued. It is on an isolated function, only run for one specific platform, for which I have now thoroughly tested. So I think it is very low risk of regressions.

In any case we need to release this. We discussed it long in a post-SU meeting and we decided to not block 3.4.0 release but do a bugfix release as soon as it was solved. So basically we have two options: we merge into `mender-convert/3.0.x` and do the bugfix release as planned or branch out from `mender-convert/master` to `3.1.x` and put that into the 3.4.1 release. I prefer the former case.

kacf:
> I don't agree that it is a "new platform". It fixes the integration on newer images for a supported platform.

Maybe "new platform" was a bad expression. What I meant was a "new major version", like how Ubuntu 22 would be a new version of the supported platform Ubuntu. In general adding support for such platforms is not expected in patch releases unless the changes are pretty minor.

But I trust your judgment, I brought it up more as a possible re-evaluation than as disagreement.

kacf (approved):

lluiscampos:
> In general adding support for such platforms is not expected in patch releases unless the changes are pretty minor.
> 

You are right and I support the guideline. This time is an exception given the circumstances.

> But I trust your judgment, I brought it up more as a possible re-evaluation than as disagreement.

Thanks for bringing it up anyway. Good to double-check these decisions :)

PR #498, opened 2022-10-26
fix: Fix grub.d integration on architecture cross-build

lluiscampos:
Changelog: Fix grub.d integration when host and target 32/64 bit architectures don't match. Namely: pass the correct `--target` parameter to `grub-install` command instead of relaying on host architecture.

Ticket: MEN-5979

### DIFF
--- a/modules/probe.sh
+++ b/modules/probe.sh
@@ -36,8 +36,10 @@ probe_arch() {
     fi
 
     target_arch="unknown"
-    if grep -q x86-64 <<< "${file_info}"; then
+    if grep -Eq "ELF 64-bit.*x86-64" <<< "${file_info}"; then
         target_arch="x86_64"
+    elif grep -Eq "ELF 32-bit.*386" <<< "${file_info}"; then
+        target_arch="i386"
     elif grep -Eq "ELF 32-bit.*ARM" <<< "${file_info}"; then
         target_arch="arm"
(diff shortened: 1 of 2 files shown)


modules/probe.sh
```diff
@@ -38,6 +38,8 @@ probe_arch() {
     target_arch="unknown"
     if grep -q x86-64 <<< "${file_info}"; then
         target_arch="x86_64"
+    elif grep -q "Intel 80386" <<< "${file_info}"; then
```
lluiscampos, line 41: I chose this string based on my experiments:

```
$file -b /bin/nano
ELF 32-bit LSB pie executable, Intel 80386, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux.so.2, BuildID[sha1]=a007931407fcfdcd7d640f588e53fe50a944a095, for GNU/Linux 3.2.0, stripped
```

And that is equivalent to the `x86_64` above.

But we could add something more strict like `ELF 32-bit.*386` (and for consistency `ELF 64-bit.*x86_64` above).

tranchitella: I'd vote for the strict version, testing `ELF 32-bit.*` and `ELF 64-bit.*`. However, I don't consider it a blocker.

tranchitella (approved):

kacf (approved):
Excellent!

lluiscampos:
* 3.3.x

PR #500, opened 2022-10-31
style(linkbot): Add a prettier config

oleorhagen:
Simply copied the prettier configuration from Mantra


mzedel (approved):
Thank you for that!

PR #526, opened 2023-01-03
fix: Update to grub-mender-grubenv with lock fix.

kacf:
Changelog: grub-mender-grubenv: Add lock file to prevent data races
when accessing boot environment concurrently.

Ticket: None

Signed-off-by: Kristian Amlie <kristian.amlie@northern.tech>

PR #529, opened 2023-01-11
fix: Update to grub-mender-grubenv with lock fix.

kacf:
Changelog: grub-mender-grubenv: Add lock file to prevent data races when accessing boot environment concurrently.

Changelog: grub-mender-grubenv: Fail execution if both environments are corrupt.

Changelog: grub-mender-grubenv: Recompute checksum after restoring environment (new versions of grub-editenv add extra data).

Changelog: Add the XFS module to the grub efi binary module list

Ticket: None

(cherry picked from commit c0fee00b635e026bf08d4e1583fcfd8a5b5aaad1)


lluiscampos:
@kacf Good commit message! However, why this PR has one more changelog entry than #526?

kacf:
The grub-mender-grubenv SHA was a bit more behind than in master, so it included several fixes. Perhaps I should have changed the title to `fix: Update grub-mender-grubenv to get several fixes.`, that's on me. But the changelog is correct.

PR #542, opened 2023-02-09
MEN-6301: fix: Broken version check

lluiscampos:


oleorhagen (approved):
Hah! 

Does this not get caught in anyway whilst running this... :unamused: 

Good catch though

lluiscampos:
I pushed another fix: `log_warn` does not support multi-line:
image

PR #549, opened 2023-02-24
QA-554: Update to Ubuntu 22.04 Jammy and fix test.

kacf:
```
commit 0623c4cd1877a23e18c3c5157575dfd422a0d024
Date:   Fri Feb 24 14:34:03 2023

    test: Ignore some grub.cfg differences in `00_header` section.
    
    Ubuntu 22.04 Jammy can trigger differences in searching for the root
    device when running live as opposed to mender-convert's jailroot.
    Since we have grub scripts which later set root back to the value it
    should be, it is safe to ignore these differences. However, since it
    is dangerous to ignore these differences *outside* `00_header`, make
    the sed script fairly specific so that it is only ignored inside this
    section.
    
    Ticket: QA-554
    
commit c4bd09a03a675b8ac56ba960e8cd74d74dd81f34
Author: Lluis Campos <lluis.campos@northern.tech>
Date:   Thu Feb 16 09:14:48 2023

    chore: Update Ubuntu test image to Jammy distribution
    
    Ticket: QA-554
    
```

### DIFF
--- a/tests/test_grub_integration.py
+++ b/tests/test_grub_integration.py
@@ -152,12 +152,22 @@ class TestGrubIntegration:
             connection.run("rm -f /data/new-grub-efi-modified.cfg")
 
         # Another few differences we work around in the main grub files:
+        #
         # * `--hint` parameters are not generated in offline copy.
+        #
         # * `root` variable is not set in offline copy.
+        #
         # * `fwsetup` is added somewhat randomly depending on availability both
         #   on build host and device.
+        #
         # * locale, lang and gettext settings and module may or may not be
         #   present depending on test host.
+        #
+        # * Inside the `00_header` section, ignore the `search` functions which
+        #   set the root variable. This is safe because we have a later script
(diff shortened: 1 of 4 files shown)


lluiscampos (approved):
I don't fully follow the `sed` script, but the reasoning (and the test results) look good to me :+1: Thank you!

tests/test_grub_integration.py
```diff
@@ -185,6 +224,33 @@ def test_offline_and_runtime_boot_scripts_identical(self, connection):
...
+  };
+  \%END /etc/grub.d/00_header% ! {
+    p;
+    n;
+    b header_loop;
+  };
+};
+p;
+' """
+                "/data/old-grub-modified.cfg /data/new-grub-modified.cfg"
```
lluiscampos, line 252:
```suggestion
                "/data/old-grub-mender-grubenv-modified.cfg /data/new-grub-mender-grubenv-modified.cfg"
```

kacf: Oops, thanks!

PR #585, opened 2023-08-16
fix: Fix convert error when using part-UUIDs and grub.d together. 

kacf:
The new grub-mender-grubenv requires us to pass in the
`mender_boot_uuid` variable as well.

Changelog: Fix convert error when using `MENDER_GRUB_D_INTEGRATION`
and `MENDER_ENABLE_PARTUUID` together.

Ticket: MEN-6645


modules/grub.sh
```diff
@@ -31,11 +31,14 @@ EOF
...
         rootfsa_partuuid=$(disk_get_partuuid_from_device "${root_part_a_device}")
         rootfsb_partuuid=$(disk_get_partuuid_from_device "${root_part_b_device}")
+        log_info "Using boot partition partuuid in grubenv: $boot_partuuid"
         log_info "Using root partition A partuuid in grubenv: $rootfsa_partuuid"
         log_info "Using root partition B partuuid in grubenv: $rootfsb_partuuid"
         cat <<- EOF >> work/grub-mender-grubenv-${MENDER_GRUBENV_VERSION}/mender_grubenv_defines
+mender_boot_uuid=${boot_partuuid}
 mender_rootfsa_uuid=${rootfsa_partuuid}
 mender_rootfsb_uuid=${rootfsb_partuuid}
 EOF
```
lluiscampos, line 44: For coherence, I would now move the two inconditional lines above:
```
mender_rootfsa_part=${MENDER_ROOTFS_PART_A_NUMBER}
mender_rootfsb_part=${MENDER_ROOTFS_PART_B_NUMBER}
```
to inside the `else` that follows, something like:
```
    else
        cat <<- EOF >> work/grub-mender-grubenv-${MENDER_GRUBENV_VERSION}/mender_grubenv_defines
mender_kernel_root_base=${MENDER_STORAGE_DEVICE_BASE}
mender_rootfsa_part=${MENDER_ROOTFS_PART_A_NUMBER}
mender_rootfsb_part=${MENDER_ROOTFS_PART_B_NUMBER}
EOF
```

configs/mender_grub_config
```diff
@@ -26,7 +26,7 @@ GRUB_VERSION=2.04
 MENDER_GRUB_KERNEL_BOOT_ARGS=""
 
 # grub-mender-grubenv is the Mender integration for the GRUB bootloader
-MENDER_GRUBENV_VERSION="a525df022c10950ddbadad05adaefe7b2451ec62"
+MENDER_GRUBENV_VERSION="ca17bb8574f0c4a9560bc11eeba93479ed8b03fc"
```
lluiscampos, line 29: Nitpick: use here the merge commit (once we merge the other PR).

lluiscampos (approved):
Looks good otherwise :+1:

oleorhagen (approved):

oleorhagen (approved):

modules/grub.sh
kacf: The `mender_rootfsa_part` and `mender_rootfsb_part` are still needed. This is somewhat historical and definitely not optimal, but we still rely on simple one or two digit partition indexes when interacting with the bootloader from the Mender client. For example, if `mender_rootfsa_part` and `mender_rootfsb_part` are set to 2 and 3, then we know that those indexes refer to `mender_rootfsa_uuid` and `mender_rootfsb_uuid`, respectively. The key variable, `mender_boot_part` is only set to the indexes, not the partition UUIDs, so we need this mapping at boot time to know which partition UUID to select.

lluiscampos (approved):
It looks safe to me.

We can remove `git` dependency altogether from the Dockerfile [here](https://github.com/mendersoftware/mender-convert/blob/0bb7124769e42e23b6a8b1c5edee0edb70235f96/Dockerfile#L40) - it says that it is used to download `mender-grub-env` but as far as I can see the code uses `wget`.

kacf:
> We can remove `git` dependency altogether from the Dockerfile [here](https://github.com/mendersoftware/mender-convert/blob/0bb7124769e42e23b6a8b1c5edee0edb70235f96/Dockerfile#L40) - it says that it is used to download `mender-grub-env` but as far as I can see the code uses `wget`.

Good thinking, done!

PR #588, opened 2023-08-30
Remove data contents before calculating rootfs size

LorenzoRuffati:

### Description

In the current execution flow the rootfs size is estimated while the contents of the "work/rootfs/data" folder are still present

In cases when the contents of the data folder are large this leads the package script to assume that there won't be enough space to write the rootfs to the final image and the execution terminates with an error

An alternative solution would be to use the "--exclude" flag in the du command

### DIFF
--- a/mender-convert-package
+++ b/mender-convert-package
@@ -162,6 +162,11 @@ else
     image_name=$DEPLOY_IMAGE_NAME
 fi
 
+# NOTE: Since the data partition has a dependency on the rootfs.img checksum
+# when writing the bootstrap Artifact we have to move the datadir out of the
+# way, and write the rootfs image first.
+run_and_log_cmd "mv --force work/rootfs/data work/data"
+
 actual_rootfs_size=$(sudo du -s --block-size=512 work/rootfs | cut -f 1)
 
 # KiB -> 512 sectors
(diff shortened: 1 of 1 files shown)


mender-convert-package
```diff
@@ -200,10 +205,6 @@ else
     log_fatal "Could not determine root file-system type. Aborting..."
 fi
 
-# NOTE: Since the data partition has a dependency on the rootfs.img checksum
-# when writing the bootstrap Artifact we have to move the datadir out of the
-# way, and write the rootfs image first.
-run_and_log_cmd "mv --force work/rootfs/data work/data"
```
kacf, line 208: Please remove the extra newline which is now present here (our shell format checker is complaining about it).

kacf (changes requested):
Thanks, looks sensible to me!

A couple of formal requirements:

* Can you add a `Changelog: Title` to the bottom of the commit message, just above the signoff? This should make the commit format checker happy.
* One issue below.

mender-convert-package
LorenzoRuffati: Does the commit to fix this also need to follow the "fix" and "Changelog" convention?

kacf: Yes. But please just amend the original commit, and force-push, don't make a new one.

kacf:
> * Can you add a `Changelog: Title` to the bottom of the commit message, just above the signoff? This should make the commit format checker happy.

Oops, sorry my bad, I forgot that `Ticket: None` is also needed. Our checker is really picky! 🙂

kacf (approved):
Thank you @LorenzoRuffati!

kacf:
Cherry-picked to 4.0.x [here](https://github.com/mendersoftware/mender-convert/pull/589), 3.0.x deals with this differently, and is soon EOL anyway, so won't cherry-pick there.

PR #592, opened 2023-09-11
feat: Add support for operating on squashfs files

LorenzoRuffati:
### Motivation
By creating the disk image directly into a squash file system [^1] and then bind mounting the image onto a file in the input directory we can save a significant amount of space, especially for larger input images

### Potential issue
partx, which is used to get partition information in mender-convert-extract generates error messages on stderr since the disk image is read only, however the correct output is generated on stdout and the commands exit with 0 exit code

### Example
I'm using a golden device with a disk size of 120GB, the rootfs partition is 60GB of which 35-40 are used. The squashfs image obtained by dd-ing the whole disk is 24GB without requiring manual alterations of the partition table and filesystems after creating the golden image

[^1]: By using dd in combination with [the command in this article](https://semjonov.de/posts/2021-05/create-squashfs-archive-from-stdin-and-sign-it-on-the-fly/)

### DIFF
--- a/mender-convert-package
+++ b/mender-convert-package
@@ -370,6 +370,9 @@ case "${MENDER_COMPRESS_DISK_IMAGE}" in
     none)
         :
         ;;
+    sqfs)
+        :
+        ;;
     *)
         log_fatal "Unknown MENDER_COMPRESS_DISK_IMAGE value: ${MENDER_COMPRESS_DISK_IMAGE}"
         ;;
--- a/modules/decompressinput.sh
+++ b/modules/decompressinput.sh
@@ -39,6 +39,9 @@ function compression_type()  {
         *.xz)
             echo "lzma"
             ;;
+        *.sqsh)
+            echo "sqfs"
+            ;;
         *)
             log_fatal "Unsupported compression type: ${disk_image}. Please uncompress the image yourself."
             ;;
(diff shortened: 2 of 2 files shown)


kacf:
@LorenzoRuffati: I think the code here is good. But I'm curious, in what scenario do these sqsh files come up? I might be wrong about this, but I thought squash images contained files, and not additional embedded images?

LorenzoRuffati:
> squash images contained files

Yes, the squash files in this scenarios are simply a squash file system whose only content is a single disk.img file

In my use case I was looking for ways to save space while converting pretty large disk images without having to manually trim the partitions.

I initially thought of using compressed images that are already handled in mender convert, however this ends up using more space in the conversion process since it first inflates the image

The full image is only needed to create clones of the first and second partitions, as such it does not need to be writable. Since squashfs is a read only compressed file system I can mount it and have mender convert operate on the (transparently compressed) file within for the first part of the process

In my case I started from a 120GB disk containing a 60GB root partition (of which around 50GB were actually used). By using this method I was able to use a 25GB sqsh file as input saving anything from 25GB (if I had manually shrunk the partition table and filesystem) to 90GB of disk space and it also saved the time that would've been spent on inflating a gz file

For reference this is the script I use to create the sqsh file from the golden device:

```
#!/bin/sh
if [ "$#" -ne 2 ]; then
    echo "Error: This script requires exactly two arguments."
    echo "Usage: $0 device outputname"
    exit 1
fi
temp=$(mktemp -d)
dd if=$1 bs=4M status=progress conv=fsync | mksquashfs $temp $2 -p "disk.img f 644 0 0 cat"
ext=$?
rmdir $temp
exit $ext
```

kacf:
Hmm, this seems very specific to your environment. I suspect other people will not understand how to use this.

Can't you simply pre-mount the squashfs onto the `input` folder and use it as direct uncompressed input instead?

LorenzoRuffati:
Yes, the "extraction" can be done before invoking mender-convert, the integration into decompressinput was in case this could be useful more generally.

kacf:
Understood, and we do of course appreciate it! But until the `.sqsh` image type is more generally available and used, I think we should reject this to keep things simple, especially since the workaround for you is pretty simple.

PR #593, opened 2023-09-17
feat: add support for BTRFS filesystems

Maik93:
When trying to integrate Mender in an Uboard (x86_64) with Ubuntu Server, I have noticed that there was no support for BTRFS filesystems, so here it is!


kacf (approved):
Thanks @Maik93!

I force pushed minor commit fixes to make the CI happy.

PR #594, opened 2023-09-20
fix: hosted bootstrap support for 'eu' region

Maik93:
Here is a little improvement that took me a while to debug.
Devices will always fail to connect to Hosted Mender if in the `eu` region is not specified.

I have updated the docs accordingly with a separate pull request.

### DIFF
--- a/scripts/bootstrap-rootfs-overlay-hosted-server.sh
+++ b/scripts/bootstrap-rootfs-overlay-hosted-server.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# Copyright 2022 Northern.tech AS
+# Copyright 2023 Northern.tech AS
 #
 #    Licensed under the Apache License, Version 2.0 (the "License");
 #    you may not use this file except in compliance with the License.
@@ -18,6 +18,7 @@ set -o errexit
 
 tenant_token=""
 output_dir=""
+region="us"
 while (("$#")); do
     case "$1" in
         -t | --tenant-token)
@@ -28,9 +29,13 @@ while (("$#")); do
             output_dir="${2}"
             shift 2
             ;;
+        -r | --region)
+            region="${2}"
+            shift 2
+            ;;
         *)
             echo "Sorry but the provided option is not supported: $1"
-            echo "Usage:  $(basename $0) --output-dir ./input/rootfs_overlay_demo --tenant-token <paste your token here>"
+            echo "Usage:  $(basename $0) [--region 'us','eu'] --output-dir ./input/rootfs_overlay_demo --tenant-token <paste your token here>"
             exit 1
             ;;
     esac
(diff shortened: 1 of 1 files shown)


scripts/bootstrap-rootfs-overlay-hosted-server.sh
```diff
@@ -28,6 +29,10 @@ while (("$#")); do
             output_dir="${2}"
             shift 2
             ;;
+        -r | --region)
+            region="${2}"
+            shift 2
+            ;;
         *)
             echo "Sorry but the provided option is not supported: $1"
             echo "Usage:  $(basename $0) --output-dir ./input/rootfs_overlay_demo --tenant-token <paste your token here>"
```
kacf, line 38: Option should be documented as optional here.

scripts/bootstrap-rootfs-overlay-hosted-server.sh
```diff
@@ -50,10 +55,20 @@ if [ -e ${output_dir} ]; then
      sudo chown -R $(id -u) ${output_dir}
      sudo chgrp -R $(id -g) ${output_dir}
 fi
+
+if [ "${region}" = "eu" ]; then
+    region_appendix="eu."
+elif [ "${region}" = "us" ]; then
```
kacf, line 61: Should uppercase also be checked? I suppose it might be a pretty easy mistake to make.

kacf (changes requested):
Hey @Maik93, this makes sense.

A couple of things in addition to the comments below:

* Make sure that you add the following to your commit message:
  ```
  Changelog: Title
  Ticket: None
  ```
* Please update the copyright year of the file to 2023, at the top.

scripts/bootstrap-rootfs-overlay-hosted-server.sh
```diff
@@ -56,9 +56,10 @@ if [ -e ${output_dir} ]; then
      sudo chgrp -R $(id -g) ${output_dir}
 fi
 
-if [ "${region}" = "eu" ]; then
+# check lowercase or uppercase region
+if [ "${region}" = "eu" || "${region}" = "EU" ]; then
```
Maik93, line 60: more general, but I was tempted to write `if [ "${region,,}" = 'eu' ]; then`, but I don't know if it would work on any bash version (on zsh does not).

Maik93:
should I squash commits?

scripts/bootstrap-rootfs-overlay-hosted-server.sh
Maik93: please tell me if the new formatting is ok

kacf: Looks good!

kacf:
> should I squash commits?

@Maik93: Yes, please do! Make sure the Changelog and Ticket tags are included in the squashed commit.

Code looks good otherwise!

kacf (approved):
Thanks! I think the CI failure was just a hiccup in our system, I have restarted it.

PR #624, opened 2024-01-03
test: Run tests for uncompressed images in all cases except one

lluiscampos:
This is to speed up the process, where otherwise `mender-convert` will decompress + convert + compress and then the tests will decompress again.

We do have an explicit test to verify that the compressing logic works fine, so limit it to only here.

### DIFF
--- a/scripts/test/run-tests.sh
+++ b/scripts/test/run-tests.sh
@@ -104,12 +104,16 @@ if [ -n "$PREBUILT_IMAGE" ]; then
 
 else
   if [ "$TEST_ALL" == "1" -o "$TEST_PLATFORM" == "ubuntu-qemux86-64" ]; then
+    # For this test, we explicitly test compressed and uncompressed image conversion
+    # For most of the other tests, we just test the uncompressed image conversion to
+    # speed up the process
     wget --progress=dot:giga -N ${UBUNTU_IMAGE_URL} -P input/image/
+    UBUNTU_IMAGE_COMPRESSED="${UBUNTU_IMAGE_URL##*/}"
     mkdir -p input/tests
     sudo cp -r "tests/ssh-public-key-overlay" "input/tests/"
     convert_and_test "qemux86-64" \
                      "release-1" \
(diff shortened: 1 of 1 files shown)


scripts/test/run-tests.sh
```diff
@@ -148,10 +155,12 @@ else
...
   if [ "$TEST_ALL" == "1" -o "$TEST_PLATFORM" == "raspberrypi3" ]; then
     wget --progress=dot:giga -N ${RASPBIAN_IMAGE_URL} -P input/image/
-    RASPBIAN_IMAGE="${RASPBIAN_IMAGE_URL##*/}"
+    RASPBIAN_IMAGE_COMPRESSED="${RASPBIAN_IMAGE_URL##*/}"
+    RASPBIAN_IMAGE_UNCOMPRESSED="${RASPBIAN_IMAGE_COMPRESSED%.xz}"
+    unxz --force "input/image/${RASPBIAN_IMAGE_COMPRESSED}"
     convert_and_test "raspberrypi3" \
                      "release-1" \
-                     "input/image/${RASPBIAN_IMAGE}" \
+                     "input/image/${RASPBIAN_IMAGE_UNCOMPRESSED}" \
```
kacf, line 163: I think this one (or one of the others) should be kept, because we are testing xz, distinct from gz.

scripts/test/run-tests.sh
```diff
@@ -104,12 +104,15 @@ if [ -n "$PREBUILT_IMAGE" ]; then
 
 else
   if [ "$TEST_ALL" == "1" -o "$TEST_PLATFORM" == "ubuntu-qemux86-64" ]; then
+    # For this test, we explicitly test compressed and uncompressed image conversion
+    # The rest of the tests run uncompressed conversions form of to speed up the process
```
oleorhagen, line 108: Nit: `run _the_ uncompressed

oleorhagen (approved):

scripts/test/run-tests.sh
lluiscampos: Ok. Fair compromise :)

PR #626, opened 2024-01-08
fix: Direct download of deb binary packages with different source package name

lluiscampos:
Up until now we had assumed that the binary package to download had the same name as the source package, but this is not true anymore when `mender-client` source builds `mender-auth` or `mender-update` binary packages.

### DIFF
--- a/mender-convert-modify
+++ b/mender-convert-modify
@@ -104,7 +104,7 @@ if [ "${MENDER_CLIENT_INSTALL}" = "y" ]; then
     case "${MENDER_CLIENT_VERSION}" in
         1.* | 2.* | 3.*)
             log_info "Installing Mender client version ${MENDER_CLIENT_VERSION}"
-            deb_get_and_install_package mender-client "${MENDER_CLIENT_VERSION}"
+            deb_get_and_install_package mender-client mender-client "${MENDER_CLIENT_VERSION}"
             mender_client_deb_name="${DEB_NAME}"
             ;;
         *)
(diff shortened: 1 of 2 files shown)


mender-convert-modify
```diff
@@ -104,20 +104,20 @@ if [ "${MENDER_CLIENT_INSTALL}" = "y" ]; then
...
             ;;
         *)
             # For 4.x, master and latest(*) install the new packages
             # At the time of writing, this will be broken for "latest" until mender-client v4 is
             # released. Having an special handling deemed to complex for the short time benefit.
             log_info "Installing Mender Auth and Mender Update version ${MENDER_CLIENT_VERSION}"
-            deb_get_and_install_package mender-auth "${MENDER_CLIENT_VERSION}"
-            deb_get_and_install_package mender-update "${MENDER_CLIENT_VERSION}"
+            deb_get_and_install_package mender-auth  mender-client "${MENDER_CLIENT_VERSION}"
+            deb_get_and_install_package mender-update mender-client "${MENDER_CLIENT_VERSION}"
```
kacf, line 116: Isn't this the wrong order? Isn't mender-client the recipe, and mender-update the package?

lluiscampos: It is correct, but now that I read your comment I can see that it is inconsistent. On `deb_get_and_install_package` is first package then recipe, while on `deb_from_repo_pool_get` is first recipe then package. I'll invert the arguments in the former.

kacf (approved):

oleorhagen (approved):

PR #628, opened 2024-01-11
Update the base images used

oleorhagen:


.gitlab-ci.yml
```diff
@@ -27,8 +27,8 @@ variables:
       Version of mender-configure for the converted image- 'latest', 'master' or a specific version
 
   ## Auto-update
-  RASPBIAN_URL: "https://downloads.raspberrypi.org/raspios_lite_armhf/images/raspios_lite_armhf-2022-09-26/2022-09-22-raspios-bullseye-armhf-lite.img.xz"
-  RASPBIAN_NAME: 2022-09-22-raspios-bullseye-armhf-lite
+  RASPBIAN_URL: "https://downloads.raspberrypi.org/raspios_lite_armhf/images/raspios_lite_armhf-2023-05-03/2023-05-03-raspios-bullseye-armhf-lite.img.xz"
+  RASPBIAN_NAME: 2023-05-03-raspios-bullseye-armhf-lite
```
lluiscampos, line 31: This image will end up in the pre-built images that we publish, should it have a changelog entry? I think we have not been consistent in the past but it makes sense to me.

oleorhagen: Makes sense to me

PR #636, opened 2024-01-15
chore: Use `dpkg-deb --extract` instead of custom commands

lluiscampos:


oleorhagen (approved):
Nice!

lluiscampos:
I couldn't figure out why the tests fail. Anyway this work is now obsolete in favor of #641.

PR #638, opened 2024-01-17
fix: Remove unneeded and potentially harmful storage assignment.

kacf:
Defining the `mender_grub_storage_device` variable manually has actually not been needed in grub-mender-grubenv for a long time, because it figures it out on its own using the `root` variable, which GRUB fills in during boot. This works better when storage devices change, like when inserting a USB stick, for example.

Although it is definitely unneeded, it's not entirely clear why it messes with the boot in certain instances, since this assignment should happen before any assignment from `root`. But a user has confirmed that this helped.

Changelog: Remove unneeded `mender_grub_storage_device` assignment which could boot from the wrong device if a USB stick was inserted.

Ticket: CE-321

### DIFF
--- a/configs/mender_grub_config
+++ b/configs/mender_grub_config
@@ -29,10 +29,6 @@ MENDER_GRUB_KERNEL_BOOT_ARGS=""
 MENDER_GRUBENV_VERSION="64e32b01d1bf54784d2a290ad0469c583e843864"
 MENDER_GRUBENV_URL="${MENDER_GITHUB_ORG}/grub-mender-grubenv/archive/${MENDER_GRUBENV_VERSION}.tar.gz"
 
-# Name of the storage device containing root filesystem partitions in GRUB
-# format.
-MENDER_GRUB_STORAGE_DEVICE=hd0
-
 # Type of kernel (bzImage or zImage)
 #
 # mender-convert will try to determine this value on its own, only set this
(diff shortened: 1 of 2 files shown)


lluiscampos (approved):

oleorhagen (approved):

TheMeaningfulEngineer:
@kacf 
Do you plan running the integration test before the merge?

kacf:
@TheMeaningfulEngineer: Sorry, I know it's confusing: Specifically for mender-convert we don't run the integration tests, because mender-convert has its own set of tests (and integration doesn't use mender-convert anyway). So this is in fact all green!

buentead:
I successfully tested this PR with a Lenovo M90n as follows:
- Creating an image with this version of mender-convert
- Loaded the image onto an SSD (/dev/nvme0n1) from USB stick
- Leave USB stick plugged-in and executing 'reboot'
UEFI will assign hd0 to the USB stick and hd1 to the SSD 
- Booting the OS from SSD (hd1) while USB stick (hd0) was inserted
- Remove USB stick from device and reboot OS
Now, hd0 will be assigned to the SSD by UEFI
- Booting the OS from SSD (hd0) without any USB stick inserted

Many thanks for the fix. I hope it makes it rather fast into the LTS version ;-)

kacf:
Thanks for testing @buentead! This has already been [cherry-picked to the 4.2.x](https://github.com/mendersoftware/mender-convert/pull/640) branch, so it should be part of the next patch update.

PR #641, opened 2024-01-24
Features to enable support for Mender client 4.0.0

kacf:


modules/container.sh
```diff
@@ -0,0 +1,69 @@
...
+        run_and_log_cmd "sudo mount /sys $directory/sys -o bind,ro"
+
+        # Run command.
+        "$@"
+    ) || ret=$?
+
+    # Very important that these are unmounted, otherwise Docker may start to remove files inside
+    # them while tearing down the container. You can guess how I found that out... We run without
+    # the logger because otherwise the message from the previous command, which is the important
+    # one, is lost.
```
oleorhagen, line 65: Hmm, why is it lost? :thinking:

modules/container.sh
```diff
@@ -0,0 +1,69 @@
+#!/usr/bin/env bash
+# Copyright 2024 Northern.tech AS
```
oleorhagen, line 2: tbh, I think I would name it `chroot.sh`

modules/container.sh
```diff
@@ -21,6 +21,14 @@ function run_with_chroot_setup() {
     shift
     # The rest of the arguments are the command arguments.
 
+    # Keep a copy of the resolv.conf, and copy in our own, we will need it during chroot
```
oleorhagen, line 26: O: This would have read a lot easier if it was a function wit ha name like - ensure DNS something something

mender-convert-modify
```diff
@@ -222,15 +222,13 @@ run_and_log_cmd "sudo chmod 600 work/rootfs/data/mender/mender.conf"
 if [ -z "${MENDER_DEVICE_TYPE}" ]; then
     # Observed systems who do not have this file, e.g images generated with mkosi
     if [ -f work/rootfs/etc/hostname ]; then
-        device_type=$(cat work/rootfs/etc/hostname)
+        MENDER_DEVICE_TYPE=$(cat work/rootfs/etc/hostname)
```
oleorhagen, line 170: Nice!

mender-convert-modify
```diff
@@ -277,6 +222,61 @@ ${data_part_device}   /data          auto       ${MENDER_DATA_PART_FSTAB_OPTS}
 $fstab
 EOF
 
+log_info "Installing Mender client and related files"
+
+if [ "${MENDER_CLIENT_INSTALL}" = "y" ]; then
+    case "${MENDER_CLIENT_VERSION}" in
+        1.* | 2.* | 3.*)
+            log_info "Installing Mender client version ${MENDER_CLIENT_VERSION}"
```
oleorhagen, line 230: I'm going to leave a few comments on pre-existing stuff, feel free to do with it as you like.

But this message should be outside the `case`. It's generic

mender-convert-modify
```diff
@@ -277,6 +222,61 @@ ${data_part_device}   /data          auto       ${MENDER_DATA_PART_FSTAB_OPTS}
...
+    esac
+
+    # Save installed client version for tests in Yocto variable format
+    testscfg_add "PREFERRED_VERSION:mender-client" "$(echo ${mender_client_deb_name} | sed -r 's/.*_([0-9]+\.[0-9]+\.[0-9]+).*/\1/')"
+fi
+
+if [ "${MENDER_ENABLE_SYSTEMD}" == "y" ]; then
+    case "${MENDER_CLIENT_VERSION}" in
+        1.* | 2.* | 3.*)
+            run_and_log_cmd "sudo ln -sf /lib/systemd/system/mender-client.service \
```
oleorhagen, line 273: Hmm, are the packages not enabled by default? :thinking:

mender-convert-modify
```diff
@@ -277,6 +222,61 @@ ${data_part_device}   /data          auto       ${MENDER_DATA_PART_FSTAB_OPTS}
...
+        1.* | 2.* | 3.*)
+            log_info "Installing Mender client version ${MENDER_CLIENT_VERSION}"
+            deb_get_and_install_package mender-client mender-client "${MENDER_CLIENT_VERSION}"
+            mender_client_deb_name="${DEB_NAME}"
+            ;;
+        *)
+            # For 4.x, master and latest(*) install the new packages
+            # At the time of writing, this will be broken for "latest" until mender-client v4 is
+            # released. Having an special handling deemed to complex for the short time benefit.
+            log_info "Installing Mender Auth and Mender Update version ${MENDER_CLIENT_VERSION}"
```
oleorhagen, line 238: S: This could be logged in the `deb_get_and_install_package` function

mender-convert-modify
```diff
@@ -277,6 +222,61 @@ ${data_part_device}   /data          auto       ${MENDER_DATA_PART_FSTAB_OPTS}
...
+            deb_get_and_install_package mender-client mender-auth "${MENDER_CLIENT_VERSION}"
+            deb_get_and_install_package mender-client mender-update "${MENDER_CLIENT_VERSION}"
+            mender_client_deb_name="${DEB_NAME}"
```
oleorhagen, line 241: Implicit value return - also pre-existing. But I wish we returned it from the function instead

mender-convert-modify
```diff
@@ -277,6 +222,61 @@ ${data_part_device}   /data          auto       ${MENDER_DATA_PART_FSTAB_OPTS}
...
+if [ "${MENDER_ENABLE_SYSTEMD}" == "y" ]; then
+    case "${MENDER_CLIENT_VERSION}" in
+        1.* | 2.* | 3.*)
+            run_and_log_cmd "sudo ln -sf /lib/systemd/system/mender-client.service \
+                work/rootfs/etc/systemd/system/multi-user.target.wants/mender-client.service"
+            ;;
+        *)
+            run_and_log_cmd "sudo ln -sf /lib/systemd/system/mender-auth.service \
+                work/rootfs/etc/systemd/system/multi-user.target.wants/mender-auth.service"
+            run_and_log_cmd "sudo ln -sf /lib/systemd/system/mender-update.service \
```
oleorhagen, line 279: This we de quite a few places. Could make a utility function out of it

modules/deb.sh
```diff
@@ -95,31 +96,67 @@ function deb_from_repo_pool_get()  {
...
+        log_fatal "deb_install_package() requires 2 arguments"
     fi
     local -r deb_package="$(pwd)/${1}"
     local -r dest_dir="$(pwd)/${2}"
 
-    local -r extract_dir=$(mktemp -d)
-    cd ${extract_dir}
-    run_and_log_cmd "ar -xv ${deb_package}"
-    mkdir -p files
+    (
```
oleorhagen, line 111: Nit - random whitespace

modules/deb.sh
```diff
@@ -95,31 +96,67 @@ function deb_from_repo_pool_get()  {
...
 
-    run_and_log_cmd "sudo rsync --archive --keep-dirlinks --verbose ${extract_dir}/files/ ${dest_dir}"
+    # Copy package to somewhere reachable from within the chroot.
+    cp "$deb_package" "$dest_dir/tmp/"
+
+    local ret=0
+    run_in_chroot_and_log_cmd_noexit "$dest_dir" "env \
+        DEBIAN_FRONTEND=noninteractive \
+        DEVICE_TYPE=${MENDER_DEVICE_TYPE} \
+        dpkg -i /tmp/$(basename $deb_package)" || ret=$?
```
oleorhagen, line 133: S: `s/-i/--install`

oleorhagen: Q: Why not install with `apt` straight away? :thinking:

modules/container.sh
```diff
@@ -29,6 +31,13 @@ function run_with_chroot_setup() {
     fi
     cp /etc/resolv.conf "$directory/etc/resolv.conf"
 
+    local arch="$(probe_arch)"
+    if [ "$arch" != "$(uname -m)" ]; then
+        # Foreign architecture, we need to use QEMU.
+        cp "$(which "qemu-$arch-static")" "$directory/tmp/"
```
oleorhagen, line 37: Nice! I was not familir with `Qemu` static

modules/container.sh
```diff
@@ -68,15 +87,29 @@ function run_in_chroot_and_log_cmd() {
...
+    local arch="$(probe_arch)"
+    local maybe_qemu=
+    if [ "$arch" != "$(uname -m)" ]; then
+        # Use env, because qemu does not look in PATH.
+        maybe_qemu="/tmp/qemu-$arch-static /usr/bin/env"
+    fi
+
     local ret=0
-    run_and_log_cmd_noexit "sudo chroot $directory $@" || ret=$?
+    run_and_log_cmd_noexit "sudo chroot $directory $maybe_qemu $@" || ret=$?
```
oleorhagen, line 113: Nice!

oleorhagen (approved):
Looks good I think!

Left a few nitpicks and observations.

In general, try and use `read-only` variables as much as you can. There are enough moving pieces as there is

modules/container.sh
```diff
@@ -0,0 +1,69 @@
...
+# Takes a directory to set up for chroot and a name of a function to run. Note that this function
+# does not run *inside* the chroot, it just runs with chroot set up. To actually run commands, use
+# one of the `run_in_chroot` functions individually for each command.
+function run_with_chroot_setup() {
+    local directory="$1"
+    shift
+    # The rest of the arguments are the command arguments.
+
+    local ret=0
+    (
```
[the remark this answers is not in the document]
kacf: It's [`shfmt`](https://github.com/mendersoftware/mendertesting/blob/master/.gitlab-ci-check-shell-format.yml#L61) which adds these, not me.

modules/container.sh
kacf: I don't recall the exact details, but I think it's because `run_and_log_cmd` prints *the last* command you executed only, when there is an error. Therefore the output is really confusing because the "error" is always going to be empty = successful unmount.

modules/container.sh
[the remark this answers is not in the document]
oleorhagen: Oh true.. I knew this actually...

mender-convert-modify
kacf: Hmm, this might not be necessary anymore. I'll check it.

modules/deb.sh
kacf: AFAIK you cannot install a package file with apt, only a repository package. I did consider doing exactly that in the case when we are requesting the latest package, but then the two code paths would be different, so I chose to keep the dpkg command in both cases.

oleorhagen: You don't have to do this obviously, but we could consider doing something similar to this: https://github.com/mendersoftware/mender-dist-packages/blob/4cd13030f105bb6f922d57877e468ce174f8fdb7/tests/test_install_mender_sh.py#L99

modules/deb.sh
```diff
@@ -95,31 +96,74 @@ function deb_from_repo_pool_get()  {
...
+            fi
 
-    log_info "Successfully installed $(basename ${deb_package}) into ${dest_dir}"
+            # It's ok if dpkg fails, as long as a subsequent "apt install" succeeds. This happens
+            # when we try to install a package and its dependencies are not immediately
+            # satisfied. Then we need to run apt afterwards to remedy the problem.
+            run_in_chroot_and_log_cmd "$dest_dir" "env \
+                DEBIAN_FRONTEND=noninteractive \
+                DEVICE_TYPE=${MENDER_DEVICE_TYPE} \
+                apt --fix-broken -y install"
```
lluiscampos, line 160: Once you are using the long option `--fix-broken`:
```suggestion
                apt --fix-broken --assume-yes install"
```

mender-convert-modify
lluiscampos: I also assume that these should be now handled by `dpkg` in postinst scripts.

lluiscampos (approved):
Thank you @kacf for a really nice solution!

(And thank you @oleorhagen for a thorough review - I have nothing of value to add)

mender-convert-modify
kacf: The message is different for the two cases due to the different packages being installed. I'd rather leave it like this.

mender-convert-modify
kacf: I think it should be logged here, because the deb installer logging is much more detailed and low level, with a full URL. Which is also useful, but much harder to read. The message right here is the message I was looking for myself when I was watching the log go by.

mender-convert-modify
kacf: I agree. But it's not so trivial to fix this, because it relies on the fetching logic. We'd either have to:

1. Repeat the call in order to get the name (expensive, because we have to re-download every package).
2. Restructure the code so that the download happens first, and then this is passed into the installer afterwards when we already have the name. Doable, but quite an intrusive change, since it puts more work on `mender-convert-modify`.
3. Keep it as is, but print the `DEB_NAME` in the function. Error prone, because there is a lot of output from the deb installer as well.

Definitely not unsolvable, but I think I'll let this dragon sleep for today.

modules/deb.sh
kacf: Not sure I see the gain though.. It looks even more complicated than [what we're already doing](https://github.com/mendersoftware/mender-convert/blob/master/modules/deb.sh#L29).

mender-convert-modify
kacf: It is. I removed this part.

mender-convert-modify
oleorhagen: Yes, not going to force this on you then :)

modules/container.sh
kacf: I did this, but then reverted it again (same for the `qemu-$arch-static` copy). It relies on variables I deduce in this function, so it becomes more messy than it's worth IMHO, since we have to then recalculate them four times (twice in setup functions, twice in teardown functions).

PR #649, opened 2024-02-20
feat: add support for `MENDER_CLIENT_VERSION="auto"`

tranchitella:
Changelog: feat: add support for `MENDER_CLIENT_VERSION="auto"`, which installs the Mender Client 3.x series for Debian bullseye and buster and for Ubuntu jammy and focal, the Mender Client 4.x series for all the other distributions
Ticket: MEN-6976

### DIFF
--- a/configs/mender_convert_config
+++ b/configs/mender_convert_config
@@ -124,9 +124,13 @@ MENDER_CLIENT_INSTALL="y"
 #
 # This is used to fetch the correct binaries
 #
-# Valid values are "latest" (default), "master" or a specific version
-# MEN-6948 mender-convert is not yet integrated with latest mender-client 4.0.0
-MENDER_CLIENT_VERSION="3.5.2"
+# Valid values are "auto" (default), "latest", "master" or a specific version
+# If set to "auto" then the Mender Client 4.0 or later will be installed unless
+# the distribution is Ubuntu jammy or older or Debian bullseye or older;
+# the latest Mender 3.x client will be installed in those cases.
+# If set to "latest", it will install the latest version of the Mender Client
+# from the 4.x series.
+MENDER_CLIENT_VERSION="auto"
 
 # Mender flash version
 #
--- a/mender-convert-modify
+++ b/mender-convert-modify
@@ -225,19 +225,32 @@ EOF
 log_info "Installing Mender client and related files"
 
 if [ "${MENDER_CLIENT_INSTALL}" = "y" ]; then
+    mender_client_deb_version="$MENDER_CLIENT_VERSION"
+    if [ "$MENDER_CLIENT_VERSION" = "auto" ]; then
+        deb_distro=$(probe_debian_distro_name)
+        deb_codename=$(probe_debian_distro_codename)
+        if ([ "$deb_distro" = "debian" ] && [ "$deb_codename" = "bullseye" ]) \
+            || ([ "$deb_distro" = "debian" ] && [ "$deb_codename" = "buster" ]) \
+            || ([ "$deb_distro" = "ubuntu" ] && [ "$deb_codename" = "jammy" ]) \
+            || ([ "$deb_distro" = "ubuntu" ] && [ "$deb_codename" = "focal" ]); then
+            MENDER_CLIENT_VERSION="3.5.x"
+            mender_client_deb_version="latest"
+        else
+            MENDER_CLIENT_VERSION="latest"
+            mender_client_deb_version="latest"
+        fi
+    fi
     case "${MENDER_CLIENT_VERSION}" in
         1.* | 2.* | 3.*)
-            log_info "Installing Mender client version ${MENDER_CLIENT_VERSION}"
-            deb_get_and_install_package mender-client mender-client "${MENDER_CLIENT_VERSION}"
+            log_info "Installing Mender client version ${mender_client_deb_version}"
+            deb_get_and_install_package mender-client mender-client "${mender_client_deb_version}"
             mender_client_deb_name="${DEB_NAME}"
             ;;
         *)
             # For 4.x, master and latest(*) install the new packages
-            # At the time of writing, this will be broken for "latest" until mender-client v4 is
-            # released. Having an special handling deemed to complex for the short time benefit.
-            log_info "Installing Mender Auth and Mender Update version ${MENDER_CLIENT_VERSION}"
-            deb_get_and_install_package mender-client mender-auth "${MENDER_CLIENT_VERSION}"
-            deb_get_and_install_package mender-client mender-update "${MENDER_CLIENT_VERSION}"
+            log_info "Installing Mender Auth and Mender Update version ${mender_client_deb_version}"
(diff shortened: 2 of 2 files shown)


mender-convert-modify
```diff
@@ -225,19 +225,34 @@ EOF
...
 
 if [ "${MENDER_CLIENT_INSTALL}" = "y" ]; then
+    MENDER_CLIENT_DEB_VERSION="$MENDER_CLIENT_VERSION"
+    if [ "$MENDER_CLIENT_VERSION" = "auto" ]; then
+        deb_distro=$(probe_debian_distro_name)
+        deb_codename=$(probe_debian_distro_codename)
+        if ([ "$deb_distro" = "debian" ] && [ "$deb_codename" = "bullseye" ]) \
+            || ([ "$deb_distro" = "debian" ] && [ "$deb_codename" = "buster" ]) \
+            || ([ "$deb_distro" = "ubuntu" ] && [ "$deb_codename" = "jammy" ]) \
+            || ([ "$deb_distro" = "ubuntu" ] && [ "$deb_codename" = "focal" ]); then
```
kacf, line 235: Suggestion: Could also consider using `VERSION_ID` which is formatted as `12` and `22.04` for Debian and Ubuntu, respectively. I *think* you can use numeric comparisons on both, but not completely sure about the second one (unsure how the shell treats decimal point).

kacf (approved):

configs/mender_convert_config
```diff
@@ -124,9 +124,11 @@ MENDER_CLIENT_INSTALL="y"
...
 # This is used to fetch the correct binaries
 #
-# Valid values are "latest" (default), "master" or a specific version
-# MEN-6948 mender-convert is not yet integrated with latest mender-client 4.0.0
-MENDER_CLIENT_VERSION="3.5.2"
+# Valid values are "auto" (default), "latest", "master" or a specific version
+# If set to "auto" then the Mender Client 4.0 or later will be installed unless
+# the distribution is Ubuntu jammy or older or Debian bullseye or older;
+# the latest Mender 3.x client will be installed in those cases.
+MENDER_CLIENT_VERSION="auto"
```
lluiscampos, line 131: Okey, and what "latest" will mean in this context? We should document it

lluiscampos: Independently, we need to update [this docs page](https://docs.mender.io/operating-system-updates-debian-family/variables#mender_client_version).

mender-convert-modify
```diff
@@ -225,19 +225,34 @@ EOF
...
+            MENDER_CLIENT_VERSION="3.5.x"
```
lluiscampos, line 236: I am missing something, how is this supposed to work? We won't be able to download a 3.5.**x** package, we need somewhere to translate this into a tag.

I think rather you want to select on package name, like for this case `mender-client` and in the _else_ below the `mender-client4` package.

lluiscampos (changes requested):

mender-convert-modify
tranchitella: @lluiscampos 3.5.x will download the latest version of the `mender-client` package; in fact, anything which starts with "3." will trigger the first stanza of the if. The effect is "latest" on the "mender-client" package.

configs/mender_convert_config
tranchitella: Latest is the latest version, so 4.0.2 as of today. 
I will follow up with the documentation.

mender-convert-modify
lluiscampos: You are right - I didn't pay attention to that we are already installing `mender-auth` and `mender-udpate` below.

configs/mender_convert_config
lluiscampos, line 131: We should set `MENDER_CLIENT_VERSION` in CI to `auto` too (and take care of [this line](https://github.com/mendersoftware/mender-convert/blob/cd3d3067688ae34d3666d7f5327bdae42aaefeb6/.gitlab-ci.yml#L230)) so that we run tests with the C++ client as we were doing up until now.

mender-convert-modify
```diff
@@ -225,19 +225,34 @@ EOF
 log_info "Installing Mender client and related files"
 
 if [ "${MENDER_CLIENT_INSTALL}" = "y" ]; then
+    MENDER_CLIENT_DEB_VERSION="$MENDER_CLIENT_VERSION"
```
lluiscampos, line 228: Then I suggest we rename this lowercase - we use uppercase for configuration or "global" variables.
```suggestion
    mender_client_deb_version="$MENDER_CLIENT_VERSION"
```

lluiscampos (approved):
Sorry for my wrong assessment before - it looks good!

I commented on one new thing wrt CI, but pre-approving here :+1:

configs/mender_convert_config
tranchitella: I don't get why you want to change the `MENDER_CLIENT_VERSION` to `auto` in the pipeline. Let's discuss it when you have time for it.

lluiscampos: I was wrong - letting it be `latest` will in effect install and test the C++ client

lluiscampos (approved):

robertalexa:
Hi guys,

```
# Valid values are "auto" (default), "latest", "master" or a specific version
# If set to "auto" then the Mender Client 4.0 or later will be installed unless
# the distribution is Ubuntu jammy or older or Debian bullseye or older;
# the latest Mender 3.x client will be installed in those cases.
# If set to "latest", it will install the latest version of the Mender Client
# from the 4.x series.
MENDER_CLIENT_VERSION="auto"
```

Does the above mean that Mender Client 4.0 is not supported on bullseye or older? Asking this because if true, it would suggest Debian bookworm should be used instead, but from the info on the forums it doesn't seem bookworm is supported?

Thanks in advance

lluiscampos:
> Hi guys,
> 
> ```
> # Valid values are "auto" (default), "latest", "master" or a specific version
> # If set to "auto" then the Mender Client 4.0 or later will be installed unless
> # the distribution is Ubuntu jammy or older or Debian bullseye or older;
> # the latest Mender 3.x client will be installed in those cases.
> # If set to "latest", it will install the latest version of the Mender Client
> # from the 4.x series.
> MENDER_CLIENT_VERSION="auto"
> ```
> 
> Does the above mean that Mender Client 4.0 is not supported on bullseye or older? Asking this because if true, it would suggest Debian bookworm should be used instead, but from the info on the forums it doesn't seem bookworm is supported?
> 
> Thanks in advance

Hi @robertalexa It is supported (it will be soon, we are finalizing the formalities of the release).

What the message meant to say is that using the "auto" option (which is the default) it will still use 3.x for existing distros and only select 4.x for newer distros starting in Bookworm. You can set it to `latest` to install latest 4.x on any distro, or you can still pin it to an specific version as before.

robertalexa:
Thank you @lluiscampos . Appreciate your input. Will keep our eyes peeled for the official release.

PR #655, opened 2024-04-05
Rpi fixes

kacf:
Ticket: MEN-7143

### DIFF
--- a/configs/raspberrypi_config
+++ b/configs/raspberrypi_config
@@ -101,9 +101,11 @@ function platform_modify() {
 
     # Raspberry Pi applications expect to find this on the device and in some
     # cases parse the options to determine the functionality. Create symlinks
-    # to all files
+    # to all files that don't already exist.
     for f in $(ls -1 work/boot); do
-        run_and_log_cmd "sudo ln -fs /uboot/${f} work/rootfs/boot/${f}"
+        if [ ! -f "work/rootfs/boot/${f}" ]; then
+            run_and_log_cmd "sudo ln -fs /uboot/${f} work/rootfs/boot/${f}"
+        fi
     done
 
     # Raspberry Pi headless configuration and other first boot scripts inspect the boot partition
--- a/configs/raspberrypi_config
+++ b/configs/raspberrypi_config
@@ -83,6 +83,10 @@ function platform_modify() {
         run_and_log_cmd "sudo cp work/boot/${RASPBERRYPI_KERNEL_IMAGE} work/rootfs/boot/${MENDER_KERNEL_IMAGETYPE}"
     fi
 
+    # Make sure the on-board bootloader boots our bootloader (which we do via the kernel option).
+    sed -i -e '/^kernel=/d' work/boot/config.txt
+    echo "kernel=${RASPBERRYPI_BOOTLOADER_IMAGE}" >> work/boot/config.txt
+
     # Replace kernel with U-boot and add boot script
     run_and_log_cmd "sudo mkdir -p work/rootfs/uboot"
     run_and_log_cmd "sudo cp work/rpi/binaries/u-boot.bin work/boot/${RASPBERRYPI_BOOTLOADER_IMAGE}"
--- a/configs/raspberrypi_config
+++ b/configs/raspberrypi_config
@@ -96,7 +96,7 @@ function platform_modify() {
     # for files. Modify the hard-coded mount point for the partition from /boot to /uboot.
 
     # Integrate with systemd services with "/boot/" hardcoded
-    files_to_modify=$(find work/rootfs/lib/ -type f -name "*.service" | xargs grep '/boot/' | cut -d: -f1 | uniq)
+    files_to_modify="$(find work/rootfs/lib/ -type f -name "*.service" -exec grep -l '/boot/' '{}' ';')"
 
     # MEN-5944: Integrate with raspberrypi-sys-mods/firstboot
     # See: https://github.com/RPi-Distro/raspberrypi-sys-mods
--- a/configs/raspberrypi_config
+++ b/configs/raspberrypi_config
@@ -14,6 +14,11 @@ RASPBERRYPI_BINARIES="${RASPBERRYPI_CONFIG}-2020.01.tar.gz"
 RASPBERRYPI_BINARIES_URL="${MENDER_STORAGE_URL}/mender-convert/uboot/raspberrypi/${RASPBERRYPI_BINARIES}"
 
 function platform_modify() {
+    if file -L work/rootfs/bin/sh | grep -qF "ELF 64-bit"; then
+        echo "There is no 64-bit support for Raspberry Pi yet, please use a 32-bit image".
+        exit 1
+    fi
+
     mkdir -p work/rpi/binaries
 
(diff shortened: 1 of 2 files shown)


configs/raspberrypi_config
```diff
@@ -78,25 +83,40 @@ function platform_modify() {
         run_and_log_cmd "sudo cp work/boot/${RASPBERRYPI_KERNEL_IMAGE} work/rootfs/boot/${MENDER_KERNEL_IMAGETYPE}"
     fi
 
+    # Make sure the on-board bootloader boots our bootloader (which we do via the kernel option).
+    sed -i -e '/^kernel=/d' work/boot/config.txt
+    echo "kernel=${RASPBERRYPI_BOOTLOADER_IMAGE}" >> work/boot/config.txt
```
merlin-northern, line 88:
```suggestion
    sed -i -e "s|^kernel=.*|kernel=${RASPBERRYPI_BOOTLOADER_IMAGE}|" work/boot/config.txt
```
I am assuming that there is a `kernel=something` only on the line, i.e.: `.*` is enough, and we do not have to do other funny stuff.

configs/raspberrypi_config
```diff
@@ -78,25 +83,40 @@ function platform_modify() {
...
     run_and_log_cmd "sudo cp work/rpi/binaries/fw_env.config work/rootfs/etc/"
     run_and_log_cmd "sudo cp work/rpi/binaries/uboot-git-log.txt work/boot"
 
+    # Disable 64-bit mode and force 32-bit mode. Even though we check for 32-bit binaries at the
+    # start of this function, some images use a 64-bit kernel even with 32-bit userspace. Due to our
+    # custom U-Boot bootloader we need to stick to 32-bit for now.
+    if ! grep -q '^arm_64bit=0' work/boot/config.txt; then
+        log_info "Forcing 32-bit kernel"
+        sed -i -e '/^arm_64bit=/d' work/boot/config.txt
+        echo "arm_64bit=0" >> work/boot/config.txt
```
merlin-northern, line 103:
```suggestion
        sed -i -e 's/^arm_64bit=[0-9]*/arm_64bit=0/' work/boot/config.txt
```
this can be done also without the if, since we can replace only `arm_64bit=[^0]*`, but I understand that you want the log line. I am unsure about `[0-9]*` in the above, can we assume that there is always zero or more digits after `arm_64bit=`? but either way, you can do it with one sed. as most of the things. of which I know nothing, but I know you can ;-)

configs/raspberrypi_config
```diff
@@ -78,25 +83,40 @@ function platform_modify() {
...
+    fi
+
     # Raspberry Pi applications expect to find this on the device and in some
     # cases parse the options to determine the functionality. Create symlinks
-    # to all files
+    # to all files that don't already exist.
     for f in $(ls -1 work/boot); do
```
merlin-northern, line 109: f should be a local variable in this scope. and you should traverse the files via:
```bash
while read -r f; do 
 # now remember that "$f" contains the full path
 [[ -e "$f" ]] && run_and_log_cmd "sudo ln -fs /uboot/`basename ${f}` work/rootfs/boot/${f}"
 # I do not uderstand ln command, see below comment.
done < <(find work/boot -type f)
```
`-e` is on purpose there? I am so used to `-f` :)

configs/raspberrypi_config
```diff
@@ -78,25 +83,40 @@ function platform_modify() {
...
-        run_and_log_cmd "sudo ln -fs /uboot/${f} work/rootfs/boot/${f}"
+        if [ ! -e "work/rootfs/boot/${f}" ]; then
+            run_and_log_cmd "sudo ln -fs /uboot/${f} work/rootfs/boot/${f}"
```
merlin-northern, line 111: I do not understand the ln command here, `/uboot/${f}` is the target, and link name is `work/rootfs/boot/${f}`, just please double check if thats what we want here (I lack the knowledge). also if you decide to go along with the find, remember that it will contain the full path, so you have to use basename appropriately.

configs/raspberrypi_config
```diff
@@ -78,25 +83,40 @@ function platform_modify() {
...
+        fi
     done
 
     # Raspberry Pi headless configuration and other first boot scripts inspect the boot partition
     # for files. Modify the hard-coded mount point for the partition from /boot to /uboot.
 
     # Integrate with systemd services with "/boot/" hardcoded
-    files_to_modify=$(find work/rootfs/lib/ -type f -name "*.service" | xargs grep '/boot/' | cut -d: -f1 | uniq)
+    files_to_modify="$(find work/rootfs/lib/ -type f -name "*.service" | ( xargs -r grep -l '/boot/' || true ) )"
```
merlin-northern, line 119: is there a reason why we dont use `-exec` to `find` here?
I also a bit fail to understand why we dont `grep -l '/boot/' work/rootfs/lib/*.service`?

merlin-northern (commented):
I am not an expert, I tried to be constructive. treat as suggestions.

lluiscampos (approved):
The fixes look good to me. Great work. @merlin-northern did a pretty good job reviewing the actual shell syntax, so follow-up there where propitiate :smiley:

configs/raspberrypi_config
kacf: `kernel=` can occur multiple times in the file, because it can be listed under each `[rpi3]` / `[rpi4]` tag. The idea behind this is to delete all of them, and add exactly one at the bottom, which is under `[all]`. While this is not completely ideal either (the file we point to is actually not cross-rpi-compatible), it is less confusing than having multiple entries where most of them point to the wrong file.

configs/raspberrypi_config
kacf: Same argument as in the previous comment.

configs/raspberrypi_config
kacf: `f` is local when it is for-loop-initialized, isn't it?

`-e` was originally on purpose, but maybe `-f` is better. Then at least we will get an error message if it's a directory.

I'm not sure I get your variant though, what does it fix?

configs/raspberrypi_config
kacf: `/uboot` is the boot partition, and `/boot` is on the root partition (I know, confusing, but that's how it is). We want to link to some files (most importantly, dtb files) on the boot partition, so this should indeed be correct: We link files in `/boot` to `/uboot`. This isn't a very clever integration, because it increases the reliance on the boot partition, which isn't updated in rootfs-image updates, but it's not something I want to change now.

kacf:
> but wow, great work, it would take me a week to come up with that. how can it break like this just like that?

The heart of the issue is that we're lagging behind. Our integration is 32-bit only, and based on a very old fork of U-Boot. Meanwhile they are moving more and more towards 64-bit. If they stop releasing 32-bit images, then our integration will also stop working.

Ideally we should move to a more modern integration, either by utilizing the built-in Raspberry Pi bootloader more, or by moving to UEFI.

configs/raspberrypi_config
kacf: I don't think I will do any more followup here, the loop is anyway pre-existing.

PR #658, opened 2024-04-14
fix: missing DTB inclusion in armbian-based images

emanuele-f:
This fix allows U-Boot to properly load the DTB files when chainloading GRUB via UEFI in armbian-based images. In such images, DTBs are normally located in the symlinked `/boot/dtb` directory and should retain the original directory structure.

This fixes missing wifi interface and cpu sensors with the Orange Pi Zero 3 DietPi converted image.

This complements work done in c321684a

### DIFF
--- a/modules/grub.sh
+++ b/modules/grub.sh
@@ -181,5 +181,11 @@ function grub_install_mender_grub() {
                 break
             fi
         done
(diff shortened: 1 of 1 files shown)


kacf (approved):
Force pushed to add changelog tag. Otherwise looks good to me, thanks!

PR #664, opened 2024-05-07
ci: from unavailable public runner to self-hosted

oldgiova:
Using self-hosted runner because the docker runner is no longer available

Ticket: SEC-1133
Changelog: None

the form below:

```
fix: <SHORT DESCRIPTION OF FIX>

<OPTIONAL LONGER DESCRIPTION>

Changelog: <USER-FRIENDLY-CHANGE-DESCRIPTION> or <None>
Ticket: <TICKET NUMBER> or <None>
```

### Description

Using self-hosted runner because the docker runner is no longer available

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -384,7 +384,7 @@ test_acceptance_ubuntu_raspberrypi3:
 .template:publish:docker-image:
   stage: publish
   tags:
-    - docker
+    - hetzner-amd-beefy
   image: docker
   services:
     - name: docker:20.10.21-dind


.gitlab-ci.yml
```diff
@@ -384,7 +384,7 @@ test_acceptance_ubuntu_raspberrypi3:
 .template:publish:docker-image:
   stage: publish
   tags:
-    - docker
+    - hetzner-amd-beefy
```
lluiscampos, line 387: I did not research further, but changing yesterday jobs from public `docker` tags to `hetzner-amd-beefy` did not work for `mender-artifact` (link [here](https://github.com/mendersoftware/mender-artifact/pull/610#issuecomment-2097861179)) due to the runners having different Docker in Docker setup.

This job also we cannot test until we merge the PR, so I am okey with merging as-is as long as you keep an eye on the master pipeline :+1:

lluiscampos (approved):

.gitlab-ci.yml
oldgiova: It looks like the `gitlab-org-docker` runner is [actually using privileged runners](https://docs.gitlab.com/ee/development/pipelines/internals.html#runner-tags). I think this is bad for a security point of view. Let me check in master, then

oldgiova: The issues could be caused by the custom MTU setting: see https://github.com/mendersoftware/mender-convert/pull/665

PR #689, opened 2024-06-18
feat: allow installation of arbitrary debian packages

danielskinstad:
Changelog: Add the possibility to install arbitrary debian packages located in `MENDER_INSTALL_INPUT_PACKAGES_PATH` to the converted image.

Ticket: MEN-7264

### DIFF
--- a/configs/mender_convert_config
+++ b/configs/mender_convert_config
@@ -157,6 +157,19 @@ MENDER_ADDON_CONFIGURE_INSTALL="n"
 # Valid values are "latest" (default), "master" or a specific version
 MENDER_ADDON_CONFIGURE_VERSION="latest"
 
+# Install arbitrary debian packages by placing them in
+# MENDER_INSTALL_INPUT_PACKAGES_PATH. Specify the order
+# of installation (in case of dependencies where package
+# B depends on A) by adding a file called `order` to the directory.
+#
+# Directory containing deb packages to be installed in the image
+#
+MENDER_INSTALL_INPUT_PACKAGES_PATH="input/deb"
+
+# Install deb packages located in MENDER_INSTALL_INPUT_PACKAGES_PATH
+#
+MENDER_INSTALL_INPUT_PACKAGES="n"
+
 # Install Mender-setup
 # If set to "auto" then it is installed if Mender client 4.0 or later is
 # installed. Otherwise it is not installed, but the same functionality is built
--- a/modules/deb.sh
+++ b/modules/deb.sh
@@ -130,8 +130,8 @@ function deb_install_package_in_chroot() {
     run_in_chroot_and_log_cmd_noexit "$dest_dir" "env \
         DEBIAN_FRONTEND=noninteractive \
         DEVICE_TYPE=${MENDER_DEVICE_TYPE} \
-        dpkg --install /tmp/$(basename $deb_package)" || ret=$?
-    rm -f "$dest_dir/tmp/$(basename $deb_package)"
+        dpkg --install \"/tmp/$(basename "$deb_package")\"" || ret=$?
+    rm -f "$dest_dir/tmp/$(basename "$deb_package")"
     case $ret in
         0)
             # Success
--- a/modules/deb.sh
+++ b/modules/deb.sh
@@ -221,6 +221,45 @@ function deb_get_and_install_package() {
     deb_install_package "work/deb-packages/${DEB_NAME}" "work/rootfs/"
 }
 
+#  Install binary files of a deb package into work/deb-packages from a local directory
+#  $1 - Directory path
+#
+function deb_get_and_install_input_package() {
+    if ! [[ $# -eq 1 ]]; then
+        log_fatal "deb_get_and_install_input_package() requires 1 argument"
+    fi
+
+    local -r package_path="$1"
+    local -r package_order_file="${package_path}/order"
+    local -a order=()
+
+    # Install packages by set order if file exists
+    if [ -f "${package_order_file}" ]; then
+        while read package; do
+            order+=("${package_path}/${package}")
+        done < "$package_order_file"
+    fi
+
+    # Add the rest of the packages to the order array
+    while read -r -d $'\0' package; do
+        # Match `order packages` with `package` to
+        # to avoid duplication in the array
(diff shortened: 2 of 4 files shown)


modules/deb.sh
```diff
@@ -221,6 +221,22 @@ function deb_get_and_install_package() {
...
+#  $1 - Directory path
+#
+function deb_get_and_install_input_package() {
+    if ! [[ $# -eq 1 ]]; then
+        log_fatal "deb_get_and_install_input_package() requires 1 argument"
+    fi
+    local package_path="$1"
+    local packages=$(ls ${package_path}/*.deb)
+
+    for package in $packages; do
```
lluiscampos, line 234: This will break if any files contains spaces. See [here](https://github.com/mendersoftware/monitor-client/blob/45ac1a1e4620a5a7cd2096fbd4d5503eedcff9d1/src/daemon.sh#L165) for an example of how we have solved this in other places and [here](https://mywiki.wooledge.org/BashPitfalls#for_f_in_.24.28ls_.2A.mp3.29) for a nice explanation of the issue (I am saving this page in my bookmarks for future me :smile: )

modules/deb.sh
```diff
@@ -221,6 +221,22 @@ function deb_get_and_install_package() {
...
+        log_info "Installing $package"
+        deb_install_package "${package}" "work/rootfs/"
```
lluiscampos, line 236: Installing the packages one by one has the limitation that if if package A depends on package B, it will break if we try to install package B first. We can solve this by passing all packages at ones to `dpkg --install`.

However how `deb_install_package` and related functions are designed they only accept one package at a time, so I won't make this a hard requirement, only something for you to consider. The task at hand is about mender-monitor and that will work fine.

:thinking: A simpler way to solve it would be for the user to pass a list of files rather than a directory, and then we assume that the user pass them in the correct order.

configs/mender_convert_config
```diff
@@ -157,6 +157,14 @@ MENDER_ADDON_CONFIGURE_INSTALL="n"
...
 MENDER_ADDON_CONFIGURE_VERSION="latest"
 
+# Directory containing deb packages to be installed in the image
+#
+MENDER_INSTALL_INPUT_PACKAGES_PATH="input/deb"
+
+# Install deb packages located in MENDER_INSTALL_INPUT_PACKAGES_PATH
+#
+MENDER_INSTALL_INPUT_PACKAGES="n"
+
```
lluiscampos, line 167: Thinking out loud: should we call these variables something else by Mender? Like `USER_INSTALL_INPUT_PACKAGES_PATH`, or just `INSTALL_PACKAGES_PATH`. The reason being that this is a method for install any package.

If so, we should also move it somewhere else that it is not next to the other MENDER variables.

lluiscampos (changes requested):
Simple and nice solution.

Requesting changes for the comment wrt the buggy for loop. The rest of the comments are suggestions for you to think.

configs/mender_convert_config
lluiscampos: Hi @fcervigni-EP and thanks for your input. Highly appreciated. I don't have an strong opinion, I was thinking out loud. It is totally fine with me to leave the `MENDER_` prefix. Do you agree @danielskinstad?

Overall I agree with your principle, that a tool should stick to the same prefix everywhere to avoid environment poisoning of any sort. However even in this file we have not been consistent, as they are other variables that don't have the prefix. Nevertheless it is a good idea to keep using `MENDER_` as much as possible in the new features :+1:

danielskinstad: Yes, I agree :) I will keep using `MENDER_`.

modules/deb.sh
danielskinstad: Let me know what you think about this solution; i.e. instead of changing the functions to take many packages at once (which, as you said, is unrelated to this task) I added the possibility of manually setting the order if needed. So the script checks for the existence of a `package_order` file in `input package path` and installs the files listed first in the selected order, then continues to install the rest. Otherwise it just continues as before.

modules/deb.sh
```diff
@@ -130,8 +130,8 @@ function deb_install_package_in_chroot() {
     run_in_chroot_and_log_cmd_noexit "$dest_dir" "env \
         DEBIAN_FRONTEND=noninteractive \
         DEVICE_TYPE=${MENDER_DEVICE_TYPE} \
-        dpkg --install /tmp/$(basename $deb_package)" || ret=$?
-    rm -f "$dest_dir/tmp/$(basename $deb_package)"
+        dpkg --install '/tmp/$(basename "$deb_package")'" || ret=$?
```
lluiscampos, line 133: You should use double quotes.

With single quotes, the `$(...` and `$deb...` are not expanded and just passed verbatim. This works only because at the end, in `run_and_log_cmd_noexit`, we have an `eval` command which will then expand.

modules/deb.sh
```diff
@@ -221,6 +221,45 @@ function deb_get_and_install_package() {
...
+    if ! [[ $# -eq 1 ]]; then
+        log_fatal "deb_get_and_install_input_package() requires 1 argument"
+    fi
+
+    local -r package_path="$1"
+    local -r package_order_file="${package_path}/package_order"
+    local -a order=()
+
+    # Install packages by set order if file exists
+    if [ -e "${package_order_file}" ]; then
```
lluiscampos, line 237: I would check for regular file instead
```suggestion
    if [ -f "${package_order_file}" ]; then
```

Otherwise you risk to break if is a directory for example

modules/deb.sh
lluiscampos: It is a good solution! Nice idea.

Two comments:

I would name the file only `order`. Alternatively I would use `series` which is a common convention for applying source code patches in a directory, for example.

Also, it should be documented in the `mender_convert_config` file. Your commit message is good, you can copy paste from there.

modules/deb.sh
lluiscampos: IMO this is a good fix and we should mark the commit as `fix`

lluiscampos (approved):
Looking good!

Few comments below but at this stage I trust you with the rework. Well done.

lluiscampos (approved):

PR #695, opened 2024-08-08
add support for Debian 12 "bookworm", 32bit and 64bit on Raspberry Pi 4

TheYoctoJester:
Preliminary PR for review and discussion!

Ticket references have to be added once finished, and commits will be signed then.

- [ ] Make sure that all commits follow the conventional commit [specification](https://www.github.com/mendersoftware/mendertesting/commitlint/grammar.md) for the Mender project.

the form below:

```
fix: <SHORT DESCRIPTION OF FIX>

<OPTIONAL LONGER DESCRIPTION>

Changelog: <USER-FRIENDLY-CHANGE-DESCRIPTION> or <None>
Ticket: <TICKET NUMBER> or <None>
```

### Description

Thank you!

### DIFF
--- a/configs/raspberrypi4_64_config
+++ b/configs/raspberrypi4_64_config
@@ -1,0 +1,3 @@
+MENDER_DEVICE_TYPE="raspberrypi4_64"
+
+source configs/raspberrypi_64_config
--- a/configs/raspberrypi_64_config
+++ b/configs/raspberrypi_64_config
@@ -1,0 +1,121 @@
+# Binaries generated with the following script:
+#    https://github.com/mendersoftware/mender-convert-integration-scripts/blob/master/build-uboot-rpi64.sh
+
+# in the Aarch64 case, it seems to always be "kernel8.img"
+RASPBERRYPI_KERNEL_IMAGE="kernel8.img"
+
+RASPBERRYPI_BINARIES="raspberrypi_arm64-2024.04.tar.gz"
+RASPBERRYPI_BINARIES_URL="${MENDER_STORAGE_URL}/mender-convert/uboot/raspberrypi64/${RASPBERRYPI_BINARIES}"
+
+RASPBERRYPI_OS_REQUIRED_ID="debian"     # also applies to the "Raspberry Pi OS" images
+
+RASPBERRYPI_OS_NEEDS_64BIT="y"
+
+source configs/include/raspberrypi_common_config
+
+function platform_modify() {
+    check_image_bitsize
+
+    if [ ${RASPBERRYPI_OS_REQUIRED_ID:-0} -ne 0 ] && ! check_image_os_id; then
+        echo "This configuration requires an image based on $RASPBERRYPI_OS_REQUIRED_ID."
+        exit 1
+    fi
+
+    if [ ${RASPBERRYPI_OS_MIN_VERSION:-0} -ne 0 ] && ! check_image_os_version_less_than ${RASPBERRYPI_OS_MIN_VERSION}; then
+        echo "This configuration requires an image based on $RASPBERRYPI_OS_REQUIRED_ID, version $RASPBERRYPI_OS_MIN_VERSION or later."
+        exit 1
+    fi
+
+    if [ ${RASPBERRYPI_OS_MAX_VERSION:-0} -ne 0 ] && ! check_image_os_version_greater_than ${RASPBERRYPI_OS_MAX_VERSION}; then
+        echo "This configuration requires an image based on $RASPBERRYPI_OS_REQUIRED_ID, version $RASPBERRYPI_OS_MAX_VERSION or later."
+        exit 1
+    fi
+
+    mkdir -p work/rpi/binaries
+
+    run_and_log_cmd "wget -q ${RASPBERRYPI_BINARIES_URL} -P work/rpi/binaries"
+    run_and_log_cmd "tar xzvf work/rpi/binaries/${RASPBERRYPI_BINARIES} -C work/rpi/binaries"
+
+    # By default, we modify cmdline.txt.
+    #
+    # In Ubuntu, the file is called btcmd.txt; if this file exists, use it instead
+    CMDLINE="cmdline.txt"
+
+    # Make a copy of Linux kernel arguments and modify.
+    run_and_log_cmd "cp work/boot/$CMDLINE work/rpi/$CMDLINE"
+
+    # Set a dynamic rootfs part (required for Mender A/B update strategy)
+    run_and_log_cmd "sed -i 's/\b[ ]root=[^ ]*/ root=\${mender_kernel_root}/' work/rpi/$CMDLINE"
+
+    # Root filesystem can not be resized when the disk is partition according
+    # to Mender layout, where the rootfs partition is the not last one which
+    # is a requirement to be able to do an "online" resize.
+    #
+    # This disables resize of rootfs on boot but applies the changes to
+    # cmdline.txt that are performed in the init_resize.sh script.
+    #
+    # Extracted from /usr/lib/raspi-config/init_resize.sh
+    run_and_log_cmd "sed -i 's| init=/usr/lib/raspi-config/init_resize\.sh||' work/rpi/$CMDLINE"
+    run_and_log_cmd "sed -i 's| sdhci\.debug_quirks2=4||' work/rpi/$CMDLINE"
+    if ! grep -q splash work/rpi/$CMDLINE; then
+        run_and_log_cmd "sed -i 's/ quiet//g' work/rpi/$CMDLINE"
+    fi
+
+    # Update Linux kernel command arguments with our custom configuration
+    run_and_log_cmd "sudo cp work/rpi/$CMDLINE work/boot/"
+
+    # Mask udisks2.service, otherwise it will mount the inactive part and we
+    # might write an update while it is mounted which often result in
+    # corruptions.
+    #
+    # TODO: Find a way to only blacklist mmcblk0pX devices instead of masking
+    # the service.
+    run_and_log_cmd "sudo ln -sf /dev/null work/rootfs/etc/systemd/system/udisks2.service"
+
+    # Extract Linux kernel and install to /boot directory. In the 64bit case, the previously existing kernel file
+    # is actually a gzipped vmlinux file. Hence, extract and copy as such
+    RASPBERRYPI_BOOTLOADER_IMAGE="${RASPBERRYPI_KERNEL_IMAGE}"
+    # this is required because gunzip is very picky about the filename!
+    run_and_log_cmd "sudo cp work/boot/${RASPBERRYPI_KERNEL_IMAGE} work/rootfs/boot/${RASPBERRYPI_KERNEL_IMAGE}.gz"
+    run_and_log_cmd "sudo gunzip work/rootfs/boot/${RASPBERRYPI_KERNEL_IMAGE}.gz"
+    run_and_log_cmd "sudo mv work/rootfs/boot/${RASPBERRYPI_KERNEL_IMAGE} work/rootfs/boot/vmlinux"
+
+    # Make sure the on-board bootloader boots our bootloader (which we do via the kernel option).
+    sed -i -e '/^kernel=/d' work/boot/config.txt
+    echo "kernel=${RASPBERRYPI_BOOTLOADER_IMAGE}" >> work/boot/config.txt
+
+    # Replace kernel with U-boot and add boot script
+    run_and_log_cmd "sudo cp work/rpi/binaries/u-boot.bin work/boot/${RASPBERRYPI_BOOTLOADER_IMAGE}"
+    run_and_log_cmd "sudo cp work/rpi/binaries/fw_env.config work/rootfs/etc/"
+    run_and_log_cmd "sudo cp work/rpi/binaries/uboot-git-log.txt work/boot"
+
+    # In order to have the DTBs inside the root filesystem and accessible from u-boot,
+    # copy all *.dtb files from the boot partition into the /boot directory
+    run_and_log_cmd "sudo cp work/boot/*.dtb work/rootfs/boot/"
+
+    # having the boot partition mounted at /uboot means that we have
+    # to adjust all (known) references to it accordingly
+    if [ "${RPI_FIXUP_UBOOT_PATH:-}" == "y" ]; then
+        log_info "Adjusting /uboot directory references"
+        run_and_log_cmd "sudo mkdir -p work/rootfs/uboot"
+        fixup_boot_part_path_references
+    fi
+
+    run_and_log_cmd "sudo install -m 755 work/rpi/binaries/fw_printenv work/rootfs/sbin/fw_printenv"
+    run_and_log_cmd "sudo ln -fs /sbin/fw_printenv work/rootfs/sbin/fw_setenv"
+
+    # Remove original 'resize2fs_once' script and its symbolic link.
+    if [ -L work/rootfs/etc/rc3.d/S01resize2fs_once ]; then
+        run_and_log_cmd "sudo unlink work/rootfs/etc/rc3.d/S01resize2fs_once"
+    fi
+    run_and_log_cmd "sudo rm -f work/rootfs/etc/init.d/resize2fs_once"
+
+    # Enable UART. This is required for U-Boot to operate correctly, see
+    # https://source.denx.de/u-boot/u-boot/-/blob/master/arch/arm/mach-bcm283x/Kconfig
+    log_info "Enabling UART in U-Boot configuration"
+    cat <<- EOF >> work/boot/config.txt
+
+# Enable UART
+enable_uart=1
+EOF
+}
--- a/configs/raspberrypi_64_config
+++ b/configs/raspberrypi_64_config
@@ -33,8 +33,7 @@ function platform_modify() {
 
     mkdir -p work/rpi/binaries
 
-    run_and_log_cmd "wget -q ${RASPBERRYPI_BINARIES_URL} -P work/rpi/binaries"
-    run_and_log_cmd "tar xzvf work/rpi/binaries/${RASPBERRYPI_BINARIES} -C work/rpi/binaries"
+    check_and_extract_asset $RASPBERRYPI_BINARIES $RASPBERRYPI_BINARIES_URL
 
     # By default, we modify cmdline.txt.
     #
--- a/configs/raspberrypi_config
+++ b/configs/raspberrypi_config
@@ -28,8 +28,7 @@ function platform_modify() {
 
     mkdir -p work/rpi/binaries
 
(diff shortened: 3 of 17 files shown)


configs/raspberrypi_common_config
```diff
@@ -0,0 +1,28 @@
...
+    return 0
+}
+
+# convenience wrapper
+function check_image_os() {
+    if ( ! check_image_os_id ) || ( ! check_image_os_minimal_version ); then
+        return 1
+    fi
+    return 0
+}
```
lluiscampos, line 46: Missing new line

assets/raspberrypi_arm64-2024.04.tar.gz
lluiscampos, line 1: I don't think we should commit the binaries to this repository. I know these are hardly going to change, but by principle they should be decoupled so that released versions of mender-convert could get newer bootloader binaries.

configs/raspberrypi_64_config
```diff
@@ -0,0 +1,111 @@
...
+        exit 1
+    fi
+    
+    mkdir -p work/rpi/binaries
+
+    if [ -f "assets/${RASPBERRYPI_BINARIES}" ]; then
+        run_and_log_cmd "cp assets/${RASPBERRYPI_BINARIES} work/rpi/binaries"
+    else
+        run_and_log_cmd "wget -q ${RASPBERRYPI_BINARIES_URL} -P work/rpi/binaries"
+    fi
```
lluiscampos, line 34: See my above comment. It we remove the binary this if/else is unnecessary.

lluiscampos (commented):
Few more comments on top of what Kristian already commented :eyes:

configs/raspberrypi_64_config
```diff
@@ -0,0 +1,111 @@
...
+
+    if ! check_image_os; then
+        echo "This configuration requires an image based on $RASPBERRYPI_OS_REQUIRED_ID, version $RASPBERRYPI_OS_MINIMAL_VERSION or later."
+        exit 1
+    fi
+    
+    mkdir -p work/rpi/binaries
+
+    if [ -f "assets/${RASPBERRYPI_BINARIES}" ]; then
+        run_and_log_cmd "cp assets/${RASPBERRYPI_BINARIES} work/rpi/binaries"
```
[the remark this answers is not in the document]
TheYoctoJester: Actually it will be removed, this is a tradeoff I decided to go for in order to provide a first usable branch to the community AND a preliminary PR for discussion. Once we have the binaries at a downloadable place, I think this should not be there - unless we decide to encourage people to inject their own binaries (which can be useful at times)

configs/raspberrypi_common_config
```diff
@@ -1,19 +1,37 @@
...
+# 4MB alignment
+MENDER_PARTITION_ALIGNMENT="4194304"
+
+# compress the images to save some space
+MENDER_COMPRESS_DISK_IMAGE="gzip"
+
+# from the "bookworm" release onwards, the Raspberry Pi OS
+# specific modifications expect the boot partition to be mounted
+# at /boot/firmware
+MENDER_BOOT_PART_MOUNTPOINT="/boot/firmware"
```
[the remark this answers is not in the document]
TheYoctoJester: The more I think about a combined configuration for Debian 11 and 12, the more I dislike it, sorry.

The current config for RPi mangles a lot of stuff into one place, like several RPi OS versions, Ubuntu, and it makes it very hard to understand what actually is supported, and what is not.

I would suggest to go the opposite direction: factor out the Debian 11 specific parts, and then create separate configurations. That might sound a bit much at first, but it excels at clarity and expectation management. Something like this structure, possibly:

```
configs/rpi/common
configs/rpi/bullseye/raspberrypi3_32
configs/rpi/bullseye/raspberrypi3_64
configs/rpi/bullseye/raspberrypi4_32
...
configs/rpi/bookworm/...
...
```

I think you get it. Thoughts? Also, @lluiscampos

assets/raspberrypi_arm64-2024.04.tar.gz
TheYoctoJester: Fully agreed. This is part of the "usable branch" vs. PR tradeoff. If there are no bigger concerns, I'll remove it for the polished PR.

configs/raspberrypi_common_config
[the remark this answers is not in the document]
kacf: Hmm, I see your point, but I'm not really thrilled to remove compatibility code. To me we are just shifting the responsibility to the user; at least in my mind mender-convert should ideally autodetect as much as possible.

As a middle ground, I would perhaps suggest: Yes, let's organize it (more or less) the way you proposed, but still use autodetection to include one file or the other. This still makes it pretty easy to remove the legacy file once those platforms are out of support.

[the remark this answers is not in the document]
lluiscampos: > As a middle ground, I would perhaps suggest: Yes, let's organize it (more or less) the way you proposed, but still use autodetection to include one file or the other. This still makes it pretty easy to remove the legacy file once those platforms are out of support.

I vote for this too :crossed_fingers:

janaperic:
Hi, I tested this PR using Raspberry Pi 4 and a 64bit [official Rasbian image](https://www.raspberrypi.com/software/operating-systems/#raspberry-pi-os-64-bit) (Lite Bookworm version released on 4th of July).

Here are the steps I took to test it: 

1. Convert the image ✅
Worked out of the box. The config I used was the custom one based on _raspberrypi4_64_config_, where I modified only the partition size.

2. Mender [integration checklist ](https://docs.mender.io/client-installation/integration-checklist) ✅
All of the steps from the checklist worked okay. 

3. Update the device from Mender hosted 
For testing this step, I tried 2 different artifacts. First, I connected my device (with mender-converted 64 bit image on it) to Mender hosted and tried to install the following artifacts: 
- mender-demo-artifact-3.75 ❌ 
failed with the following error: `No compatible busybox binary for this architecture.`
- another mender-converted image ❌ 
This image was converted from the same golden image (64 bit Rasbian Lite), using the same config file, only the release name was changed.
Deployment was skipped due to the following error: `No compatible artifact found`
The device type is: _raspberrypi4_
But my release is compatible with: _raspberrypi4_64_
Is there a way to change the device type to raspberrypi4_64 to make them compatible?

Let me know if I should take any additional steps to test this, or if any of my steps are wrong. 

Disclaimer: Because I'm on MacOS, the branch I am actually using is [this one](https://github.com/TheYoctoJester/mender-convert/tree/rpi64_wip_multiarch), but it should contain the same changes for 64 bit systems.

lluiscampos:
> Hi, I tested this (...)

Thanks @janaperic for testing!

>     * mender-demo-artifact-3.75 ❌
>       failed with the following error: `No compatible busybox binary for this architecture.`

Can you please report this in JIRA - we need to fix

>     * another mender-converted image ❌
>       This image was converted from the same golden image (64 bit Rasbian Lite), using the same config file, only the release name was changed.
>       Deployment was skipped due to the following error: `No compatible artifact found`
>       The device type is: _raspberrypi4_
>       But my release is compatible with: _raspberrypi4_64_
>       Is there a way to change the device type to raspberrypi4_64 to make them compatible?

The device type is specified when you create the Artifact. How did you create the artifact? If you used the Get started guide, pay attention in the [step 2](https://docs.mender.io/get-started/deploy-an-operating-system-update#step-2-setup-shell-variables-on-your-workstation) when you are supposed to specify the device type for your Artifact.

janaperic:
> Can you please report this in JIRA? We need to fix

Here's the [bug](https://northerntech.atlassian.net/browse/MEN-7473?atlOrigin=eyJpIjoiYjIwMTJjODYyZjBkNGFmMTkxZDQwYzZkM2NkM2RiNTQiLCJwIjoiaiJ9) I created on Jira. 


> The device type is specified when you create the Artifact. How did you create the artifact? If you used the Get started guide, pay attention in the [step 2](https://docs.mender.io/get-started/deploy-an-operating-system-update#step-2-setup-shell-variables-on-your-workstation) when you are supposed to specify the device type for your Artifact.

I created the artifact by using mender-convert on the golden image, following the steps for mender-convert. I did not use the snapshot feature. Is there a way to specify device type with mender-convert?

lluiscampos:
> I created the artifact by using mender-convert on the golden image (...)

I don't really understand the flow here. But if you used mender-convert both for the base image and the Artifact (the update) you should be having the same device type via mender-convet's configuration. It is set in this PR here: https://github.com/mendersoftware/mender-convert/pull/695/files#diff-2eeb0eea3372fbf12ef59af736e2b9b5b0cfb31b9cba397b3adb3cba39e59109R1 The variable is documented in [Mender Docs MENDER_DEVICE_TYPE](https://docs.mender.io/operating-system-updates-debian-family/variables#mender_device_type)

janaperic:
@lluiscampos Both of the images were created the same way - using the mender convert on the same image. The difference is that the first image was flashed directly to the card (so device type wasn't taken into account), but the second image was meant to be used from Mender hosted. 

I see one way this can be (easily?) resolved
- RPi 4 running 64 bit image needs to be recognized as raspberrypi4_64 (instead of raspberrypi4 like it is now) on Mender Hosted.

janaperic:
If you don't manage to track it down, let's have a meeting and I'll show you how to reproduce it :) @lluiscampos

configs/raspberrypi_common_config
```diff
@@ -0,0 +1,90 @@
...
```
lluiscampos, line 16: @TheYoctoJester This should be:
```suggestion
MENDER_BOOT_PART_MOUNT_LOCATION="/boot/firmware"
```

And, additionally, we need to set it in `/configs/mender_convert_config` to avoid the unbound variable error. Something like:
```
# User specified boot mount location. If left blank it uses `/boot/efi
# when `MENDER_GRUB_EFI_INTEGRATION` is set, otherwise `/uboot`.
MENDER_BOOT_PART_MOUNT_LOCATION=""
```

configs/raspberrypi_64_config
```diff
@@ -0,0 +1,107 @@
+# Binaries generated with the following script:
+#    https://github.com/mendersoftware/mender-convert-integration-scripts/blob/master/build-uboot-rpi64.sh
+
+ASSETS_PATH="rpi_arm64"
```
lluiscampos, line 4: Remember to remove these lines

lluiscampos (changes requested):
Hi @TheYoctoJester and @janaperic,

This sprint I am fixing the demo Artifact to work for Raspberry Pi 64 bits so I gave it an spin to this PR. Several problems:

1. Unbound variable

See comment below

2. Wrong packages being installed

This is my bad, but the tool was installing Debian Buster packages into the Bookworm based image. Fixed in https://github.com/mendersoftware/mender-convert/pull/696

3. Brick device after first update :boom: 

This is the most serious issue, of course. I am posting here the steps to put everyone in the same page and help debugging :point_down: 

To be able to run mender-convert I had to:
* fix the unbound error problem (see comment)
* build u-boot from https://github.com/TheYoctoJester/mender-convert-integration-scripts/tree/rpi_64bit with some patching:
```
diff --git a/build-uboot-rpi64.sh b/build-uboot-rpi64.sh
index 2a5f339..932007c 100755
--- a/build-uboot-rpi64.sh
+++ b/build-uboot-rpi64.sh
@@ -61,7 +61,7 @@ rm -rf uboot-mender
 
 echo "$rpi_board: getting source"
 
-git clone https://github.com/mendersoftware/uboot-mender.git -b mender-rpi-${UBOOT_MENDER_BRANCH}
+git clone https://github.com/TheYoctoJester/uboot-mender.git -b mender-rpi-${UBOOT_MENDER_BRANCH}
 cd uboot-mender
 
 echo "$rpi_board: compiling source"
```

* inject it into my image with something like:
```
--- a/configs/raspberrypi_64_config
+++ b/configs/raspberrypi_64_config
@@ -27,7 +27,8 @@ function platform_modify() {
     
     mkdir -p work/rpi/binaries
 
-    run_and_log_cmd "wget -q ${RASPBERRYPI_BINARIES_URL} -P work/rpi/binaries"
+    # run_and_log_cmd "wget -q ${RASPBERRYPI_BINARIES_URL} -P work/rpi/binaries"
+    run_and_log_cmd "cp assets/raspberrypi_arm64_signage-2024.04.tar.gz work/rpi/binaries/raspberrypi_arm64-2024.04.tar.gz"
     run_and_log_cmd "tar xzvf work/rpi/binaries/${RASPBERRYPI_BINARIES} -C work/rpi/binaries"
 
     # By default, we modify cmdline.txt.
```

With this I converted to images with the same config and input image: one with `MENDER_ARTIFACT_NAME="release-1"` from which I flashed the img into the SD card and the other with `MENDER_ARTIFACT_NAME="release-2"` from which I uploaded the Mender Artifact to the server.

And then, during the deployment of the update:
9A0C1CCF-FF1C-44B9-86A1-2C8BC622DAC1

@TheYoctoJester What have I messed up, then?

lluiscampos:
@TheYoctoJester I took the liberty of rebasing your branch on top of master and force-push to bring Mac ARM support (see #698) and facilitate this way the testing by @janaperic - I hope you don't mind :innocent:

janaperic:
I tested this PR again today. In the attached [document](https://docs.google.com/document/d/1dyN-wVtBrEE_VZiIQ5u5nAdLnKSn_6C1cIk7P_JX--M/edit?usp=sharing) I'm pasting the steps I took to convert & deploy the image.

TLDR: With some tweaks I was able to convert the image, but deployment from server wouldn't start because of device incompatibility. It seems to be the same issue I reported in this [comment](https://github.com/mendersoftware/mender-convert/pull/695#issuecomment-2293285391)

lluiscampos:
> I tested this PR again today. In the attached [document](https://docs.google.com/document/d/1dyN-wVtBrEE_VZiIQ5u5nAdLnKSn_6C1cIk7P_JX--M/edit?usp=sharing) I'm pasting the steps I took to convert & deploy the image.
> 
> TLDR: With some tweaks I was able to convert the image, but deployment from server wouldn't start because of device incompatibility. It seems to be the same issue I reported in this [comment](https://github.com/mendersoftware/mender-convert/pull/695#issuecomment-2293285391)

In your document, what does it mean "-- follow instructions on Mender Hosted" -> Is this executing the get.mender.io script? If so, there is the problem! With that you are re-configuring your device with HM defaults and that means `raspberrypi4` device type (among other undesired changes).

Assuming that `--overlay input/rootfs_overlay_demo/` was generated with the [Hosted Mender script](https://github.com/mendersoftware/mender-convert/blob/master/scripts/bootstrap-rootfs-overlay-hosted-server.sh), then you shouldn't reconfigure your device, the image is already configured to connect to Hosted Mender.

janaperic:
> Is this executing the get.mender.io script?

Yes, I was connecting the device this way instead of using overlays in mender-convert. At least now I know what's causing the issue. I'll try doing it again, the proper way.

tjhgit:
I tested the [documented](https://docs.mender.io/operating-system-updates-debian-family/convert-a-mender-debian-image) `mender-convert` for debian12 on x86_64. 
The partition layout seems to be correct and the system does boot. However `mender-authd` and `mender-updated` both point to /dev/null and are files of 0 size and masked.

@TheYoctoJester any ideas why this approach does not work?

configs/raspberrypi_config
```diff
@@ -1,28 +1,36 @@
...
-# Nothing to copy
-MENDER_COPY_BOOT_GAP=n
-
-# 4MB alignment
-MENDER_PARTITION_ALIGNMENT="4194304"
 
 RASPBERRYPI_BINARIES="${RASPBERRYPI_CONFIG}-2020.01.tar.gz"
 RASPBERRYPI_BINARIES_URL="${MENDER_STORAGE_URL}/mender-convert/uboot/raspberrypi/${RASPBERRYPI_BINARIES}"
 
+RASPBERRYPI_OS_REQUIRED_ID="raspbian" # historical distro name of the Raspberry Pi OS Debian derivative
```
gabrielSoudry, line 8: I have a bookworm image 32 bits with "debian" in os_release 
image

configs/raspberrypi_64_config
TheYoctoJester: @lluiscampos actually I'd vote for keeping the asset pulling logic, and just remove this `ASSETS_PATH` assignment to provide an easy way for users to inject their custom `u-boot` if required.

mender-convert-modify
```diff
@@ -186,10 +186,14 @@ run_and_log_cmd "echo -n ${VERSION_STRING} | sudo tee work/rootfs/etc/mender/scr
 
 log_info "Installing a custom /etc/fstab (see ${MENDER_CONVERT_LOG_FILE} for more info)"
 
-if [ "${MENDER_GRUB_EFI_INTEGRATION}" == "y" ]; then
-    boot_part_mountpoint="/boot/efi"
+if [ -n "${MENDER_BOOT_PART_MOUNT_LOCATION:-}" ]; then
+    boot_part_mountpoint="${MENDER_BOOT_PART_MOUNT_LOCATION}"
```
lluiscampos, line 190: Remember to drop this commit once we merge #705

configs/raspberrypi_common_config
lluiscampos: `Ticket: TBD` in this commit

configs/raspberrypi_common_config
lluiscampos: Has this been addressed?

configs/raspberrypi4_64_config
```diff
@@ -0,0 +1,3 @@
+MENDER_DEVICE_TYPE="raspberrypi4_64"
+
```
lluiscampos, line 2: `Ticket: TBD` also here

configs/raspberrypi_config
```diff
@@ -9,13 +9,12 @@ RASPBERRYPI_BINARIES_URL="${MENDER_STORAGE_URL}/mender-convert/uboot/raspberrypi
 RASPBERRYPI_OS_REQUIRED_ID="raspbian"   # historical distro name of the Raspberry Pi OS Debian derivative
 RASPBERRYPI_OS_MINIMAL_VERSION=12       # this is "bookworm" or later, as it changed the boot partition mount directory
 
+RPI_NEEDS_32BIT="y"
```
lluiscampos, line 9: Nit pick: for consistency with the above variables, use `RASPBERRYPI_OS_NEEDS_32BIT`

configs/raspberrypi_common_config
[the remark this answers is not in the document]
lluiscampos: The answer is yes :wink:

lluiscampos (approved):
Looking good :rocket: Pre-approving 

I have not tested it myself. I trust you and @janaperic on that.

Besides comments below, couple of extra things:
- [x] Remember to squash all the "fixup" commits
    * Even some of the other ones can be squashed, for example creating a function and later using it. (It is not very important)
- [ ] CI is complaining about bash format (in addition to commit linting)
- [x] Add the new configurations to the table in the README
- [x] Nice to have: add the Raspberry Pi bookworm 64 bit example in the REAME too

janaperic (approved):

PR #696, opened 2024-08-28
fix: Install correct packages for Debian 12 and add a warning

lluiscampos:
By the looks of it, this has been unmaintained for a while and although we introduced packages for new distros we were silently installing the Debian Buster ones.

Sort the distros from newer to older, and:

Changelog: Modify the list of available APT distros from where to get Mender software by adding `debian/bookworm` and `ubuntu/noble` and removing `ubuntu/focal`.

Changelog(chore): Fallback to Debian Bullseye packages when target distribution is not supported, printing a user facing warning when doing so.


janaperic (approved):

lluiscampos:
* 4.2.x

PR #698, opened 2024-08-28
feat: Multi-platform build for `mender-convert`

lluiscampos:
This commit changes several things with the end goal of making the tool
    multi-platform, namely easing the use of it on Apple M1 machines.
    
    In short:
    * Update base image to latest Ubuntu and update a dependency
    * Modify Dockerfile to build platform aware images with buildx
    * Switch pxz build image to Debian for easier cross-compile
    * Use specific version of `mender-artifact` instead of master. The
      motivation of this specific change is to make easier the install of
      pre-built binares via deb package.
    * Replace custom build/publish jobs for CI/CD with the ones from our
      templates, which are multi-platform ready
    * Demote the `docker-build` script for local development only.
    
    Changelog: `mender-convert` Docker build is now architecture aware and
    it will install the native dependencies for the target architecture.
    This change makes the tool compatible with Apple arm machines. The Docker
    images will be published for `linux/amd64` and `linux/arm64` platforms.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,14 +1,16 @@
 
 variables:
   DOCKER_REPOSITORY: mendersoftware/mender-convert
+  MULTIPLATFORM_BUILD: "true"
+  MULTIPLATFORM_PLATFORMS: "linux/amd64,linux/arm64"
   S3_BUCKET_NAME: mender-convert-images
 
   # These variables are present elsewhere in the repository too. Make sure to
   # search for and change them too.
   MENDER_ARTIFACT_VERSION:
-    value: "master"
+    value: "3.11.2"
     description: |-
-      Version of mender-artifact for the builder image - 'master' or a specific version
+      Version of mender-artifact for the builder image - a specific version
   MENDER_CLIENT_VERSION:
     value: "latest"
     description: |-
@@ -71,13 +73,12 @@ variables:
 
 include:
   - project: 'Northern.tech/Mender/mendertesting'
-    file: '.gitlab-ci-check-commits.yml'
-  - project: 'Northern.tech/Mender/mendertesting'
-    file: '.gitlab-ci-check-license.yml'
-  - project: 'Northern.tech/Mender/mendertesting'
-    file: '.gitlab-ci-github-status-updates.yml'
-  - project: 'Northern.tech/Mender/mendertesting'
-    file: '.gitlab-ci-check-shell-format.yml'
+    file:
+      - '.gitlab-ci-check-commits.yml'
+      - '.gitlab-ci-check-license.yml'
+      - '.gitlab-ci-github-status-updates.yml'
+      - '.gitlab-ci-check-shell-format.yml'
+      - '.gitlab-ci-check-docker-build.yml'
 
 stages:
   - test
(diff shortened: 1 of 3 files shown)


.gitlab-ci.yml
```diff
@@ -1,14 +1,16 @@
...
   S3_BUCKET_NAME: mender-convert-images
 
   # These variables are present elsewhere in the repository too. Make sure to
   # search for and change them too.
   MENDER_ARTIFACT_VERSION:
-    value: "master"
+    value: "3.11.2"
     description: |-
-      Version of mender-artifact for the builder image - 'master' or a specific version
+      Version of mender-artifact for the builder image - an specific version
```
oldgiova, line 13: Not my strongest field, but it looks like a typo ;) 
```suggestion
      Version of mender-artifact for the builder image - a specific version
```

oldgiova (approved):
LGTM!

lluiscampos:
> LGTM!

Thanks! I'll wait for @janaperic to fully test it on a Mac ARM machine :wrench:

janaperic:
I tried testing on my MacBook Pro M3. When I run ./docker-build I get the following error: 

 ```
=> ERROR [ 7/13] RUN go get github.com/mendersoftware/mender              0.1s 
------                                                                          
 > [ 7/13] RUN go get github.com/mendersoftware/mender:
0.110 rosetta error: failed to open elf at /lib64/ld-linux-x86-64.so.2
0.110  Trace/breakpoint trap
------

 2 warnings found (use --debug to expand):
 - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 61)
 - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 62)
Dockerfile:64
--------------------
  62 |     ENV GOPATH "/root/go"
  63 |     
  64 | >>> RUN go get github.com/mendersoftware/mender
  65 |     WORKDIR $GOPATH/src/github.com/mendersoftware/mender
  66 |     RUN git checkout $MENDER_CLIENT_VERSION
--------------------
ERROR: failed to solve: process "/bin/sh -c go get github.com/mendersoftware/mender" did not complete successfully: exit code: 133

```

If I remember correctly, Josef had solved this issue by building mender-artifact instead of pulling it? Or something similar. It was solved on his rpi64_wip_multiarpch branch

**Comment by janaperic** [comment]
> Can you try disable the Rosetta emulator for this build?

With rosetta disabled I get a similar error: 
```
------                                                                                                                     
 > [ 7/13] RUN go get github.com/mendersoftware/mender:
0.102 qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory
------
```

**Comment by lluiscampos** [comment]
@janaperic I think you are testing the wrong branch, there is no `go get` anywhere in here.

To make sure you test what we are reviewing you can directly check-out the PR like:
```
git fetch origin pull/698/head
git checkout FETCH_HEAD
./docker-build
```

Alternatively you can add my fork to your remotes and check the branch. But the above method will make sure you are up to date with whatever GitHub has in the PR 698 (this one)

**Comment by mzedel** [comment]
as a fellow Mac ARM user I tried building it and it worked fine with 
```
docker build -t mendersoftware/mender-convert:pr --build-arg MENDER_ARTIFACT_VERSION=3.11.2 .
```
using docker Version 4.33.0 (160616)

File: Dockerfile [hunk 2]
```diff
@@ -66,8 +69,9 @@ COPY --from=build /root/pxz/pxz /usr/bin/pxz
 RUN echo "Defaults        secure_path=\"/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin:$PATH\"" > /etc/sudoers.d/secure_path_override
 RUN chmod 0440 /etc/sudoers.d/secure_path_override
 
-RUN wget -q -O /usr/bin/mender-artifact https://downloads.mender.io/mender-artifact/$MENDER_ARTIFACT_VERSION/linux/mender-artifact \
-    && chmod +x /usr/bin/mender-artifact
+RUN deb_filename=mender-artifact_${MENDER_ARTIFACT_VERSION}-1%2Bubuntu%2Bnoble_${TARGETARCH}.deb && \
+    wget "https://downloads.mender.io/repos/debian/pool/main/m/mender-artifact/${deb_filename}" \
+    --output-document=/mender-artifact.deb && dpkg-deb --extract /mender-artifact.deb /
```
lluiscampos, line 74: Note to self: install the package instead of just extracting it - we are on a Ubuntu image after all!

janaperic:
@lluiscampos you were right, I was testing the wrong thing. 
Successfully converted the image now! Go ahead and merge it :)

janaperic (approved):

Dockerfile
```diff
@@ -9,9 +10,10 @@ RUN apt-get update && env DEBIAN_FRONTEND=noninteractive apt-get install -y \
 RUN git clone https://github.com/jnovy/pxz.git /root/pxz
 RUN cd /root/pxz && make
```
lluiscampos, line 11: Note to self: this is not correctly cross-compiled on CI, the final binary is x86. I will _try_ to fix but considering that this is pure value-add if I don't find an easy solution I'll switch this job to also emulate instead of cross-compile (slower, but only on CI)

lluiscampos: Fixed in the last force-push, can @mzedel or @janaperic try the following?
```
docker run --rm --entrypoint /usr/bin/pxz registry.gitlab.com/northern.tech/mender/mender-convert:1432652685 fake-file
```

:pray:

mzedel: results in:
```
/usr/bin/pxz: can't stat 'fake-file': No such file or directory
```
is this good?

lluiscampos: That is good! :partying_face:

PR #707, opened 2024-10-25
fix: Change git_tag_or_branch command

estape11:
When multiple branches where pointing to the same reference, it will mess up the docker command as it will list not only but multiple branches names, causing the docker run command to fail.

### DIFF
--- a/docker-mender-convert
+++ b/docker-mender-convert
@@ -17,7 +17,7 @@
 set -e
 
 GIT_TAG_OR_BRANCH_NAME="$(git describe --exact HEAD 2>/dev/null || \
-    git for-each-ref "refs/heads/*" --format '%(refname:short)' --points-at HEAD 2>/dev/null)"
+    git for-each-ref "refs/heads/*" --format '%(refname:short)' --points-at HEAD 2>/dev/null | head -n1)"
 
 if [ -z "$IMAGE_NAME" -a -z "$GIT_TAG_OR_BRANCH_NAME" ]; then
   echo "Could not deduce mendersoftware/mender-convert container version from currently checked out commit. Using latest." 1>&2


docker-mender-convert
```diff
@@ -16,8 +16,8 @@
 
 set -e
 
-GIT_TAG_OR_BRANCH_NAME="$(git describe --exact HEAD 2>/dev/null || \
-    git for-each-ref "refs/heads/*" --format '%(refname:short)' --points-at HEAD 2>/dev/null)"
+GIT_TAG_OR_BRANCH_NAME="$(git describe --tags --exact-match 2>/dev/null || \
+    git branch --show current 2>/dev/null)"
```
lluiscampos, line 20: The downside of this fallback (`git branch` instead of `git for-each-ref "refs/heads/*"`) is that it won't work on detached HEAD - which is what I suspect happens in CI and motivated this piece of code in the place.

What about leaving it as it was but cutting the output to one line?
```suggestion
GIT_TAG_OR_BRANCH_NAME="$(git describe --exact HEAD 2>/dev/null || \
    git for-each-ref "refs/heads/*" --format '%(refname:short)' --points-at HEAD 2>/dev/null | head -n1)"
```

It is still imperfect, because it will pick the first branch in alphanumerical order always (a Git branch has no date information so it cannot be sorted by date, for example). But at least it won't break as you have experienced.

lluiscampos (changes requested):
Looking good! Only formalities missing:

* Squash your commits into one
* Please make a user friendly changelog entry. What kind of information gives to a user the following changelog: "Change git_tag_or_branch command" :thinking: Not much, if you ask me. I would do an entry like: "Fix an issue in the `docker-mender-convert` wrapper that arises when multiple branches where pointing to the same reference in the Git checkout of the tool" (you don't need to use `Changelog: Title`, but you can just write all that in the commit like `Changelog: bla bla bla bla`)

estape11:
Hey @lluiscampos I updated the commit with your suggestion. Thank you for your input :D

danielskinstad:
The script that checks the commits doesn't seem to like the whitespace in the `Changelog` entry. You'll make the script pass if you remove the whitespaces and add a newline above `Ticket`. In addition you should rebase your commit on top of master :) Btw, you can check if your commit passes locally by running this [script](https://github.com/mendersoftware/mendertesting/blob/master/check_commits.sh) from `mendertesting`

lluiscampos (approved):

lluiscampos:
* 4.2.x

PR #709, opened 2024-10-31
fix comparison issue that breaks platform hooks

SebSa:
fix: comparison issue that breaks platform hooks

A type comparison issue in line 19 the platform_modify function on the latest 64 bit changes yields:
```shell
2024-10-31 13:22:38 [INFO] [mender-convert-modify] Running hook: platform_modify
configs/raspberrypi_64_config: line 19: [: debian: integer expression expected
```

have fixed that with a suggestion from @danielskinstad 

Changelog: None
Ticket: None

### DIFF
--- a/configs/raspberrypi_64_config
+++ b/configs/raspberrypi_64_config
@@ -16,7 +16,7 @@ source configs/include/raspberrypi_common_config
 function platform_modify() {
     check_image_bitsize
 
-    if [ ${RASPBERRYPI_OS_REQUIRED_ID:-0} -ne 0 ] && ! check_image_os_id; then
+    if [ -n "${RASPBERRYPI_OS_REQUIRED_ID:+x}" ] && ! check_image_os_id; then
         echo "This configuration requires an image based on $RASPBERRYPI_OS_REQUIRED_ID."
         exit 1
     fi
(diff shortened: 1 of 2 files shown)


danielskinstad (changes requested):
Hey, thanks for your contribution!

Please take a look at this: https://github.com/mendersoftware/mendertesting/blob/master/commitlint/grammar.md

A fix requires a changelog and a ticket (can be set to none).

See comment below.

danielskinstad:
The current pipeline will fail because the commit is not signed off.

`Commit 19ed07227f1132071542ef92db6e1e50ddbc2a85 is not signed off! Use --signoff with your commit. Make sure that the Author of the commit matches the one in Signed-off-by`

danielskinstad:
Thanks a lot, it looks good

I just saw that we have the same issue in `raspberrypi_config` as well, I'll be happy to co-author the commit and add it myself, but you can also do it if you want :)

danielskinstad:
> I've added this now, is there anything else I need to do?

Thanks for adding it! Can you rebase it on top of master? 
The only thing is the commit; while you're rebasing, could you also change the title to something like
`fix: correctly check that RASPBERRYPI_OS_REQUIRED_ID is set`
 and set 
`Changelog: Title` 
 so we create a changelog entry for this fix?

danielskinstad (approved):

PR #712, opened 2024-11-07
Reorganize RPi files

ghost:
```
commit fd1692e75cdd6a8cd12c48feb27cbd564910bdba
Date:   Tue Nov 5 14:49:52 2024

    chore: Remove config that doesn't belong the raspberrypi files.
    
    It's already the default, and is not really related to Raspberry Pi.
    
commit 90ffa630874ac1b7da16efc200f96ff4694393b5
Date:   Thu Nov 7 10:29:06 2024

    refac: Reorganize RPi configs and put into sub folders.
    
    Changelog: Raspberry Pi configs have changed names. Please see the
    `README.md` files for the new names.
    
    Ticket: None
    
commit 84ac1d2374d977fb111dfde1ce3bb15f56c2a3fc
Date:   Tue Nov 5 08:23:07 2024

    chore: Move versioned URL into config file.
    
    Gives a bit more visibility than having it inline in the code.
    
```

### DIFF
--- a/configs/mender_convert_config
+++ b/configs/mender_convert_config
@@ -303,6 +303,9 @@ IMAGE_ROOTFS_EXTRA_SPACE="0"
 # why it is missing a "MENDER_" prefix.
 IMAGE_OVERHEAD_FACTOR="1.5"
 
+MENDER_CLIENT_DATA_DIR_SERVICE_URL="https://raw.githubusercontent.com/mendersoftware/\
+meta-mender/ab9c1e65abf73cde1065e2a4c2005cfe61ee6e3d/meta-mender-core/recipes-mender/mender-client/files/mender-data-dir.service"
+
 source configs/mender_grub_config
 
 # Function to create Mender Artifact
--- a/mender-convert-modify
+++ b/mender-convert-modify
@@ -136,9 +136,6 @@ if [ "${MENDER_GRUB_EFI_INTEGRATION}" == "y" ]; then
 
 fi
 
-MENDER_CLIENT_DATA_DIR_SERVICE_URL="https://raw.githubusercontent.com/mendersoftware/\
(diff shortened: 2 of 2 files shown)


kacf:
@TheYoctoJester: Almost all of this is just moving code around, so I highly recommend viewing locally with `--color-moved`. Makes it much easier to read. 🙂

kacf:
@lluiscampos: I'm mainly pinging you for the `.gitlab-ci.yml` part.

kacf:
> Nitpick: I would call it `RASPBERRYPI_BASE_CONFIG` (as opposed to `RASPBERRYPI_BARE_CONFIG`), but that's just me maybe.

I thought about this when I first named it. I was trying to convey that it's not a "smaller" or "different" config, just the same config expressed as a bare name. But I dunno... Let's see what Lluis says, if he has the same opinion, I'll change it.

lluiscampos (approved):
I also think that `_BASE_` is better.

The CI file looks good :+1:

lluiscampos:
> (...) I was trying to convey that it's not a "smaller" or "different" config, (...)

Right, but we use the `basename` command to get it :thinking:

kacf:
> > (...) I was trying to convey that it's not a "smaller" or "different" config, (...)
> 
> Right, but we use the `basename` command to get it 🤔

Yes, but `basename` does produce something smaller, out of a larger path.

Anyway, we're splitting hairs here, the gut feeling of two people is enough to change it! 😄

lluiscampos:
> Yes, but `basename` does produce something smaller, out of a larger path.

What if we use `RASPBERRYPI_CONFIG_NAME` then?

(I see you already force-pushed, so likely not worth a second rename :sweat_smile: )

kacf:
> > Yes, but `basename` does produce something smaller, out of a larger path.
> 
> What if we use `RASPBERRYPI_CONFIG_NAME` then?
> 
> (I see you already force-pushed, so likely not worth a second rename 😅 )

I like it, it's more distinctive. Sorry, Gitlab, I have to bother you some more then!

PR #719, opened 2024-11-16
ci: Fix also the `rm` command in the publish job

lluiscampos:
Same reasoning as for having wildcards for copying the files: we only know for sure the name of the tarball and the directory, so remove these instead.

Amends 9630b5e, c235e30, and the rest of the chain.


lluiscampos:
@oldgiova Yet another fix... Sorry for launching so many. The issue is that it can only be tested once it hits master, so my latest fixes not been tested in CI.

This should be the last one :crossed_fingers:

oldgiova (approved):

oldgiova:
> @oldgiova Yet another fix... Sorry for launching so many. The issue is that it can only be tested once it hits master, so my latest fixes not been tested in CI.
> 
> This should be the last one :crossed_fingers: 

No worries!

PR #721, opened 2024-11-19
MEN-7753: Binfmt fixes

kacf:


scripts/test/run-tests.sh
```diff
@@ -191,4 +198,9 @@ if [ "$TEST_ALL" == "1" -o "$TEST_PLATFORM" == "debian-qemux86-64" ]; then
                    || test_result=$?
 fi
 
+if [ "$MATCHED_A_TEST" = 0 ]; then
+  echo "No test matched!" 1>&2
```
lluiscampos, line 187: Thank you :pray:

scripts/test/run-tests.sh
```diff
@@ -151,30 +151,15 @@ if [ "$TEST_ALL" == "1" -o "$TEST_PLATFORM" == "ubuntu-qemux86-64-no-grub-d" ];
                    || test_result=$?
 fi
 
-if [ "$TEST_ALL" == "1" -o "$TEST_PLATFORM" == "raspberrypi3" ]; then
```
lluiscampos: Actually, there has been a long internal discussion and we will restore raspberrypi3 this sprint. This is independent of your PR and I am okey merging yours as-is (because, well, at the time we are not building the image)

scripts/test/run-tests.sh
```diff
@@ -151,30 +151,15 @@ if [ "$TEST_ALL" == "1" -o "$TEST_PLATFORM" == "ubuntu-qemux86-64-no-grub-d" ];
...
-                   "input/image/${RASPBIAN_IMAGE_COMPRESSED}" \
-                   "--config configs/raspberrypi3_config $EXTRA_CONFIG" \
-                   "--" \
-                   "$@" \
-                   || test_result=$?
-fi
-
-if [ "$TEST_ALL" == "1" -o "$TEST_PLATFORM" == "ubuntu-raspberrypi3" ]; then
-  MATCHED_A_TEST=1
-  wget --progress=dot:giga -N ${UBUNTU_SERVER_RPI_IMAGE_URL} -P input/image/
```
lluiscampos: This one is good to fully remove. Also please remove `UBUNTU_SERVER_RPI_IMAGE_URL` declaration above

lluiscampos (approved):
Very tricky issue! Thank you for the good comment in the code. And how did you figure it out?

kacf:
> Very tricky issue! Thank you for the good comment in the code. And how did you figure it out?

A lot if experimentation. The final result is a one liner, but it took more than a day to figure out, and a lot of experimenting on the build host. The main clue for me was comparing the installed files of the `qemu-user-static` package (which is package that installs the mappings) between ubuntu 22 and 24. I could see there were different file formats and different install locations, which eventually led me to the discovery that systemd-binfmt has taken over the task of binfmt-support tools, using different configuration files. One of those tools is the one that communicates the mappings to the kernel, through that special binfmt_misc filesystem.

scripts/test/run-tests.sh
kacf: Good call, done!

kacf:
> Very tricky issue! Thank you for the good comment in the code. And how did you figure it out?

This was the error message btw:

```
sudo chroot work/rootfs /tmp/qemu-aarch64-static /usr/bin/env apt update
/usr/bin/env: ‘apt’: Exec format error
```

I did suspect that binfmt portion of the code, since I was the one to put it there in the first place. So I guess I was primed with that knowledge.

PR #722, opened 2024-11-21
feat: create sparse images

ygerlach:

### Description

Instead of allocating the whole image on the disk, only allocate the parts that actually contain data. This reduces the required disk space to store or create images. The image does appear to have the normal size but if inspected with `du` or `ls -lsh` the difference becomes apparent.

Example with the image i build:
```
$ ls -lsh
total 244G
234G -rw-r--r-- 1 root root 235G Nov 20 17:30 image-x86_64-mender_old.img
 11G -rw-r--r-- 2 root root 235G Nov 20 17:54 image-x86_64-mender_sparse.img
```

This is done by the `conv=sparse` flag of dd. Alternatively a file can made sparse with `fallocate -d <file>`. But this may only collapse a file, not reduce its footprint while it is created. This may also improve read and write performance of those files in the empty areas, but i did not observe a big speed-up during the image creation.

### DIFF
--- a/configs/debian-qemux86-64_config
+++ b/configs/debian-qemux86-64_config
@@ -19,3 +19,5 @@ MENDER_DEVICE_TYPE="qemux86-64"
 
 # Nothing to copy
 MENDER_COPY_BOOT_GAP="n"
+
+MENDER_SPARSE_IMAGE="y"
--- a/configs/mender_convert_config
+++ b/configs/mender_convert_config
@@ -194,6 +194,14 @@ MENDER_SNAPSHOT_INSTALL="auto"
 # Valid values are "latest" (default), "master" or a specific version
 MENDER_SNAPSHOT_VERSION="latest"
 
+# Mender sparse image file generation
+#
+# By default, mender-convert writes the entire image. When enabled, mender-convert attempts
+# to avoid allocating empty spaces within the image. This can result in the created image
+# using less space on the storage medium while maintaining the expected size.
+# However, some tools may not handle sparse files correctly.
+MENDER_SPARSE_IMAGE="n"
+
 # File storage, containing binary files, do not modify this unless you know
 # what you are doing.
 MENDER_STORAGE_URL="https://downloads.mender.io"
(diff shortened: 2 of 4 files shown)


lluiscampos:
Hi @ygerlach and thank you for your contribution.

How much have you tested this?

We have been debating internally because we used to have sparse images in our yocto integrations but had problems wrt calculating the size and other parts of the integration would break. So I am a bit concern on merging it as the automated testing that we have on mender-convert won't catch all the issues that arise on real devices later.

Being an optimization, an alternative could be to have this as an opt-in feature controlled by a configuration parameter - then a user can have it on for his/her integration but we keep providing a default value that we think will work for most or the users. What do you think?

ygerlach:
> How much have you tested this?

not very much. I have made my image and copied it onto a normal disk using `dd`. I have not used any special hardware or software.

> Being an optimization, an alternative could be to have this as an opt-in feature controlled by a configuration parameter 

If other tools have problems with sparse files, that sounds like a bug in those tools. But i guess adding it as optional flag might be the best curse of action, until those bugs are sorted out. After some time we can get back to this and decide if we want to move it to enabled by default.
I am going to make a second commit proposing a possible implementation with the option disabled by default.

In a third commit i would enable it in some of the example configs like (configs/debian-qemux86-64_config). Are there some configs that i should avoid?

lluiscampos (changes requested):
This looks great!

> I am going to make a second commit proposing a possible implementation with the option disabled by default.

Please squash these two commits together, there is no much value in having the separately (specially if both changelog entries are so tightly related). You could then elaborate a bit more the changelog. Example:

"Changelog: Support for sparse images to save storage space. It is disabled by default and can be enabled with `MENDER_SPARSE_IMAGE` configuration option"

> 
> In a third commit i would enable it in some of the example configs like (configs/debian-qemux86-64_config). Are there some configs that i should avoid?

Good idea, please use `configs/ubuntu-qemux86-64_config` and `configs/debian-qemux86-64_config` :+1:

lluiscampos:
@ygerlach Would you follow-up documenting the new variable in [our docs page](https://github.com/mendersoftware/mender-docs/blob/master/04.Operating-System-updates-Debian-family/99.Variables/docs.md)? :innocent:

lluiscampos:
* 4.3.x

PR #723, opened 2024-11-25
MEN-7759: Upgrade Raspberry Pi 3 pre-built image to Debian 12 64bits

lluiscampos:
This commit restores the Raspberry Pi 3 pre-built image build and publish, but modifying it from being based in the 32bits one to the 64 bits one.

As the commit that removed it completely did not have an explicit changelog entry on this, the changelog here is only about "upgrading" the image.


.gitlab-ci.yml
```diff
@@ -376,11 +336,7 @@ test_acceptance_raspberrypi4_bookworm_64bit:
   needs:
     - job: convert_raspbian_raspberrypi4_bookworm_64bit
       artifacts: true
-    - job: convert_raspbian_raspberrypi4_bookworm_32bit
-      artifacts: true
-    - job: convert_raspbian_raspberrypi4_bullseye_64bit
-      artifacts: true
-    - job: convert_raspbian_raspberrypi4_bullseye_32bit
+    - job: convert_raspbian_raspberrypi3_bookworm_64bit
```
danielskinstad, line 339: The loop at line 348 should also include `raspberrypi3_bookworm_64bit_config`, right?

danielskinstad (approved):
Nice, looks good to me

lluiscampos:
* 4.3.x

PR #746, opened 2025-03-07
fix: Fix writing MSDOS partition UUIDs

jo-lund:
rev works on unicode strings and is therefore not safe to use to swap bytes.

Ticket: MEN-7937
Changelog: Title


jo-lund:
@mender-test-bot start client pipeline

danielskinstad (approved):
I've tested it a bit, and the fix seems correct to me.

PR #750, opened 2025-03-31
fix: Expand the data partition into the empty space on boot

jo-lund:
Make sure that the data partition is expanded into the remaining space on boot for all systems.

Ticket: MEN-8191
Changelog: Title

### DIFF
--- a/configs/systemd_common_config
+++ b/configs/systemd_common_config
@@ -1,0 +1,32 @@
+#
+# Resize the data partition to fill the remaining space, using parted, with systemd
+#
+function grow_data_partition() {
+    log_info "Adding systemd init script to run parted and resize the data partition on boot"
+    log_info "to fill all the available space on the storage media"
+    run_and_log_cmd "sudo mkdir -p work/rootfs/etc/systemd/system/data.mount.wants/"
+    cat <<- EOF > work/rootfs/etc/systemd/system/mender-grow-data.service
+		  [Unit]
+		  Description=Mender service to grow data partition size
+		  DefaultDependencies=no
+		  Before=data.mount
+		  Before=systemd-growfs@data.service
+
+		  [Service]
+		  Type=oneshot
+		  User=root
+		  Group=root
+		  ExecStart=/sbin/parted --script $(disk_get_device_base "${MENDER_STORAGE_DEVICE_BASE}") resizepart ${MENDER_DATA_PART_NUMBER} 100%
+
+		  [Install]
+		  WantedBy=data.mount
+	EOF
+
+    # Install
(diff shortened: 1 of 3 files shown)


jo-lund:
@mender-test-bot start client pipeline

configs/systemd_common_config
```diff
@@ -0,0 +1,35 @@
...
+		  Group=root
+		  ExecStart=/sbin/parted --script $(disk_get_device_base "${MENDER_STORAGE_DEVICE_BASE}") resizepart ${MENDER_DATA_PART_NUMBER} 100%
+
+		  [Install]
+		  WantedBy=data.mount
+	EOF
+
+    # Install
+    run_and_log_cmd "cp work/rpi/etc/systemd/system/mender-grow-data.service \
+                      work/rootfs/etc/systemd/system/"
```
lluiscampos, line 28: Why was this done in `/rpi/` and then copied over? If the `/rpi/` is a vendor specific requirement for Raspberry Pi images, we should keep it for such boards but in any case removed from the generic.

lluiscampos (commented):
Aha, so we had the functionality in mender-convert but only for a subset of the boards. Thanks.

In my opinion this is a feature and not a fix. As a user I wouldn't expect such a behavior change on a bugfix release.

Please let's elaborate a bit more the changelog entry: is the feature enabled by default? how to disable it? I think for such a feature/fix the title is just not enough.

Finally, see my comment below: we should not pollute the generic scripts with `/rpi/..` (which, to be honest, I don't know why is there in the first place)

jo-lund:
> Aha, so we had the functionality in mender-convert but only for a subset of the boards. Thanks.
> 
> In my opinion this is a feature and not a fix. As a user I wouldn't expect such a behavior change on a bugfix release.

The point here is that if MENDER_DATA_PART_GROWFS is set to true (which it is by default), the expected behaviour should be to expand the partition. This was not done, however, and in my view this is then a bug. In addition the raspberrypi config didn't even check if MENDER_DATA_PART_GROWFS was set, but added the service in all cases.

> Please let's elaborate a bit more the changelog entry: is the feature enabled by default? how to disable it? I think for such a feature/fix the title is just not enough.
> 

I will add some more details to this

> Finally, see my comment below: we should not pollute the generic scripts with `/rpi/..` (which, to be honest, I don't know why is there in the first place)

I used what we already had for the raspberrypi and assumed this worked well and had been tested properly. A quick test confirmed that the data partition was expanded. I agree, however, that this should be changed and is not correct, besides doing unnecessary copying.

lluiscampos (approved):
Thank you for the f2f discussion. This looks very good! Great changelog entry :medal_sports:

PR #755, opened 2025-04-22
Fix: mender grow fs service

tactmaster:
Changelog: Title

Mender grow fs service not failure 
Service got blank device to work on with parted.
Just need MENDER_STORAGE_DEVICE_BASE not disk_get_device_base

Ticket:  None

### DIFF
--- a/configs/systemd_common_config
+++ b/configs/systemd_common_config
@@ -16,7 +16,7 @@ function grow_data_partition() {
 		  Type=oneshot
 		  User=root
 		  Group=root
-		  ExecStart=/sbin/parted --script $(disk_get_device_base "${MENDER_STORAGE_DEVICE_BASE}") resizepart ${MENDER_DATA_PART_NUMBER} 100%
+		  ExecStart=/sbin/parted ${MENDER_STORAGE_DEVICE_BASE} --script  resizepart ${MENDER_DATA_PART_NUMBER} 100%
 
 		  [Install]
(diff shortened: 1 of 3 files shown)


configs/systemd_common_config
```diff
@@ -16,7 +16,7 @@ function grow_data_partition() {
 		  Type=oneshot
 		  User=root
 		  Group=root
-		  ExecStart=/sbin/parted --script $(disk_get_device_base "${MENDER_STORAGE_DEVICE_BASE}") resizepart ${MENDER_DATA_PART_NUMBER} 100%
+		  ExecStart=/sbin/parted ${MENDER_STORAGE_DEVICE_BASE} --script  resizepart ${MENDER_DATA_PART_NUMBER} 100%
```
jo-lund, line 20: mender_convert_config will by default set MENDER_STORAGE_DEVICE_BASE to /dev/mmcblk0p and this will not work with parted. The correct fix here should be to remove the "p" suffix (and partition number) if it exists, else just use the MENDER_STORAGE_DEVICE_BASE as is.

tactmaster: So the problem here is the "p"
as with /dev/sda has no p.

So this may be alternative
https://github.com/mendersoftware/mender-convert/pull/757

Haven't tested it yet

May also need to fix the order of the parted command 
https://wiki.archlinux.org/title/Parted

PR #756, opened 2025-04-25
test: Change owner and permissions of ssh keys after copy

lluiscampos:
Modifying directly the Git checked out files had the annoying inconvenience that block any local operations with Git after running the tests.

### DIFF
--- a/scripts/test/test-utils.sh
+++ b/scripts/test/test-utils.sh
@@ -175,19 +175,24 @@ get_pytest_files() {
 }
 
 prepare_ssh_keys() {
-  if [ "$(stat -c %U tests/ssh-public-key-overlay/root)" != "root" ]; then
-    sudo chown -R root:root tests/ssh-public-key-overlay/root
+  # Prepare an overlay with the public key
+  mkdir --parents input/tests
+  sudo cp --recursive "tests/ssh-public-key-overlay" "input/tests/"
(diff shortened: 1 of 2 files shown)


scripts/test/test-utils.sh
```diff
@@ -175,19 +175,24 @@ get_pytest_files() {
...
-  if [ "$(stat -c %a tests/ssh-public-key-overlay/root/.ssh/authorized_keys)" != "755" ]; then
-    sudo chmod 600 tests/ssh-public-key-overlay/root/.ssh/authorized_keys
+  if [ "$(sudo stat -c %a input/tests/ssh-public-key-overlay/root/.ssh/authorized_keys)" != "755" ]; then
+    sudo chmod 600 input/tests/ssh-public-key-overlay/root/.ssh/authorized_keys
   fi
+
+  # Set also permissions for private key
   if [ "$(stat -c %a tests/ssh-priv-key/key)" != "600" ]; then
-    sudo chmod 600 tests/ssh-priv-key/key
+    chmod 600 tests/ssh-priv-key/key
```
danielskinstad, line 196: Is there a reason to not do sudo here?

lluiscampos: The private key does not need to be owned by root. It is owned by whoever checkout the repo and runs the tests (locally it would be `lluis`, in a container it would actually be root but you don't need `sudo` if you are root already)

danielskinstad (approved):
Forgot to approve it

PR #757, opened 2025-04-28
Fix: mender-grow-datas service failed to grow data

tactmaster:
Service got blank device from `disk_get_device_base`. `disk_get_device_base`  returns nothing with e.g. /dev/sda

Ticket: None
Changelog:  fix issue where `mender-grow-data.service` doesn't run correctly on some integrations and the data partition won't be expanded


tactmaster:
I have tested it  on  RPI and qemu and a x86 machine. 
On the x86 it takes 3m to run every boot. With itt waiting like this: 

`May 02 21:05:09 hostname udevadm[487]: Timed out for waiting the udev queue being empty.`

I think this maybe machine specfic. 

However i changed the service type to simple instead of one shot. However I am not sure this a good generic solution
https://github.com/mendersoftware/mender-convert/blob/420363dc332c002e14b7eb73d90be07d1d1b68d7/configs/systemd_common_config#L16

PR #780, opened 2025-09-29
MEN-8814: Fix (restore) explicit broken symlink handling

lluiscampos:
(Commit marked as chore and not fix because the bug is not released).

This reverts commit 88786e and extends the functionality to the new code.

Next time I will read the comment before "fixing" code.


vpodzime (approved):
Looks good to me. It makes me think we could make use of some `backup()` and `restore()` functions, but that's a separate story.

PR #868, opened 2026-08-12
Test more images

elkoniu:
This PR extends default Debian and Ubuntu images to test all supported Debian and Ubuntu versions conversion.


mender-convert-modify
```diff
@@ -251,8 +251,12 @@ fi
 log_info "Using boot partition device in fstab: $boot_part_device"
 log_info "Using data partition device in fstab: $data_part_device"
 
-fstab=$(sed '/\/boot/d' work/rootfs/etc/fstab)
-fstab=$(sed '/.*\s\/\s.*/d' <<< "$fstab")
+# On systemd-repart based distros fstab may be runtime created and missing in the image file
+fstab=""
+if [ -f work/rootfs/etc/fstab ]; then
+    fstab=$(sed '/\/boot/d' work/rootfs/etc/fstab)
+    fstab=$(sed '/.*\s\/\s.*/d' <<< "$fstab")
```
vpodzime: I know you are just moving this around, but let's do `sed -e '...' -e '...'` instead of these two `sed` runs with Bash magic.

modules/probe.sh
```diff
@@ -44,12 +44,9 @@ probe_arch() {
...
     file_info=""
-    for location in bin/ls usr/bin/ls; do
-        if [ -L work/rootfs/${location} ]; then
-            location=$(readlink work/rootfs/${location})
-        fi
-        if [ -e work/rootfs/${location} ]; then
-            file_info=$(file -b --dereference work/rootfs/${location})
+    for location in "work/rootfs/bin/ls" "work/rootfs/usr/bin/ls"; do
+        if [ -e "${location}" ]; then
+            file_info=$(file -bL "${location}")
```
vpodzime, line 49: This change definitely deserves more explanation in the commit message than just calling it an improvement that for some reason fixes something.

scripts/test/run-tests.sh
```diff
@@ -98,16 +104,81 @@ fi
...
+                   "--" \
+                   "$@" \
+                   || test_result=$?
+fi
+
+if [ "$TEST_ALL" == "1" -o "$TEST_PLATFORM" == "ubuntu-22-qemux86-64" ]; then
+  MATCHED_A_TEST=1
+  # For this test, we explicitly test compressed and uncompressed image conversion
+  # For most of the other tests, we just test the uncompressed image conversion to
+  # speed up the process
```
vpodzime, line 143: This used to be the case for one test, now it's in multiple places. I don't think that's necessary.

vpodzime (commented):
+ *gtub* in the last commit message.

elkoniu:
Failing Debian 13 test will be masked temparary here: https://github.com/mendersoftware/mender-image-tests/pull/114

mender-convert-modify
elkoniu: Done

scripts/test/run-tests.sh
elkoniu: The reason for copy paste of this note is that the original test has been multiplied by number of OS versions we are testing now against. But in the process I have removed those `MATCHED_A_TEST` markers.

modules/probe.sh
elkoniu: I have updated the commit message to reflect the change better.

modules/disk.sh
```diff
@@ -166,7 +166,7 @@ disk_create_file_system_from_folder() {
 
     run_and_log_cmd "mkdir -p work/output"
     run_and_log_cmd "sudo mount ${2} work/output"
-    run_and_log_cmd "sudo rsync --archive --xattrs --delete ${1} work/output/"
+    run_and_log_cmd "sudo rsync --archive --hard-links --xattrs --delete ${1} work/output/"
```
vpodzime, line 169: How is this _Fix rsync cleanup procedure_?

scripts/test/run-tests.sh
```diff
@@ -330,9 +317,4 @@ if [ "$TEST_ALL" == "1" -o "$TEST_PLATFORM" == "raspberrypi4_uefi_bookworm_64bit
                    || test_result=$?
 fi
 
-if [ "$MATCHED_A_TEST" = 0 ]; then
-  echo "No test matched!" 1>&2
```
vpodzime, line 211: Why is this deprecated? I'm not saying it's necessary, but I guess we were checking this for a reason. What makes the reason gone now?

scripts/test/run-tests.sh
vpodzime: > the original test has been multiplied by number of OS versions we are testing now against

Right. But the idea what to do it in one test to not waste time and money on doing it in multiple places. I don't see why we should do it in multiple places now. I don't think they differ regarding handling compressed vs. uncompressed images.

scripts/test/run-tests.sh
elkoniu: I tracked it back to the original PR which added this when there were only few tests. It look like marker which will just print the message when no valid test name has been passed to the script.

modules/disk.sh
elkoniu: True, I will update the commit message to add some context. When we were running the rsync without preserving the hard links in some cases / OSes conversion was failing.

scripts/test/run-tests.sh
vpodzime: Right, so it's unnecessary, but not deprecated. Deprecated would mean there's a new mechanism for the same thing.

elkoniu: I just rephrased the commit message to not use `deprecated` keyword and replace it with `unnecessary` + added some context.

vpodzime (approved):
Looks good to me now. Thanks!

PR #870, opened 2026-08-14
chore: Recognize *.raw as a supported image file

elkoniu:
Allow to recognize *.raw files the same way as *.img files


vpodzime (commented):
Would be great to have some justification/reasoning for this in the commit message -- what exactly it is required for,...

elkoniu:
@vpodzime thanks for the comment, I have updated the commit message to add explanation for this change.

vpodzime (approved):
Thanks! Please check the CI failures.

PR #873, opened 2026-08-18
feat: Recognize systemd-repart schema

elkoniu:
In old images we use to recognize only standard partitioning:
* partition 1 - EFI
* parittion 2 - rootfs

systemd-repart schema creates two pools - one for "static" partitions like rootfs to not break the references (numbers 1-11), and one for "special" partitions which may change / shift in future, like EFI (numbers 12-15).

Images compatible with systemd-repart schema will have:
* partition 1 - rootfs
* partition 14 - boot
* partition 15 - EFI

### DIFF
--- a/mender-convert-modify
+++ b/mender-convert-modify
@@ -91,6 +91,9 @@ source modules/config.sh $(printf "%q " "${configs[@]}")
 boot_part=$(disk_boot_part)
 root_part=$(disk_root_part)
 
+log_info "Boot partition set to: ${boot_part}"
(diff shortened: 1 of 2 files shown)


elkoniu:
@vpodzime just for the record - the ideal approach would be to scan unpacked partition images and find the right number by analyzing the content. The problem is we have this logic hardcoded on so many levels, that doing this "right" most likely would lead to implementing `mender-convert` from scratch :)

vpodzime (approved):
Looks good to me.

PR #874, opened 2026-08-18
feat: configurable data filesystem size

MarlonAtStogger:
### Description

added MENDER_DATA_FS_SIZE_MB variable to configure the filesystem size of the data partition. This can be used to decrease the size of the data partition in the initial deployment image.

Ticket: None

In our workflow we define the data partition size and setup the root partition size to automatically scale to the remaining space. Because of the way mender-convert works this results in the whole data partition being included in the bmap. For now the easiest solution is configuring the filesystem size of the data partition separate from the partition size.

Ideally mender-convert should not use any intermediary images but build everything directly in the final image. However this has its own set of challenges.

Let me know if this is usable as is or if needs any changes.

### DIFF
--- a/configs/mender_convert_config
+++ b/configs/mender_convert_config
@@ -111,6 +111,11 @@ MENDER_BOOT_PART_SIZE_MB="40"
 # on first boot to occupy the renaming free blocks on the physical storage.
 MENDER_DATA_PART_SIZE_MB="128"
 
+# The size of the data filesystems, in MiB. If the value is -1, mender-convert
+# will use the maximum size that will fit inside the created partition.
+#
+MENDER_DATA_FS_SIZE_MB="-1"
+
 # Alignment of partitions used when building partitioned images, expressed in
 # bytes
 #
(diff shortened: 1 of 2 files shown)


configs/mender_convert_config
```diff
@@ -111,6 +111,11 @@ MENDER_BOOT_PART_SIZE_MB="40"
 # on first boot to occupy the renaming free blocks on the physical storage.
 MENDER_DATA_PART_SIZE_MB="128"
 
+# The size of the data filesystems, in MiB. If the value is -1, mender-convert
+# will use the maximum size that will fit inside the created partition.
+#
+MENDER_DATA_FS_SIZE_MB="-1"
```
elkoniu, line 117: I think the better approach would be to check if this is defined, but empty. This is the concept we were using for other variables. `-1` sound like adding a magic number or error code which may lead to misunderstanding in future.

elkoniu: MENDER_DATA_PART_SIZE_MB variable may be related to this and already do the trick

elkoniu: https://docs.mender.io/operating-system-updates-debian-family/variables

PR #875, opened 2026-08-18
chore: Add configuration files for mkosi image creation

elkoniu:
In the past for test image creation this repo was using mkosi with the set of command line parameters.

Many of parameters has been deprecated over the years. This PR is recreating this mechanics based on modern mkosi configuration files.


elkoniu:
`test:check-shell-formatting` can be red for this as I am quoting in README some bash code which will not be executed here.

scripts/mkosi/README
```diff
@@ -0,0 +1,97 @@
...
+
+If everything is fine:
+* gzip the img file
+* push it to S3 (https://downloads.mender.io/mender-convert/images/*)
+* update link to image in scripts/test/run-tests.sh
+* pray it will work out of the box
+
+Notes:
+* old images which were working fine had an empty /etc/fstab (placeholder)
+* mkosi mixes predefined system config with user confis so value which is
```
vpodzime: *configs*

vpodzime (commented):
The _repart_ configs are the same and they basically have to be, right? I believe we should de-duplicate those, perhaps using symlinks if there is no better way.

Looks good to me otherwise.

danielskinstad:
> test:check-shell-formatting can be red for this as I am quoting in README some bash code which will not be executed here.

Won't this mean that all subsequent PRs will be red because of the failing shell check? 
You should excude the README from https://github.com/mendersoftware/mender-convert/blob/2b206e407cb8761431359474f60c932eaf3a0505/.gitlab-ci.yml#L145

elkoniu:
> > test:check-shell-formatting can be red for this as I am quoting in README some bash code which will not be executed here.
> 
> Won't this mean that all subsequent PRs will be red because of the failing shell check? You should excude the README from
> 
> https://github.com/mendersoftware/mender-convert/blob/2b206e407cb8761431359474f60c932eaf3a0505/.gitlab-ci.yml#L145

Thanks for pointing this out! After I checked the rule it looks like all *.md files are excluded so I added missing extension now :)

scripts/mkosi/README
elkoniu: Thanks, fixed

elkoniu:
> The _repart_ configs are the same and they basically have to be, right? I believe we should de-duplicate those, perhaps using symlinks if there is no better way.
> 
> Looks good to me otherwise.

I have replaced the same files with symbolic links (partitioning schema and post configuration).

scripts/mkosi/debian-12/mkosi.postinst.chroot
```diff
@@ -0,0 +1 @@
+../mkosi.postinst.chroot
```
vpodzime, line 1: Did you forget to `git add` the target of these symlinks? (and the same applies to `mkosi.repart` below)

vpodzime (commented):
Looks good to me otherwise.

scripts/mkosi/debian-12/mkosi.postinst.chroot
elkoniu: I just checked it and I think GitHub is showing it some wierd way. In each directory: debian-{11,12} and ubuntu-{22,24,26}.04 there should be a symbolic link to `../mkosi.repart` and `../mkosi.postinst.chroot`. Those files are also part of this PR.

vpodzime: Yeah, now I see it. Sorry for the noise.

danielskinstad (approved):

elkoniu:
Updated version of this PR has been merged as a part of https://github.com/mendersoftware/mender-convert/pull/868

PR #817, opened 2026-01-28
RPi5 should use kernel_2712.img (not kernel8.img)

user706:
The following (`kernel8.img`) works for RPi5, but is not optimal:

https://github.com/mendersoftware/mender-convert/blob/5.1.0/configs/raspberrypi/uboot/include/raspberrypi_64_config#L4-L5

Ref: https://www.raspberrypi.com/documentation/computers/config_txt.html#kernel  
_"The Raspberry Pi 5, 500, 500+, and Compute Module 5 firmware defaults to loading `kernel_2712.img` because this image contains optimisations specific to those models (e.g. 16K page-size). If this file is not present, then the common 64-bit kernel (`kernel8.img`) will be loaded instead."_

It's better if RPi5 uses `kernel_2712.img`, since it has optimizations.

PR #878, opened 2026-09-13
chore: Add RPi5 config favoring performance

elkoniu:
New configuration file will use kernel_2712.img instead of the default kernel8.img and will not affect the existing configs.


elkoniu:
This PR is based on https://github.com/mendersoftware/mender-convert/pull/817 and instead of changing the default RPi5 configuration file - it adds performance oriented one. This is the suggestion I got from @estenberg when we discussed the original PR.

estenberg (approved):

